PR #1, opened 2016-06-29
make all goals into verb phrases

brandondrew:
It's easier to read a list if all items in the list are grammatically similar.  All of the other items were verb phrases--which makes sense for goals!--so I changed the second item to also be a verb phrase by adding "be" to the beginning.


asterite:
Looks good, thanks!

PR #3, opened 2016-06-29
Changed paragraph title

costajob:
Changed paragraph title in order to make it properly related to the underneath content.


asterite:
Good catch, thanks!

PR #6, opened 2016-07-26
change camelcased to PascalCased

davidgrieser:
The only online reference that I found for this was from Microsoft's own [naming conventions](https://msdn.microsoft.com/en-us/library/x2dbyw72(v=vs.71).aspx).

I always have to remember to camelCase for JavaScript as I am so used to PascalCase for classes and modules in Ruby and underscore-case for methods. So, when I read this I got confused. Maybe no one else would notice or they skip over it.

If this doesn't fit with the goals, feel free to close, just thought I'd submit it anyway.

### DIFF
--- a/conventions/coding_style.md
+++ b/conventions/coding_style.md
@@ -4,7 +4,7 @@ This style is used in the standard library. You can use it in your own project t
 
 ## Naming
 
-__Type names__ are camelcased. For example:
+__Type names__ are PascalCased. For example:
 
 ```crystal
 class ParseError < Exception


asterite:
I think this is OK, the difference seems to be that this is `camelCase` and this is `PascalCase`.

In Ruby's ActiveSupport there's the `camel_case` (and `camelize`) method that by default gives you PascalCase... so that's probably why we used camel case in the first place, but I wouldn't mind changing it.

davidgrieser:
Interesting. Thanks for pointing out those methods out.

Reading into it, they simplified things by just implementing `#camelize`, but [in the docs](http://api.rubyonrails.org/classes/ActiveSupport/Inflector.html) note that it converts to `UpperCamelCase` by default but you can pass, a second argument to get `lowerCamelCase`.

Maybe going with `UpperCamelCase` over `PascalCase` would make sense? As it seems to be more common.

RX14:
I think `UpperCamelCase` and `PascalCase` are used interchangeably, but camel case means "lowerCamelCase" to me by default.

PR #16, opened 2016-08-11
add explanation and fix example by adding getters

### LINKED ISSUE #14: Documentation missing explanation on getter/setter macros and obj.@var syntax
(Moved here from crystal-lang/crystal#3135)

I just noticed the documentation seems to be missing the description of the `getter`/`setter` macros. The `property` macro is used in the _Structs_ page though.

Also, I saw in source code that `obj.@var` is valid, and I guess it's field access without "method call" (even if there was a getter method that might do something else in addition)? This syntax isn't explained either in the documentation.

Of course one can learn these things through other ways too. Just putting it out there so it won't get forgotten in case the documentation is supposed to be comprehensive. If wanted, and if my assumptions are correct, I could describe it somewhere. But I doubt my writing style is as beautiful as the one in Crystal's documentation. :)

paddor:
The next example snippet depends on this. This also adds a short
explanation on the macros getter, setter, and property.

This should fix #14.

### DIFF
--- a/syntax_and_semantics/methods_and_instance_variables.md
+++ b/syntax_and_semantics/methods_and_instance_variables.md
@@ -4,11 +4,20 @@ We can simplify our constructor by using a shorter syntax for assigning a method
 
 ```crystal
 class Person
+  getter name
+  getter age
+
   def initialize(@name : String)
     @age = 0
   end
 end
+
+jen = Person.new "Jen"
+jen.name #=> "Jen"
+jen.age #=> 0
 ```
+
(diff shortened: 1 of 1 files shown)


Sija:
GTG?

asterite:
Maybe, I don't know if it's a good idea to introduce macros at that point.

petehuang:
Agree, feels a little early. I think it's in the right section but introduced out of order within that section. My suggestion is to move the getter description to the end of that subsection and only briefly mention "macro method" if at all.

andyfleming:
I've got a proposal in https://github.com/crystal-lang/crystal-book/issues/62 that is related.

mverzilli:
Closing in favor of @andyfleming 's proposal at #62. Feel free to collaborate/discuss there!

PR #22, opened 2016-08-23
Update install guide for redhat/centos

amitsaha:


mverzilli:
I'm totally unfamiliar with `dnf`, has it replaced `yum` as the default package manager for Fedora derivatives?

mverzilli:
Closing for lack of response. If you're interested in following up we can reopen. Thanks!

PR #23, opened 2016-08-23
Update HTTP server documentation

amitsaha:


mverzilli:
I agree with @petehuang. It doesn't seem the right part of the book to introduce the topic of concurrency. Additionally, concurrency and parallelism are being worked on right now so we want to avoid documenting it further (until a more "final" design gets stabilized).

PR #25, opened 2016-09-20
Correct naming of sample project

schmijos:
We should stick to the coding conventions described here: https://crystal-lang.org/docs/conventions/coding_style.html#directory-and-file-names


asterite:
Thank you, this is correct!

PR #31, opened 2016-10-04
Recursive C structs do not need a forward declaration

molenzwiebel:
As a matter of fact, the current code will error with a double declaration of `Node`. I have also updated the example to something a bit more relevant.


asterite:
Right, thanks! We improved this a long time ago but forgot to update the docs.

PR #46, opened 2016-11-25
Fix typo in type inference section

zoso10:


mverzilli:
Good catch, thank you! ❤️

PR #48, opened 2016-11-29
update notes about running crystal after building from source

jwoertink:
This is an update for clarification from https://github.com/crystal-lang/crystal/issues/3605#issuecomment-263647886 as suggested by @bcardiff


bcardiff:
1. I would say to handle that on another PR, explaining how llvm-config is used to locate installed llvm.
2. I think git / directory usage can be assumed here.
3. A reference to https://crystal-lang.org/docs/using_the_compiler/ saying use ./bin/crystal instead of just crystal should be enough.

I would say to do (3) and then merge this. @jwoertink are you up to do this? and the we finally merge this. It is a really good addition. (sorry for the delay).

jwoertink:
Yup. I'll do that now. Thanks!

bcardiff:
Thanks!

Issue #47, opened 2016-11-27
Misleading Wording in "Methods and instance variables" Section


nodanaonlyzuul:
Firstly thank you for the wonderfully thought out  and designed resource.

I noticed that in [methods_and_instance_variables.md](https://github.com/crystal-lang/crystal-book/blob/master/syntax_and_semantics/methods_and_instance_variables.md) there's a code snippet:

```crystal
class Person
  def initialize(@name : String)
    @age = 0
  end
end
```

Followed by:

> Right now, we can't do much with a person: create it with a name, ask for its name and for its age, which will always be zero. So lets add a method that makes a person become older:

But - really - we can't query the state of `@age` or `@name` without using `getter :age, :name`.

If you feel like this is big enough to address I can send a PR...perhaps...

> Right now, we can't do much with a person aside from create it with a name. Its age will always be zero. So lets add a method that makes a person become older.

RX14:
You technically can query it's name and age but I think the docs should pretend that a certain syntax doesn't exist.

nodanaonlyzuul:
@RX14 : Well now you've piqued my interest into how peek at private instance variables.

nodanaonlyzuul:
Closing this in favor of https://github.com/crystal-lang/crystal-book/pull/50.

nodanaonlyzuul:
Thanks again for thinking of this - I'm fine with any changes to wording maintainers suggest or even if you don't want the change in. Again - Thanks for the great documentation.

PR #50, opened 2016-12-03
Change wording about what we can do with a Person

nodanaonlyzuul:
This is a PR that represents something discussed in https://github.com/crystal-lang/crystal-book/issues/47...

The [methods_and_instance_variables.md](https://github.com/crystal-lang/crystal-book/blob/master/syntax_and_semantics/methods_and_instance_variables.md) there's a code snippet:

```crystal
class Person
  def initialize(@name : String)
    @age = 0
  end
end
```

Followed by:

> Right now, we can't do much with a person: create it with a name, ask for its name and for its age, which will always be zero. So lets add a method that makes a person become older:

But - really - we can't query the state of `@age` or `@name` without using `getter :age, :name`.

If you feel like this is big enough to address I can send a PR...perhaps...

> Right now, we can't do much with a person aside from create it with a name. Its age will always be zero. So lets add a method that makes a person become older.

### DIFF
--- a/syntax_and_semantics/methods_and_instance_variables.md
+++ b/syntax_and_semantics/methods_and_instance_variables.md
@@ -10,7 +10,7 @@ class Person
 end
 ```
 
-Right now, we can't do much with a person: create it with a name, ask for its name and for its age, which will always be zero. So lets add a method that makes a person become older:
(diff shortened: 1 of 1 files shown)


mverzilli:
Thank you ❤️ !

PR #51, opened 2016-12-08
Add 'Install required libraries' on Ubuntu install

NeuwDk:
I got frustrated that the required libraries weren't referenced here when I installed Crystal.

I copied from the wiki at https://github.com/crystal-lang/crystal/wiki/All-required-libraries

It would be great if this was also added to the other install pages, but since this is my first rodeo with pull-requests I'd like to keep it simple (for me).

### DIFF
--- a/installation/on_debian_and_ubuntu.md
+++ b/installation/on_debian_and_ubuntu.md
@@ -25,6 +25,30 @@ Once the repository is configured you're ready to install Crystal:
 sudo apt-get install crystal
 ```
 
+## Install required libraries
+Before you can use Crystal you'll need to install some required libraries:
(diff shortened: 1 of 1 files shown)


mverzilli:
@NeuwDk Thank you and welcome to the community! ❤️ . This is actually because our packages don't list their dependencies properly. We're tracking that issue here: https://github.com/crystal-lang/omnibus-crystal/issues/14

To the extent possible, we want to avoid going into technical minutiae in the book. In this case it's clearly a problem in our release process, so it's better not to "canonicalize" it in the documentation. Let's fix the thing, instead of working around it! :)

PR #53, opened 2016-12-12
basic crystal-db usage

bcardiff:
This adds a Database section before the performance and concurrency guides.
Explains only the very basics of crystal db.

I think it will make sense to later add subsections inside database/ folder for

* Connection pool 
* Prepared statements 
* Transactions

Meanwhile this can be published after review IMO.

### DIFF
--- a/database/README.md
+++ b/database/README.md
@@ -1,0 +1,125 @@
+# Database
+
+To access a relational database you will need a shard designed for the database server you want to use. The package [crystal-lang/crystal-db](https://github.com/crystal-lang/crystal-db) offers a unified api across different drivers.
+
+The following packages are compliant with crystal-db
+
+* [crystal-lang/crystal-sqlite3](https://github.com/crystal-lang/crystal-sqlite3) for sqlite
+* [crystal-lang/crystal-mysql](https://github.com/crystal-lang/crystal-mysql) for mysql & mariadb
+* [will/crystal-pg](https://github.com/will/crystal-pg) for postgres
+
+This guide presents the api of crystal-db, the sql commands might need to be adapted for the concrete driver due to differences between postgres, mysql and sqlite.
+
+Also some drivers may offer additional functionality like postgres `LISTEN`/`NOTIFY`.
+
+## Installing the shard
+
+Choose the appropriate driver from the list above and add it as any shard to your application's `shard.yml`
+
+There is no need to explicitly require `crystal-lang/crystal-db`
+
+During this guide `crystal-lang/crystal-mysql` will be used.
+
+```yaml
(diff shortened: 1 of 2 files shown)


database/README.md
```diff
@@ -0,0 +1,122 @@
...
+
+This guide presents the api of crystal-db, the sql commands might need to be adapted for the concrete driver due to differences between postgres, mysql and sqlite.
+
+Also some drivers may offer additional functionality like postgres `LISTEN`/`NOTIFY`.
+
+## Installing the shard
+
+Choose the appropriate driver from the list above and add it as any shard to your application's `shard.yml`
+
+There is no need to explicit require `crystal-lang/crystal-db`
```
spalladino: **explicitly** require...

database/README.md
```diff
@@ -0,0 +1,122 @@
...
+
+```yaml
+dependencies:
+  mysql:
+    github: crystal-lang/crystal-mysql
+```
+
+## Open database
+
+`DB.open` will allow you to easy connect to a database using a connection uri. The schema of the uri determines the expected driver. The following sample connects to a local mysql database named test with user root and password blank.
```
spalladino: **easily** connect...

database/README.md
```diff
@@ -0,0 +1,122 @@
...
+end
+```
+
+Other connection uris are
+
+* `sqlite3:///path/to/data.db`
+* `mysql://user:password@server:port/database`
+* `postgres://server:port/database`
+
+Alternative you can use a non yielding `DB.open` method as long as `Database#close` is called at the end.
```
spalladino: **Alternatively** you can...

database/README.md
```diff
@@ -0,0 +1,122 @@
...
+
+```crystal
+require "db"
+require "mysql"
+
+db = DB.open "mysql://root@localhost/test"
+# ... use db to perform queries
+db.close
```
spalladino: I'd add an `ensure` for the `db.close`

bcardiff:
@spalladino done.

@asterite do you agree with the plan? eventually it will be something similar to http://go-database-sql.org/

asterite:
Mmm... OK :-)

It seems a bit strange to have this in the docs if crystal-db is not in the standard library. Why not document this in crystal-db?

(I'm not against this, it just seems a bit unorganized)

spalladino:
The other option would be to set up another set of docs, related to common use cases: webserver development, CLI apps, database access, etc. Since we are just getting started, I'd add them to the main guides for now, and then see if we have enough content to split.

asterite:
@spalladino Sounds good 👍

PR #56, opened 2016-12-23
Improve "white space" pages + minor edits

petehuang:
This PR:

1. Repurposes some section opener pages as tables of contents for their sections. A few of the section opener pages were very empty, sometimes consisting of a single line. For hygiene, I added the list of pages that follow for each section. E.g., syntax and semantics, control expressions

2. Renames the `if` page to `if / else / elsif`. I think it's clearer but I concede that other docs will also leave it as just `if`.

3. Makes minor changes in wording and linking

### DIFF
--- a/syntax_and_semantics/type_reflection.md
+++ b/syntax_and_semantics/type_reflection.md
@@ -1,3 +1,10 @@
 # Type reflection
 
-Crystal provides basic methods to do type reflection, casting and introspection.
+Crystal provides basic methods to do type reflection, casting and introspection:
+
+* [is_a?](is_a.html)
+* [nil?](nil_question.html)
+* [responds_to?](responds_to.html)
+* [as](as.html)
+* [as?](as_question.html)
+* [typeof](typeof.html)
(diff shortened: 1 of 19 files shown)


syntax_and_semantics/type_reflection.md
```diff
@@ -1,3 +1,10 @@
 # Type reflection
 
-Crystal provides basic methods to do type reflection, casting and introspection.
+Crystal provides basic methods to do type reflection, casting and introspection:
+
+* [is_a?](is_a.html)
```
andyfleming: While these lists of links are nice, they are going to be hard to maintain and are already available in the sidebar. Unless we can generate this in a more automated way, I would vote against putting redundant lists on the section parent pages.

mverzilli:
@petehuang thank you! I'm closing this PR because most of it is about providing "sub-TOCs" for the book's chapters, and without some sort of automated process to generate or at least test there aren't broken links it is very likely to get outdated. As @andyfleming mentioned, there's already a TOC at the left side to cover play that role (even if it's not as nice, it's much more likely to remain sound).

PR #59, opened 2016-12-28
Adding comparisons page

andyfleming:
Open to feedback here, but I feel like it might be helpful to have a "comparisons" page.


andyfleming:
These changes are in with #60.

PR #60, opened 2016-12-28
Adding examples of working with strings and arrays

andyfleming:
I talked with @Sija briefly about adding this section.

It's just a start and I know there is a ton that can be added here, but I was personally looking for examples like these so I thought it would be helpful to add them.

I also have fallen in love with the simplicity of the operators to access ranges of strings and arrays. Maybe this is the same as it has always been in Ruby (I don't actually know off-hand), but it's a nice clean API. One exception was the method name `gsub`, which was strange to me at first, until I realized it was short for global substitution.


syntax_and_semantics/type_workflow/arrays.md
```diff
@@ -0,0 +1,37 @@
...
+
+# Access from the end of an array
+my_array[-2] # => 'c' : Char
+
+# Access a range of an array
+my_array[1..3] # => ['b', 'c', 'd'] : Array(Char)
+
+# Start at an element in the array and continue for X elements
+my_array[2, 2] # => ['c', 'd']
+
```
Sija: unnecessary empty lines

syntax_and_semantics/type_workflow/strings.md
```diff
@@ -0,0 +1,27 @@
...
+# Access a part of a string
+"abcd"[1..2] # => "bc" : String
+
+# Start at a position and continue for X characters
+"abcd"[2,2] # => "cd" : String
+
+# Concatenate strings
+"abcd" + "e" # => "abcde"
+
+#
```
Sija: comment should be before code

andyfleming:
Made those changes. Let me know if anything else needs done.

syntax_and_semantics/type_workflow/strings.md
```diff
@@ -0,0 +1,26 @@
...
+"abcd"[1..2] # => "bc" : String
+
+# Start at a position and continue for X characters
+"abcd"[2,2] # => "cd" : String
+
+# Concatenate strings
+"abcd" + "e" # => "abcde"
+
+# Regex Match (will be evaluated as truthy in a conditional statement if a match is found)
+"abcd".match(/b/) # => #<Regex::MatchData "b"> : (Regex::MatchData | Nil)
```
Sija: I'd add example of getting all matches with `String#scan`: sth like `"Hello world".scan(/\w+/)`

syntax_and_semantics/comparisons.md
```diff
@@ -0,0 +1,33 @@
+# Comparisons
+
+Comparing Strings
+
+```crystal
+"a" == "a"		# true : Bool
+"a" === "a"		# true : Bool
+"a" == "b"		# false : Bool
+"a" === "b"		# false : Bool
```
Sija: I'd add comparisons of `String` with `Regex` too:

```cr
"a" =~ /a/ # => 0
"a" =~ /b/ # => nil

"a" !~ /a/ # => false
"a" !~ /b/ # => true
```

andyfleming: ✅

syntax_and_semantics/type_workflow/strings.md
andyfleming: ✅

andyfleming:
This should be all set, unless you had other ideas to add for this first edition.

syntax_and_semantics/comparisons.md
```diff
@@ -0,0 +1,39 @@
...
+
+"a" =~ /a/ # => 0
+"a" =~ /b/ # => nil
+
+"a" !~ /a/ # => false
+"a" !~ /b/ # => true
+
+"a" <=> "b"		# -1 : Int32
```
Sija: I'd suggest to stick to `# => <value>` syntax for return values and drop `: <type>` altogether from comments.

andyfleming: Do you think `: <type>` should be omitted from all pages and examples in this PR?

Sija: yep

mverzilli:
Thanks for this @andyfleming! I began reviewing this PR, but then I couldn't help a gut feeling that this info fits better in the Array and String API docs than in the book. Ideally I'd expect the book to be more conceptual and less of a reference (that's what the API docs are for). 

I think there's value in adding "Working with Strings" and "Working with Arrays" guides to the book, discussing how to generally get stuff done with them idiomatically and without impacting performance, maybe even walking the reader through a practical example.

I'd close this PR, but I'd like @asterite to confirm :).

asterite:
I agree with @mverzilli 

I also think we need a separate thing here, something more like a tutorial on how to use Crystal and its various APIs.

The book is about the language elements (variables, control structures, classes, structs, etc.) and their semantics, not about specific types in the language and how to use them. There are some sections about String and Array in the book, but only because they are fundamental types that have their own literals. The book merely covers that syntax without going into the whole full blown API of these types. That's the job of the API docs, and/or a tutorial to show how to combine these things to produce a useful program.

(of course the book has some snippets because otherwise it will be very hard to explain some things)

This is also why I don't think talking about `getter`, `setter` and `property` is appropriate in the book: these are just regular macros, not something that's bound to the language's syntax and semantics.

mverzilli:
I do think those tutorials could well go in their own chapter of the book (maybe under "Guides"?). And as regards `getter`, `setter` and `property`, they're so commonly used and basic features that it doesn't hurt to show them early in the book without spending many lines on them. 

Anyway, @andyfleming feel free to open a similar PR to this one in crystal-lang/crystal to enhance the API docs for `Array` and `String`.

PR #61, opened 2016-12-29
Fixing example to be more complete

andyfleming:
(to function when copy-pasted)

### DIFF
--- a/syntax_and_semantics/methods_and_instance_variables.md
+++ b/syntax_and_semantics/methods_and_instance_variables.md
@@ -7,6 +7,9 @@ class Person
   def initialize(@name : String)
     @age = 0
   end
+  def age
+    @age
+  end
 end
 ```
 
@@ -14,6 +17,12 @@ Right now, we can't do much with a person: create it with a name, ask for its na
 
 ```crystal
 class Person
+  def initialize(@name : String)
+    @age = 0
+  end
+  def age
+    @age
+  end
   def become_older
     @age += 1
   end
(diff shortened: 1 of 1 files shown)


syntax_and_semantics/methods_and_instance_variables.md
```diff
@@ -7,13 +7,25 @@ class Person
   def initialize(@name : String)
     @age = 0
   end
+  
+  def age
```
mverzilli: Maybe we should also write the `name` getter, what do you think? Mostly to comply with "ask for its name and for its age" in the paragraph below

syntax_and_semantics/methods_and_instance_variables.md
```diff
@@ -7,13 +7,25 @@ class Person
...
 
 Right now, we can't do much with a person: create it with a name, ask for its name and for its age, which will always be zero. So lets add a method that makes a person become older:
 
 ```crystal
 class Person
+  def initialize(@name : String)
+    @age = 0
+  end
+  
+  def age
```
mverzilli: Ditto.

syntax_and_semantics/methods_and_instance_variables.md
karlseguin: I think the original commit is good. The phrase that you quote appears to have been removed by https://github.com/crystal-lang/crystal-book/commit/8b493993f435c9e04658783751faa01f5b5c29c4

mverzilli:
Thank you!

PR #64, opened 2016-12-31
Fix typo [visibility]

jmcaffee:
I believe the additional 'if' was misleading.

Because public is the default if there is no `public` keyword =>
Because public is the default there is no `public` keyword

### DIFF
--- a/database/README.md
+++ b/database/README.md
@@ -74,6 +74,8 @@ db.exec "insert into contacts values (?, ?)", "John", 30
 db.exec "insert into contacts values (?, ?)", "Sarah", 33
 ```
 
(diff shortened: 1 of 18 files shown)


andyfleming:
Looks good.

----

Alternatively, we could also consider a simpler phrasing:

Before:

> Methods are public by default: the compiler will always let you invoke them. Because public is the default there is no `public` keyword.

After:

> Methods are public by default. There is no `public` keyword.

I think it communicates all that necessary for the reader.

mverzilli:
It seems someone took care of since you opened the PR, but I like this wording better, so merging.

PR #68, opened 2017-01-01
Adding example of Futures

andyfleming:
Adding an example of futures, which aren't really documented anywhere.

I think futures are important for people coming from Java, or even from JavaScript (as Promises have become commonplace).

### DIFF
--- a/guides/concurrency.md
+++ b/guides/concurrency.md
@@ -380,7 +380,7 @@ Note that the first 2 sends are executed without switching to another fiber. How
 
 ### Futures
 
-[Futures](https://en.wikipedia.org/wiki/Futures_and_promises) are also available. They are built on top of channels) and provide a simple alternative workflow for concurrent tasks. The resolved value from a future must be explicitly obtained by calling `.get()`.
+[Futures](https://en.wikipedia.org/wiki/Futures_and_promises) are also available. They are built on top of channels and provide a simple alternative workflow for concurrent tasks. The resolved value from a future must be explicitly obtained by calling `.get()`.
 
 ```crystal
 request1 = future do
--- a/guides/concurrency.md
+++ b/guides/concurrency.md
@@ -380,7 +380,7 @@ Note that the first 2 sends are executed without switching to another fiber. How
 
 ### Futures
 
-Futures are also available (as syntatic sugar, built on top of channels). They provide a simple alternative workflow for concurrent tasks.
+[Futures](https://en.wikipedia.org/wiki/Futures_and_promises) are also available. They are built on top of channels) and provide a simple alternative workflow for concurrent tasks.
 
 ```crystal
 request1 = future do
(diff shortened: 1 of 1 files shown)


guides/concurrency.md
```diff
@@ -377,3 +377,20 @@ After send
 ```
 
 Note that the first 2 sends are executed without switching to another fiber. However, in the third send the channel's buffer is full, so execution goes to the main fiber. Here the two values are received and the channel is depleted. At the third `receive` the main fiber blocks and execution goes to the other fiber, which sends another values, finishes, etc.
+
+### Futures
+
+Futures are also available (as syntatic sugar, built on top of channels). They provide a simple alternative workflow for concurrent tasks.
```
RX14: They're not really syntactic sugar, they're just built on channels.

andyfleming: Fair enough. I've updated the description.

guides/concurrency.md
```diff
@@ -377,3 +377,20 @@ After send
 ```
 
 Note that the first 2 sends are executed without switching to another fiber. However, in the third send the channel's buffer is full, so execution goes to the main fiber. Here the two values are received and the channel is depleted. At the third `receive` the main fiber blocks and execution goes to the other fiber, which sends another values, finishes, etc.
+
+### Futures
+
+[Futures](https://en.wikipedia.org/wiki/Futures_and_promises) are also available. They are built on top of channels) and provide a simple alternative workflow for concurrent tasks. The resolved value from a future must be explicitly obtained by calling `.get()`.
```
chocolateboy: > They are built on top of channels)

Stray closing parenthesis.

andyfleming: Good catch. Thanks! I updated the text accordingly.

mverzilli:
Thanks @andyfleming! Concurrency APIs are in flux, so I need to check with the team whether it's the right timing to document this. We'll get back to this PR in a couple of days.

asterite:
Indeed, the `future`, `lazy` and other methods were added to the standard library without much thought. I'm pretty sure we'll remove them or implement them in a different way in the future (no pun intended :-P).

andyfleming:
Thanks for taking a look at this @mverzilli and @asterite!

It may not be the right time to document this, but I'll take this moment to express my desire to see them remain in some form or fashion.

Futures are natural for me as someone coming from the world of Node.js (and will likely be for others with similar experience) as far as concurrency goes. One of the things I did when exploring Crystal was test out a practical concurrency example. I created gists to show the use of Futures in comparison with the use of Channels.

**Javascript Example (Promises):**
https://gist.github.com/andyfleming/ab5baa85704f8e205996ab44795ef501

**Crystal Examples (Channels and Futures):**
https://gist.github.com/andyfleming/f9e2a858579a7561f1d0248f1fffb693

andyfleming:
Also, I think presenting futures in documentation would allow for visitors to be exposed to the potential interface and contribute feedback accordingly.

We could include a disclaimer that they may be removed if we are that concerned.

mverzilli:
I still think including them in the book sends the wrong message. Nobody will read that disclaimer! :).

Maybe we can think of another place and format, a bit less "official"? Perhaps an "Experimental features (use at your own risk)" section at the crystal-lang/crystal wiki?

andyfleming:
An "Experimental Features" section could be cool. It serves as a more obvious disclaimer, but still allows things to be documented (for the sake of exposure).

mverzilli:
Cool, so I'm closing this for now.

PR #73, opened 2017-01-12
added note to centos install about installing readline too

masukomi:
added the following note to the CentOS install instructions because it's exactly what I had to do on our ancient version of CentOS. After that it worked fine.

----

### Note
If you see `/usr/bin/ld: cannot find -lreadline` when compiling you will want to install the readline development libraries:

```
sudo yum install -y readline-devel
```

### DIFF
--- a/installation/on_redhat_and_centos.md
+++ b/installation/on_redhat_and_centos.md
@@ -29,6 +29,13 @@ Once the repository is configured you're ready to install Crystal:
 sudo yum install crystal
 ```
 
+### Note
+If you see `/usr/bin/ld: cannot find -lreadline` when compiling you will want to install the readline development libraries:
+
+```
+sudo yum install -y readline-devel
+```
+
 ## Upgrade
 
 When a new Crystal version is released you can upgrade your system using:


RX14:
The fix for this is to fix the package dependencies.

mverzilli:
I agree that adding this kind of hyper-specific advice to the Crystal book will impact its readability. 

@RX14 do you mean Crystal's dependencies for that particular CentOS version? Can you point me to the artifact we should double-check? Maybe omnibus? 

@masukomi, in the best case we should be able to fix this so the note isn't necessary. In the worst case, this belongs in a Troubleshooting section somewhere (probably here: https://github.com/crystal-lang/crystal/wiki/All-required-libraries). 

I'll leave the PR open until we figure out what's the best follow up on this. Thanks!

RX14:
@mverzilli Sorry, not too familiar with the packaging size of things. It's probably in omnibus somewhere.

matiasgarciaisaia:
I'd close this in favour of crystal-lang/omnibus-crystal#14. As you've said, this is a bug in the packages.

Maybe we could list this dependencies in the Wiki - like we've done for Ubuntu and Fedora. CentOS being a Fedora derivative, is it needed to do a different section for it, or does the Fedora one work for CentOS too?

mverzilli:
Closing this in favor of crystal-lang/omnibus-crystal#14, as suggested by @matiasgarciaisaia. @masukomi if you could provide the specific version number of CentOS that caused you trouble please add it to that issue so we use it as a test case.

PR #75, opened 2017-01-29
Fixing confusion around the title "build"

andyfleming:
I've seen first-hand new visitors see "build" and initially assume it is how to build/install the actual language, when in fact it is instructions on how to build the crystal-book project.

I think there are more improvements that can be made here too, but this initial change would be helpful.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ Crystal is a programming language with the following goals:
 * Have compile-time evaluation and generation of code, to avoid boilerplate code.
 * Compile to efficient native code.
 
-## Contributing to These Docs
+## Contributing to the Language Reference
 
 Do you consider yourself a helpful person? If you find bugs or sections
 which need more clarification you're welcome to contribute to this
--- a/README.md
+++ b/README.md
@@ -10,7 +10,16 @@ Crystal is a programming language with the following goals:
 * Have compile-time evaluation and generation of code, to avoid boilerplate code.
 * Compile to efficient native code.
 
-## Build
+## Contributing to These Docs
+
+Do you consider yourself a helpful person? If you find bugs or sections
+which need more clarification you're welcome to contribute to this
+documentation. You can submit a pull request to this repository:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -10,7 +10,16 @@ Crystal is a programming language with the following goals:
 * Have compile-time evaluation and generation of code, to avoid boilerplate code.
 * Compile to efficient native code.
 
-## Build
+## Contributing to These Docs
```
RX14: "These" doesn't sound right to me.

andyfleming: I agree.

"Contributing" would be confusing because a user might think that it's talking about contributing to the language.

"Contributing to Documentation" sounds nice, but doesn't really specify that there are API docs that are important as well.

Any ideas?

mverzilli: What about "Contributing to the Crystal Book"?

RX14: Is it clear that by "crystal book" we are talking about what you're looking at though? On the crystal-lang.org homepage they're just referred to as "docs" and "api".

mverzilli: Probably it's not clear, but I think it'd be a good thing to change. IMO both the API reference and the book are documentation, so calling the latter "docs" is naturally confusing. 

What's your opinion about changing all "docs" references throughout the website and inside this repo to "book", "Crystal Book", or similar?

RX14: I would vote for "language reference" and "api reference". They're very clear and specific.

andyfleming: I like "Language Reference" and "API Reference".

mverzilli: Discussed this with the team and everyone likes "Language Reference" and "API Reference" :). So let's be explicit about that naming in this PR. We're working on a website redesign, so it's a good time to introduce the change there too.

matiasgarciaisaia: Should we rename the repo, then? I've always found "crystal-book" confusing.

matiasgarciaisaia:
Thanks, @andyfleming - and everyone for your inputs!

PR #77, opened 2017-01-30
Add requirement to instructions

diasjorge:
Document that Crystal is required in order to build from source

### DIFF
--- a/installation/from_source_repository.md
+++ b/installation/from_source_repository.md
@@ -5,17 +5,19 @@ If you want to contribute then you might want to install Crystal from sources. B
 You will also need LLVM 3.5 or 3.6 present in the path. If you are using Mac and the Homebrew formula, this will be automatically configured for you if you install Crystal adding `--with-llvm` flag.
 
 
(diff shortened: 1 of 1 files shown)


diasjorge:
I found this issue trying to compile from source. Found the solution in a comment for this issue: crystal-lang/crystal#1331

asterite:
First paragraph in the document says:

> If you want to contribute then you might want to install Crystal from sources. But Crystal is written in Crystal itself! So you first need to use one of the previous described methods to have a running compiler.

karlseguin:
Don't see too much harm in saying it twice. People might land on difference pages, or miss it once or not the other. Of course, that could be a slippery slope (why not say everything twice)...but it seems somewhere between benign and helpful in this case

mverzilli:
Yeah, I guess most people jumps straight to the numbered list and follows the steps, so it doesn't harm to be explicit there. So let's be more explicit: @diasjorge, could you make the first step of the list be sth like: "Install the latest Crystal release. To compile Crystal, you need Crystal :)." 

Have the first sentence be a link that points to `/installation`. Then we can merge :).

diasjorge:
I implemented the changes suggested per @mverzilli 
Although I think addressing the concerns mentioned I created another patch in #79 
In my case I jumped straight to the list and just skimmed over the first paragraphs. In the alternative patch I wrote everything as a list, hopefully you prefer the other patch and we can just close this one.

mverzilli:
I like what you propose in #79, so let's move there.

PR #90, opened 2017-03-14
llvm version update to 3.8, 3.9 or 4.0.

txdv:
Took this from https://github.com/crystal-lang/crystal/wiki/All-required-libraries which was referred to me by @sdogruyol


ysbaddaden:
This should state LLVM 3.8 - 4.0. There are some issues with LLVM 3.6 and below and support for these versions shall be removed soon. LLVM 4.0 is now supported, too.

txdv:
I updated it to your named versions!

mverzilli:
Thank you @txdv and and @ysbaddaden for following up!

PR #88, opened 2017-03-07
Remove `then` branch in `if` documentation

luislavena:
Clear out confusion caused by `then` being used as branch name of `if` documentation.

Fixes crystal-lang/crystal#4117

Thank you. :heart: :heart: :heart:

PR #92, opened 2017-03-16
Update sentence flow in if.md

RX14:
Just a small tweak to the sentence flow after #88.

Sorry I didn't get this into #88, I forgot to check out the new wording before ti was merged.


mverzilli:
No problem at all @RX14, I agree this is better. I noticed the same issue but given we're a bit swamped at the moment and the change already yielded value I simply merged and moved on. Thanks for this ❤️ .

luislavena:
Ouch, my bad, was so focused on the `then` wording that missed the other aspects.

Thank you!

PR #93, opened 2017-04-12
Removes octal escape from Char literal doc

arcage:
Removed descriptions about octal escape from document of the Char literal.
It has already been removed at [v0.21.0](https://github.com/crystal-lang/crystal/releases/tag/0.21.0).


mverzilli:
Thank you!

PR #95, opened 2017-04-18
Cover added for generated ebook

CodelessFuture:
I'm starting to work with Crystal. I think that learning the language I can contribute to the documentation.

As a first step I added the cover I use for the ebooks I've generated.

Official logo and font character.


Sija:
Will it merge?

PR #100, opened 2017-05-29
Change curl comamnd in Bash on Ubuntu on Windows installation to include -sSL flags

Redrield:
In my installation using the subsystem, the progress display from curl was getting in the way of the sudo prompt, adding these flags fixed that issue.


mverzilli:
Thank you!!

PR #104, opened 2017-06-19
Convert this code so that is compiles

JacobBallantine:
When I tested this code I had an error with line 14. The error said "wrong number of arguments for `Person.new`" So I added some more code so the code would work and it worked, for those who are testing and learning the things I added was.

### DIFF
--- a/syntax_and_semantics/overloading.md
+++ b/syntax_and_semantics/overloading.md
@@ -4,12 +4,10 @@ We can define a `become_older` method that accepts a number indicating the years
 
 ```crystal
 class Person
-  @age = 0
-
-    getter :age
+  getter :age
     
-    def initalize(@name : String)
-    end
+  def initialize(@name : String, @age : Int = 0)
+  end
 
   def become_older
     @age += 1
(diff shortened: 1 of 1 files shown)


syntax_and_semantics/overloading.md
```diff
@@ -6,6 +6,11 @@ We can define a `become_older` method that accepts a number indicating the years
 class Person
   @age = 0
 
+    getter :age
+    
+    def initalize(@name : String)
```
mverzilli: There's a typo here: `initalize` => `initialize`.

syntax_and_semantics/overloading.md
```diff
@@ -6,6 +6,11 @@ We can define a `become_older` method that accepts a number indicating the years
 class Person
   @age = 0
```
mverzilli: Can you move this line to the new constructor? I it'll be tidier.

mverzilli:
Thanks! I left you some comments.

JacobBallantine:
Thanks @mverzilli, I've updated the code as you've requested. 
Appreciate your feedback.

mverzilli:
Thank you!

PR #107, opened 2017-06-30
Document Apt signing key change

ivy:
Following the conversation in crystal-lang/crystal#4648, I've added a short note about the change in signing key and how users with existing installs can upgrade.

**Warning:** It appears that <https://dist.crystal-lang.org/apt/setup.sh> hasn't been updated yet to reflect this:

```sh
#!/usr/bin/env bash
apt-key adv --keyserver keys.gnupg.net --recv-keys 09617FD37CC06B54
echo "deb https://dist.crystal-lang.org/apt crystal main" > /etc/apt/sources.list.d/crystal.list
apt-get update
```

Users of the script will continue to face this issue.

### DIFF
--- a/installation/on_debian_and_ubuntu.md
+++ b/installation/on_debian_and_ubuntu.md
@@ -13,7 +13,7 @@ curl https://dist.crystal-lang.org/apt/setup.sh | sudo bash
 That will add the signing key and the repository configuration. If you prefer to do it manually, execute the following commands as *root*:
 
 ```
-apt-key adv --keyserver keys.gnupg.net --recv-keys 09617FD37CC06B54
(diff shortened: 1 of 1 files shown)


straight-shoota:
If the keys stay this way, the instructions for Ubuntu on Windows need to be updated as well.

mverzilli:
Thanks for your support @ivy and @straight-shoota! Let's see what the solution to #4648 is before merging this.

RX14:
I'm confused, did we change the whole key or just add a new subkey? If it's the latter why does the setup.sh not work? Is the key id provided for a subkey? Has the new master pubkey not been pushed to keys.gnupg.net?

straight-shoota:
This PR seems to be obsolete.

PR #109, opened 2017-07-01
Update nil_question.md

gburgessiv:
Add a missing word.


mverzilli:
Good catch! Thank you!

PR #110, opened 2017-07-21
Change wrong wording in documenting_code.md

GetRektByMe:
Change from two first person statements to one (correcting the other to third person)

### DIFF
--- a/conventions/documenting_code.md
+++ b/conventions/documenting_code.md
@@ -35,7 +35,7 @@ def horns
 end
 ``````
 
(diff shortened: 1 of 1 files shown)


mverzilli:
@GetRektByMe thanks! The example is a bit silly, because all unicorns actually are supposed to have only one horn, like, by definition :P. But the idea is suppose you have a `Unicorn` class which has a `horns` instance method: what you really want to say is that it returns the number of horns `this` (as in this instance) unicorn has. Just a silly example, no big deal :).

PR #112, opened 2017-08-22
fixes #83, document then shorthand for case

### LINKED ISSUE #83: Constants chapter needs more
How does immutability work with constants?

Constants can invoke methods / expressions? So they can be assigned something at runtime, but cannot be re-assigned something?

Include more examples for this.

opensas:
Just my tiny two cents...

### DIFF
--- a/syntax_and_semantics/case.md
+++ b/syntax_and_semantics/case.md
@@ -71,6 +71,16 @@ elsif tmp.odd?
 end
 ```
 
+You may use `then` after the `when` condition to place the body on a single line.
+
+```crystal
+case exp
+when value1, value2 then do_something
+when value3         then do_something_else
+else                     do_another_thing
+end
+```
+
 Finally, you can omit the `case`'s value:
 
 ```crystal


syntax_and_semantics/case.md
```diff
@@ -71,6 +71,16 @@ elsif tmp.odd?
 end
 ```
 
+You may use `then` after the `when` condition. This is most frequently used to place the body of the `when` on a single line.
```
RX14: I'd say "You may use then after the when condition to place the body on a single line."

syntax_and_semantics/case.md
```diff
@@ -71,6 +71,16 @@ elsif tmp.odd?
...
+
+```crystal
+case a
+when 1, 2 then puts "a is one or two"
```
RX14: I'd rather you demonstrated returning values from the when body instead of causing side effects (printing). It's a much more common use of then.

syntax_and_semantics/case.md
```diff
@@ -71,6 +71,16 @@ elsif tmp.odd?
 end
 ```
 
+You may use `then` after the `when` condition to place the body on a single line.
+
+```crystal
+case exp
+when value1, value2   then do_something
```
RX14: Why is there extra spacing before the then?

opensas: Sorry, removed the extra space, left the alignment for improved readbility

PR #114, opened 2017-09-08
added instruction for issues faced while installing on debian distro

enfageorge:
The necessary instruction with brief explanations have been added to the doc for issues commonly seen during installation in debain distro.

### DIFF
--- a/installation/on_debian_and_ubuntu.md
+++ b/installation/on_debian_and_ubuntu.md
@@ -56,9 +56,6 @@ sudo apt-get install curl
 sudo apt-get install apt-transport-https
 ```
 
-* #####GPG Error that signatures can't be verified. Public key is not available.
-
-	The key to get  repository  is not in your existing list of trusted sources. So when you try to add an external key to the list, some systems might show this issue. Installation would still continue with your permission.
 
 * #####Dirmngr not found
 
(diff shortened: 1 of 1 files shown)


installation/on_debian_and_ubuntu.md
```diff
@@ -40,3 +40,37 @@ When a new Crystal version is released you can upgrade your system using:
...
+```
+
+* #####The method driver /usr/lib/apt/methods/https could not be found
+
+	Debian package installation mechanism does not support HTTPS for its transport by default.To use HTTPS downloads, we have to install apt-transport-https . You can install the same using the following command:
+```
+sudo apt-get install apt-transport-https
+```
+
+* #####GPG Error that signatures can't be verified. Public key is not available.
```
mverzilli: We can't recommend this in Crystal's documentation, it's asking people to trust any unsigned thing, which is a very bad practice. Have you tried the advice here? https://crystal-lang.org/docs/installation/on_debian_and_ubuntu.html#setup-repository

enfarose: I installed crystal in my system with the instructions from the official documentation but that error still showed up. apt-get update always returns with error of untrusted sources.

enfageorge:
Since the rest are just packages to be installed, those can be integrated to be installed before crystal is. This is no longer needed

PR #119, opened 2017-09-26
Macros Pitfalls

Ivo-Balbaert:
Hi,

In https://crystal-lang.org/docs/syntax_and_semantics/macros.html#pitfalls
the code:
{% begin %}
  case 42
  {% for klass in %w(Int32 String) %}
    when {{klass.id}}
      p "is {{klass}}"
  {% end %}
  end
{% end %}
does not compile: Syntax error in expanded macro: macro_82789136:5: unexpected token: Int32

When replacing with the following it compiles and runs:
{% begin %}
  case 42
  {% for klass in [Int32, String] %}
    when {{klass.id}}
      p "is {{klass}}"
  {% end %}
  end
{% end %}

giving as output: "is Int32"

Thanks!


asterite:
Because `%w(...)` creates an array of strings, and then when you insert it into the code it has the double quotes.

The PR seems way too big to add a comment for this, also :-)

mverzilli:
I'll be happy to review it all, but I might ask you to spin off smaller PRs to address individual issues (that is, if this PR is flawless we'll merge it, if not, we'll ask you to reduce it until it is :)).

bew:
@Ivo-Balbaert can you edit your message to format code blocks like:
````
```crystal
# some crystal code
puts "hello #{:world}"
```
````
So that your given crystal code will render beautifully render as crystal code with proper highlighting and font:
```cr
# some crystal code
puts "hello #{:world}"
```

mverzilli:
@Ivo-Balbaert: @bew just made me realize that you're trying to merge `master` to `github-3`, which is an old branch that I should remove. Can you open a new PR requesting us to merge to `master` instead? I guess that's why this PR seems huge.

Ivo-Balbaert:
Yes I know, sorry for the wrong format.

PR #120, opened 2017-09-26
Remove explicit `macro def` syntax explanation

bew:
Remove this syntax as https://github.com/crystal-lang/crystal/pull/5040 has just been merged.


mverzilli:
Nice catch :)

PR #122, opened 2017-09-27
Update SUMMARY.md

ekremkaraca:
listing recently added "On Linux using Linuxbrew" section...


asterite:
@ekremkaraca Just a quick question: can anything from `homebrew` be installed with `linuxbrew`?

ekremkaraca:
@asterite I've just tried for crystal package and worked for it. I'm not sure if there would be macos-only stuff inside homebrew.

PR #123, opened 2017-10-02
Update requiring_files.md

Ivo-Balbaert:


syntax_and_semantics/requiring_files.md
```diff
@@ -47,7 +47,7 @@ This relative is mostly used inside a project to refer to other files inside it.
 
 ```crystal
 # in spec/project_spec.cr
-require "../src/project"
+require "./spec_helper"
```
asterite: Why this change? The idea is that in specs you refer to code in `src` using `../src`.

Maybe we can require both files? But this is just an example to show relative requires.

straight-shoota: Above it says explicitly `It is also used to refer to code from specs:` (ok, *code* could also refer to the code in `spec_helper.cr`...). But there should at least be an example of `../src/project`.

bew: Instead of writing an example file `project_spec.cr` we could give the `spec_helper.cr` as example, which will actually have the relative require `../src/project`

Ivo-Balbaert: Indeed. I proposed this change because in every generated project_spec.cr the code is:
             require "./spec_helper". 
So spec_helper.cr is required, which says:
              require "spec"
              require "../src/project"

Perhaps mentioning both is the clearest. 
Otherwise it should say:
             # in spec/spec_helper.cr:
             require "../src/project"

instead of what it is now, which is not how the code is generated, and thus a bit misleading
             # in spec/project_spec.cr
             require "../src/project"

straight-shoota:
@Ivo-Balbaert Could you re-add `require "../src/project"`?

straight-shoota:
Merged in aa287494. I couldn't update the PR for some reason.

PR #125, opened 2017-10-07
Add section about class methods

straight-shoota:


syntax_and_semantics/class_methods.md
```diff
@@ -0,0 +1,34 @@
...
+When they're defined inside a class or module scope it is easier to use `self` instead of the class name.
+
+Class methods can also be defined by [extending a `Module`](modules.md#extend-self).
+
+A class method can be called under the same name as it was defined (`CaesarCipher.decrypt("HELLO")`).
+When called from within the same class or module scope the receiver can be `self` or implicit (like `encrypt(string)`).
+
+# Constructors
+
+Constructors are basically just normal class methods which [create a new instance of the class](new,_initialize_and_allocate.md).
```
RX14: I think this section is confusingly worded in general.

straight-shoota: Any suggestions? =)

mverzilli: I think if you basically just remove some modifiers from that sentence it'll be clearer :D. Basically just remove basically just:

"Constructors are normal class methods which [create a new instance of the class](new,_initialize_and_allocate.md). All classes in Crystal have at least one constructor called `new`, but they may also define other constructors with different names. "

syntax_and_semantics/new,_initialize_and_allocate.md
```diff
@@ -55,7 +55,7 @@ class Person
 end
 ```
 
-First, note the `self.new` notation. This means that the method belongs to the **class** `Person`, not to particular instances of that class. This is why we can do `Person.new`.
+First, note the `self.new` notation. This means is is a [class method](class_methods.md) and belongs to the **class** `Person`, not to particular instances of that class. This is why we can do `Person.new`.
```
mverzilli: "This means it is a [class method](class_methods.md) and it belongs (...)"

RX14: How about just "This is a class method (...)"

syntax_and_semantics/class_methods.md
straight-shoota: Thanks, that sounds much better! =)

"All classes" is technically not 100% correct because the `new` class method could be hidden as a private method, so it wouldn't be callable on that class. I prefixed the sentence with "By default" maybe that's better?

syntax_and_semantics/class_methods.md
```diff
@@ -0,0 +1,33 @@
...
+    string.chars.map{ |char| ((char.upcase.ord - 52) % 26 + 65).chr }.join
+  end
+end
+
+CaesarCipher.encrypt("HELLO") # => "URYYB"
+```
+
+Class methods are defined by prefixing the method name with the type name and a period.
+
+```
```
mverzilli: Any reason not to use syntax highlighting on this code block? It looks strange, especially given the block above does highlight.

straight-shoota: No reason, just didn't make it into my fingers 😄

mverzilli:
Thank you @straight-shoota!

PR #126, opened 2017-10-08
Update macros.md

Ivo-Balbaert:


mverzilli:
Thank you @Ivo-Balbaert!

PR #129, opened 2017-10-09
Add a guide about testing

### LINKED ISSUE #111: anything about spec?
I found that there is no content about spec, I have no idea about what spec is, is it like something about test module? Maybe it is better to have something about that. :)

straight-shoota:
Closes #111

### DIFF
--- a/guides/testing.md
+++ b/guides/testing.md
@@ -1,8 +1,8 @@
 # Testing Crystal Code
 
-Crystal comes with a full-feature spec library in the [`Spec` module](https://crystal-lang.org/api/latest/Spec.html). It provides the structure for writing executable examples of how your code should behave.
+Crystal comes with a fully-featured spec library in the [`Spec` module](https://crystal-lang.org/api/latest/Spec.html). It provides a structure for writing executable examples of how your code should behave.
 
-Inspired by [Rspec](http://rspec.info/) it includes a domain specific language (DSL) that allows to write examples in a way similar to plain english..
+Inspired by [Rspec](http://rspec.info/), it includes a domain specific language (DSL) that allows you to write examples in a way similar to plain english.
 
 A basic spec looks something like this:
 
@@ -17,11 +17,11 @@ describe Array do
   end
 
   describe "#empty?" do
-    it "is empty when no elements are in the array" do
+    it "is true when no elements are in the array" do
       ([] of Int32).empty?.should be_true
     end
 
-    it "is not empty if there are elements in the array" do
+    it "is false if there are elements in the array" do
       [1].empty?.should be_false
     end
   end
@@ -66,8 +66,8 @@ actual.should be_nil          # passes if actual.nil?
 ```crystal
 actual.should be_true         # passes if actual == true
 actual.should be_false        # passes if actual == false
-actual.should be_truthy       # passes if actual is truthy (neither nil nor false)
-actual.should be_falsey       # passes if actual is falsey (nil or false)
+actual.should be_truthy       # passes if actual is truthy (neither nil nor false nor Pointer.null)
+actual.should be_falsey       # passes if actual is falsey (nil, false or Pointer.null)
 ```
 
 ### Comparisons
--- a/guides/testing.md
+++ b/guides/testing.md
@@ -1,6 +1,6 @@
 # Testing Crystal Code
 
-Crystal comes with a full-feature spec library in the [`Spec`module](https://crystal-lang.org/api/latest/Spec.html). It provides the structure for writing executable examples of how your code should behave.
+Crystal comes with a full-feature spec library in the [`Spec` module](https://crystal-lang.org/api/latest/Spec.html). It provides the structure for writing executable examples of how your code should behave.
 
 Inspired by [Rspec](http://rspec.info/) it includes a domain specific language (DSL) that allows to write examples in a way similar to plain english..
 
@@ -91,13 +91,6 @@ actual.should match(expected)           # passes if actual =~ expected
 These matchers run a block and pass if it raises a certain exception.
 
 ```crystal
-expect_raises do
-  # Passes if this block raises any exception.
-  #
-  # NOTE: This matcher should be avoided because it rescues and thus
-  # hides *any* unexpected exception raised by the example block.
-end
-
 expect_raises(MyError) do
   # Passes if this block raises an exception of type MyError.
 end
--- a/guides/testing.md
+++ b/guides/testing.md
@@ -1,0 +1,106 @@
+# Testing Crystal Code
+
+Crystal comes with a full-feature spec library to write test cases in the [`Spec`module](https://crystal-lang.org/api/latest/Spec.html).
+
+It is inspired by [Rspec](https://relishapp.com/rspec).
+
+## Anatomy of a spec file
+
+To use the spec module and DSL, you need to add `require "spec"` to your spec files. Many projects use a custom [spec helper](#spec-helper) which organizes these includes.
+
+Test files are structured through `describe` and `context` sections. Typically a top level `describe` defines the outer unit (such as a class) to be tested by the spec. Further `describe` sections can be nested within the outer unit to specify smaller units under test (such as individual methods).
+
+For unit tests, it is recommended to follow the conventions for method names: Outer `describe` is the name of the class, inner `describe` targets methods. Instance methods are prefixed with `#`, class methods with `.`.
+
+To establish certain contexts - think *empty array* versus *array with elements* - the `context` method may be used to communicate this to the reader. It behaves exactly like `describe`.
+
+`describe` and `context` take a description as argument (which should usually be a string) and a block containing the individual specs or nested groupings.
+
+Concrete test cases are defined in `it` blocks. An optional (but strongly recommended) descriptive string states it's purpose and a block contains the main logic performing the test.
+
+Test cases that have been defined or outlined but are not yet expected to work can be defined as `pending` instead of `it`. They will not be run but show up in the spec report as pending.
+
+Specs then use the `should` or `should_not` method on actual to verify that the value matches the expectation provided as argument.
+
+### Expectations
+
+Expectations define if a actual value matches a certain value or specific criteria.
+
+| Matcher        | Description |
+| --             | -- |
+| `eq(value)`    | passes if actual equals *value* (`==`) |
+| `be(value)`    | passes if actual and *value* are identical (`.same?`) |
+| `be_nil`       | passes if actual is nil (`== nil`) |
+| `be_true`      | passes if actual is true (`== true`) |
+| `be_false`     | passes if actual is false (`== false`) |
+| `be_truthy`    | passes if actual is truthy (neither `nil` nor `false`) |
+| `be_falsey`    | passes if actual is falsy (`nil` or `false`) |
+| `be.<(value)`  | passes if actual is lesser than *value* (`<`) |
+| `be.<=(value)` | passes if actual is lesser than or equal *value* (`<=`) |
+| `be.>(value)`  | passes if actual is greater than *value* (`>`) |
+| `be.>=(value)` | passes if actual is greater than or equal *value* (`>=`) |
+| `be_close(expected, delta)` | passes if actual is within *delta* of *expected* |
+| `contain`      | passes if actual includes *expected* (`.includes?`) |
+| `match(value)` | passes if actual matches *value* (`=~`) |
+| `be_a(type)`   | passes if actual is of type *type* (`is_a?`) |
+| `expect_raises(&block)` | passes if the block raises an exception |
+| `expect_raises(klass, &block)` | passes if the block raises an exception of type *klass* |
+| `expect_raises(klass, message, &block)` | passes if the block raises an exception of type *klass* and the error message contains *message* |
+
+## Running specs
+
+All specs of a project are compiled and executed through the command `crystal spec`.
+
+By convention, specs live in the `spec/` directory of a project. Spec files must end with `_spec.cr` to be recognizable as such by the compiler command.
+
+You can compile and run specs from folder trees, individual files or specific lines in a file.
+
+Run  all specs in files matching `spec/**/*_spec.cr`:
+```
+crystal spec
+```
+
+Run all specs in files matching `spec/my/test/**/*_spec.cr`:
+```
+crystal spec spec/my/test/
+```
+
+Run all specs in file `spec/my/test/file_spec.cr`:
+```
+crystal spec spec/my/test/file_spec.cr
+```
+
+Run the spec defined in line `14` of file `spec/my/test/file_spec.cr`:
+```
+crystal spec spec/my/test/file_spec.cr:14
+```
+
+The default formatter outputs the file and line style command for failing specs which makes it easy to rerun just this individual spec.
+
+## Spec helper
+
+Many projects use a custom spec helper file, usually named `spec/spec_helper.cr`.
+
+This file is used to require `spec` and other includes like code from the project needed for every spec file. This is also a good place to define global helper methods that make writing specs easier and avoid code duplication.
+
+```crystal
+# spec/spec_helper.cr
+require "spec"
+require "../src/my_project.cr"
+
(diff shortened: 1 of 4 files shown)


guides/testing.md
```diff
@@ -0,0 +1,106 @@
+# Testing Crystal Code
+
+Crystal comes with a full-feature spec library to write test cases in the [`Spec`module](https://crystal-lang.org/api/latest/Spec.html).
+
+It is inspired by [Rspec](https://relishapp.com/rspec).
```
RX14: Maybe a better link would be http://rspec.info/?

guides/testing.md
```diff
@@ -0,0 +1,106 @@
...
+
+## Anatomy of a spec file
+
+To use the spec module and DSL, you need to add `require "spec"` to your spec files. Many projects use a custom [spec helper](#spec-helper) which organizes these includes.
+
+Test files are structured through `describe` and `context` sections. Typically a top level `describe` defines the outer unit (such as a class) to be tested by the spec. Further `describe` sections can be nested within the outer unit to specify smaller units under test (such as individual methods).
+
+For unit tests, it is recommended to follow the conventions for method names: Outer `describe` is the name of the class, inner `describe` targets methods. Instance methods are prefixed with `#`, class methods with `.`.
+
+To establish certain contexts - think *empty array* versus *array with elements* - the `context` method may be used to communicate this to the reader. It behaves exactly like `describe`.
```
RX14: `` It behaves exactly like `describe`, but has a different name. `` perhaps?

guides/testing.md
```diff
@@ -0,0 +1,106 @@
...
+
+Crystal comes with a full-feature spec library to write test cases in the [`Spec`module](https://crystal-lang.org/api/latest/Spec.html).
+
+It is inspired by [Rspec](https://relishapp.com/rspec).
+
+## Anatomy of a spec file
+
+To use the spec module and DSL, you need to add `require "spec"` to your spec files. Many projects use a custom [spec helper](#spec-helper) which organizes these includes.
+
+Test files are structured through `describe` and `context` sections. Typically a top level `describe` defines the outer unit (such as a class) to be tested by the spec. Further `describe` sections can be nested within the outer unit to specify smaller units under test (such as individual methods).
```
RX14: Perhaps note that these sections are only for humans, and are used only in the spec output and when reading the spec source? I'm not sure if it's worth saying.

guides/testing.md
```diff
@@ -0,0 +1,106 @@
...
+
+For unit tests, it is recommended to follow the conventions for method names: Outer `describe` is the name of the class, inner `describe` targets methods. Instance methods are prefixed with `#`, class methods with `.`.
+
+To establish certain contexts - think *empty array* versus *array with elements* - the `context` method may be used to communicate this to the reader. It behaves exactly like `describe`.
+
+`describe` and `context` take a description as argument (which should usually be a string) and a block containing the individual specs or nested groupings.
+
+Concrete test cases are defined in `it` blocks. An optional (but strongly recommended) descriptive string states it's purpose and a block contains the main logic performing the test.
+
+Test cases that have been defined or outlined but are not yet expected to work can be defined as `pending` instead of `it`. They will not be run but show up in the spec report as pending.
```
RX14: Perhaps "defined using" is a bit clearer here.

guides/testing.md
```diff
@@ -0,0 +1,106 @@
...
+
+Specs then use the `should` or `should_not` method on actual to verify that the value matches the expectation provided as argument.
```
RX14: "the value being tested" instead of "actual"?

guides/testing.md
```diff
@@ -0,0 +1,106 @@
...
+| Matcher        | Description |
+| --             | -- |
+| `eq(value)`    | passes if actual equals *value* (`==`) |
+| `be(value)`    | passes if actual and *value* are identical (`.same?`) |
+| `be_nil`       | passes if actual is nil (`== nil`) |
+| `be_true`      | passes if actual is true (`== true`) |
+| `be_false`     | passes if actual is false (`== false`) |
+| `be_truthy`    | passes if actual is truthy (neither `nil` nor `false`) |
+| `be_falsey`    | passes if actual is falsy (`nil` or `false`) |
+| `be.<(value)`  | passes if actual is lesser than *value* (`<`) |
```
RX14: I think these would be better formatted as `be < value`, which should work.

guides/testing.md
straight-shoota: I'm not sure about the benefit of elaborating this.
It's also relevant for running specs, you can select only a specific describe/context section to be run.

guides/testing.md
```diff
@@ -1,8 +1,32 @@
...
 
-It is inspired by [Rspec](http://rspec.info/).
+Inspired by [Rspec](http://rspec.info/) it includes a domain specific language (DSL) that allows to write examples in a way similar to plain english..
+
+A basic spec looks something like this:
+
+```crystal
+require "spec"
+
+describe "Array" do
```
bew: I think we could mention it as `describe Array do ...` (without the quotes), as it's usually used like that!

guides/testing.md
```diff
@@ -20,60 +44,92 @@ Concrete test cases are defined in `it` blocks. An optional (but strongly recomm
...
-| `expect_raises(klass, message, &block)` | runs the block and passes if it raises an exception of type *klass* and the error message contains *message* |
+### Equivalence, Identity and Type
+```crystal
+actual.should eq(expected)    # passes if actual == expected
+actual.should be(expected)    # passes if actual.same?(expected)
+actual.should be_a(expected)  # passes if actual.is_a?(expected)
+actual.should be_nil          # passes if actual.nil?
+```
+
+### Truthiness
```
bew: As this section is not _only_ about truthiness, maybe we could named it `Thruthiness, Falseiness` (not sure it should be spelled though)

guides/testing.md
```diff
@@ -20,60 +44,92 @@ Concrete test cases are defined in `it` blocks. An optional (but strongly recomm
...
+actual.should be <  expected  # passes if actual <  expected
+actual.should be <= expected  # passes if actual <= expected
+actual.should be >  expected  # passes if actual >  expected
+actual.should be >= expected  # passes if actual >= expected
+```
+
+### Other matchers
+```crystal
+actual.should be_close(expected, delta) # passes if actual is within delta of expected:
+                                        # (actual - expected).abs <= delta
```
bew: I think the example would be easier to grasp if written as: `expected - delta <= actual <= expected + delta`

guides/testing.md
```diff
@@ -20,60 +44,92 @@ Concrete test cases are defined in `it` blocks. An optional (but strongly recomm
...
+actual.should contain(expected)         # passes if actual.includes?(expected)
+actual.should match(expected)           # passes if actual ==~ expected
```
bew: `==~` is not an operator (or I don't know how to use it), I think it should be `=~`.

guides/testing.md
```diff
@@ -20,60 +44,92 @@ Concrete test cases are defined in `it` blocks. An optional (but strongly recomm
...
+```
+
+### Expecting errors
+
+These matchers run a block and pass if it raises a certain exception.
+
+```crystal
+expect_raises do
+  # passes if this block raises any exception
+  # NOTE: This should be avoided because it can hide errors.
```
bew: Explain why it can hide errors? (=> It will catch ALL kind of exceptions)

guides/testing.md
```diff
@@ -20,60 +44,92 @@ Concrete test cases are defined in `it` blocks. An optional (but strongly recomm
...
+
+expect_raises(MyError) do
+  # passes if this block raises an exception of type MyError
+end
+
+expect_raises(MyError, "error message") do
+  # passes if this block raises an exception of type MyError
+  # and the error message contains "error message"
+end
+```
```
bew: There is also `expect_raises(MyError, /message/)` (with a regex instead of a string)

guides/testing.md
straight-shoota: I's possible to use a constant, but there is no real gain from it. This is different in rspec, where the first argument to `describe` is - in case of a class - available as `described_class` and an instance of that class is created as [implicit subject](https://relishapp.com/rspec/rspec-core/v/3-0/docs/subject/implicitly-defined-subject).

In Crystal, it is just a a descriptor, the constant get's converted to `String` anyway. So, I'm not sure it should if we should promote this.

guides/testing.md
straight-shoota: I'm sure *truthiness* can be used as generic term for both ([example](https://learn.co/lessons/truthiness-in-ruby-readme))

guides/testing.md
straight-shoota: All examples show the exact logic used in the matchers... I don't really think it makes much sense to change this.

guides/testing.md
straight-shoota: I'll copy the full note proposed for the methods API doc.

guides/testing.md
bew: I agree, but I still think that we should encourage using it as a Type, without quotes.

It's not really a _problem_ to have a typo or a mistake in an `it` description, but when it's a type description, I would like to have an error if the dev wrote `Arary` instead of `Array` (writing `"Arary"` with quotes won't give an error, and could be mis-leading).

We could say the same for the methods, where we usually write `describe "#empty?"` (with quotes), but as there's currently no way to check for the method existance, I think we can leave that out.

guides/testing.md
RX14: I find this version much easier to understand, I read it as "the absolute difference between the actual value and the expected value should be less than delta".

guides/testing.md
```diff
@@ -0,0 +1,170 @@
+# Testing Crystal Code
+
+Crystal comes with a full-feature spec library in the [`Spec`module](https://crystal-lang.org/api/latest/Spec.html). It provides the structure for writing executable examples of how your code should behave.
```
asterite: It's not full-feature. There's no `setup` and `teardown`. That's why I think documenting this right now is not good. Maybe the build-in test library needs to change to have these features. But I doubt that can be done with `describe` and `it`. So maybe we'll have to switch to something like unit test or minitest? I don't know. But that's why I'd prefer not to document this yet.

straight-shoota:
@asterite If the test library is replaced at some point (whenever that might be), this will most certainly mean some changes to the syntax and features described in this guide. But it is unlikely to change the general function and concepts of testing.

At that point it would need to be updated, sure. But  for *now* (and that might be a long time) there should be some documentation on testing Crystal at it's current state. I think we can all agree that testing is one - if not *the* - most important part of writing software. Yet, there is little documentation on how to this in Crystal. There is [an article on ruby-journal.com](http://ruby-journal.com/crystal/testing-whirlwind-tour-in-crystal-lang/)) but nothing official.

I wouldn't mind having a guide here even if it *might* need some changes later on.
Otherwise we could think about enhancing the API docs (see crystal-lang/crystal#5092) or publishing this elsewhere.

guides/testing.md
straight-shoota: Okay, let's not call it `full-feature` (though that certainly depends).

RX14: @asterite I strongly disagree btw, there's nothing that the spec library can't do currently which isn't better expressed using helper methods. The spec harness currently is fantastic.

RX14: Typo: no space in `` `Spec`module ``.

guides/testing.md
```diff
@@ -0,0 +1,170 @@
...
+actual.should contain(expected)         # passes if actual.includes?(expected)
+actual.should match(expected)           # passes if actual =~ expected
+```
+
+### Expecting errors
+
+These matchers run a block and pass if it raises a certain exception.
+
+```crystal
+expect_raises do
```
RX14: If this overload was removed in latest crystal shouldn't it not be mentioned?

robacarp:
I like this idea, this particular sentence was useful to me today as I was diagnosing why specs weren't running.
 
> By convention, specs live in the `spec/` directory of a project. Spec files must end with `_spec.cr` to be recognizable as such by the compiler command.

Thanks for the helpful guide.

guides/testing.md
```diff
@@ -0,0 +1,163 @@
+# Testing Crystal Code
+
+Crystal comes with a full-feature spec library in the [`Spec` module](https://crystal-lang.org/api/latest/Spec.html). It provides the structure for writing executable examples of how your code should behave.
```
RX14: > fully-featured

> provides a structure

guides/testing.md
```diff
@@ -0,0 +1,163 @@
...
+
+Inspired by [Rspec](http://rspec.info/) it includes a domain specific language (DSL) that allows to write examples in a way similar to plain english..
```
RX14: A comma after `Rspec` please.

> allows you

two full stops at the end of the sentence

guides/testing.md
```diff
@@ -0,0 +1,163 @@
...
+
+describe Array do
+  describe "#size" do
+    it "correctly reports the number of elements in the Array" do
+      [1, 2, 3].size.should eq 3
+    end
+  end
+
+  describe "#empty?" do
+    it "is empty when no elements are in the array" do
```
RX14: I'd use "is true" and "is false" instead of "is empty" and "is not empty" because we are describing a method, not array. This means it reads "Array #empty? is true when no elements are in the array" instead of "Array #empty? is empty when no elements are in the array".

guides/testing.md
```diff
@@ -0,0 +1,163 @@
...
+    end
+  end
+end
+```
+
+## Anatomy of a spec file
+
+To use the spec module and DSL, you need to add `require "spec"` to your spec files. Many projects use a custom [spec helper](#spec-helper) which organizes these includes.
+
+In test files, specs are structured in example groups enclosed in `describe` and `context` sections. Typically a top level `describe` defines the outer unit (such as a class) to be tested by the spec. Further `describe` sections can be nested within the outer unit to specify smaller units under test (such as individual methods).
```
RX14: Do you mean "specs are structured by example groups which are defined by `describe` and `context`? I'm not sure entirely how to interpret this paragraph. The core problem is that you haven't defined what an "example group" is.

guides/testing.md
```diff
@@ -0,0 +1,163 @@
...
+Each example can contain multiple expectations, but it should test only one specific behaviour.
+
+## Expectations
+
+Expectations define if the value being tested (*actual*) matches a certain value or specific criteria.
+
+### Equivalence, Identity and Type
+```crystal
+actual.should eq(expected)    # passes if actual == expected
+actual.should be(expected)    # passes if actual.same?(expected)
```
RX14: There should be a note on the difference between `==` and `same?`, and that `same?` isn't applicable on structs.

guides/testing.md
```diff
@@ -0,0 +1,163 @@
...
+actual.should be_a(expected)  # passes if actual.is_a?(expected)
+actual.should be_nil          # passes if actual.nil?
+```
+
+### Truthiness
+```crystal
+actual.should be_true         # passes if actual == true
+actual.should be_false        # passes if actual == false
+actual.should be_truthy       # passes if actual is truthy (neither nil nor false)
```
RX14: Technically `Pointer.null` is falsey too. I'm not sure if it's worth mentioning here.

guides/testing.md
straight-shoota: Yes, I'll take your rephrased version. I don't know if example groups need to be explained more detailed as *a group of examples*.

RX14: I'd rather not mention "example groups" at all. I'd introduce `it` before `describe`/`context` to get the term "example" into the reader's head associated with `it` and then say:

> In test files, examples are structured in groups defined by `describe` and `context` blocks. [...]

or something similar.

guides/testing.md
straight-shoota: I guess it makes sense. If matching values are listed, it should be complete.

guides/testing.md
straight-shoota: Sounds good. I think I had actually intended to put the `it` description first but somehow missed to reorder this.

mverzilli:
Thank you @straight-shoota!

PR #132, opened 2017-10-17
Clarify Comment Page

HCLarsen:
The character used for comments is commonly referred to as the 'number sign,' not the 'sharp character' as was previously written.

### DIFF
--- a/syntax_and_semantics/comments.md
+++ b/syntax_and_semantics/comments.md
@@ -1,6 +1,6 @@
 # Comments
 
-Comments start with the number (`#`) sign. Only one-line comments are supported for now.
+Comments start with the (`#`) character. Only one-line comments are supported for now.
 
 ```crystal
 # This is a comment
--- a/syntax_and_semantics/comments.md
+++ b/syntax_and_semantics/comments.md
@@ -1,6 +1,6 @@
 # Comments
 
-Comments start with the sharp (`#`) character. Only one-line comments are supported for now.
+Comments start with the number (`#`) sign. Only one-line comments are supported for now.
 
 ```crystal
 # This is a comment
--- a/syntax_and_semantics/comments.md
+++ b/syntax_and_semantics/comments.md
@@ -1,6 +1,6 @@
 # Comments
 
-Comments start with the (`#`) character. Only one-line comments are supported for now.
+Comments start with the `#` character. Only one-line comments are supported for now.
 
 ```crystal
 # This is a comment


syntax_and_semantics/comments.md
```diff
@@ -1,6 +1,6 @@
 # Comments
 
-Comments start with the sharp (`#`) character. Only one-line comments are supported for now.
+Comments start with the number (`#`) sign. Only one-line comments are supported for now.
```
straight-shoota: I think it would even be better to move the `#` behind "number sign" instead of having it hanging strangely in between.

mverzilli: Why not just "Comments start with `#`. [...]"

straight-shoota: For searchability it's good to write the name of the character, because searching for `#` is PITA.

mverzilli: I tried googling for "crystal comments start with the sharp" and it doesn't find it. Besides that, if you don't know how to comment Crystal code you search for "crystal how to comment", I've got some difficulties figuring out a scenario where I'd be searching for "crystal sharp" or "crystal number sign" and wanting to get to this page as a result.

straight-shoota: I don't have a convincing example either (maybe something like `number sign in programming language`) but that's not really the point. Is there any benefit from *not* including the name?

jhass: Agreed, why would I search for the character when I don't know how do to comments at all. Also there's way too many names for `#`, pound sign, hash, crosshatch, octothorpe (according to unicode anyway :D) ...

straight-shoota: If nothing else, it helps to clarify it is `#` and none of these similar looking - depending on the font - unicode characters `⋕⌗♯ⵌ井ꖛ﹟＃𐦕𝄰𝄱`.

jhass: People that confuse it with those kinda deserve the pain if you ask me ... :P

mverzilli: Quick poll: which of each of those chars is "number sign", which is "hash sign", which is "number sign"? I'll invite you a beer if all of us here get it right for all cases :D.

RX14: I vote for drop the name entirely. If I heard "number sign" I wouldn't think `#`, but I would think `#` for "sharp" or "hash". Since the name is likely to be contentious, lets just use the specific character.

HCLarsen: In 38 years of speaking English, I have NEVER heard that referred to as a "sharp" outside of music class. The terms "number sign" or "pound sign" are pretty universal, so that's why I put "number sign."

Also, the Oxford English dictionary agrees that it's the number sign.
https://en.oxforddictionaries.com/definition/number_sign

HCLarsen: However, if you'd all rather just say "# character," then I'll update it to that instead.

asterite: Maybe only musicians know it as sharp? :-)

RX14: I thought most programmers would know about C# being pronounced C sharp.

HCLarsen: I thought most programmers have enough attention to detail to notice that the C sharp logo is a sharp, not a number sign. Although the two are similar, they're not the same.

HCLarsen: Snarkiness aside, the musical sharp ♯ is Unicode 266F, while the number sign # is Unicode 0023. Can we put this to rest now and move on to the best wording for this page?

RX14: Lets just use the character with no name. Please.

HCLarsen: Can we get a vote on this? We've had a number of commenters here, so I'd like to make sure we've got a consensus before I update the PR.

HCLarsen: Thumbs up here if you want it to say:

Comments start with the (`#`) character.

HCLarsen: Or, if you want it to say:

Comments start with the number sign (`#`) .

RX14: @HCLarsen we already have 2 votes from the core team on not using a name, surely that's enough.

asterite:
Looks good, but I wonder if that parentheses is needed now. Instead of:

> Comments starts with the (`#`) character

maybe better:

> Comments starts with the `#` character

HCLarsen:
I could do that if you prefer it with the parentheses.

RX14:
I think without parentheses looks better here.

HCLarsen:
Ok, I should have that for you in a few hours.

mverzilli:
Thanks @HCLarsen !

HCLarsen:
No problem. Happy to help.

Issue #136, opened 2017-10-21
Document magic constants __FILE__ and __DIR__


RX14:

robacarp:
This can now be closed

mverzilli:
Thanks!

da99:
Thanks @RX14  for opening it.

PR #138, opened 2017-10-27
document magic constants

robacarp:
For issue #136 

I also added \_\_LINE__ but perhaps that is not needed. As always, comments and suggestions are welcome.

### DIFF
--- a/syntax_and_semantics/constants.md
+++ b/syntax_and_semantics/constants.md
@@ -29,12 +29,38 @@ end
 TEN #=> 10
 ```
 
-# Magic Constants
+# Psuedo Constants
 
-These constants are provided automatically by Crystal:
+Crystal provides a few psuedo-constants which provide reflective data about the source code being executed.
 
-Constant | Meaning
----------|--------
-`__FILE__` | The full path to the current file
-`__DIR__` | The full path to the directory where the current file is located
-`__LINE__` | The current line number in the file
+`__LINE__` is the current line number in the currently executing crystal file. When `__LINE__` is declared as the default value to a method parameter, it represents the line number at the location of the method call.
+
+`__END_LINE__` is the line number of the `end` of the calling block. Can only be used as a default value to a method parameter.
+
+`__FILE__` references the full path to the currently executing crystal file.
+
+`__DIR__` references the full path to the directory where the currently executing crystal file is located.
+
+```crystal
+# Assuming this example code is saved at: /crystal_code/psuedo_constants.cr
+#
+def psuedo_constants(caller_line = __LINE__, end_of_caller = __END_LINE__)
+  puts "Called from line number: #{caller_line}"
+  puts "Currently at line number: #{__LINE__}"
+  puts "End of caller block is at: #{end_of_caller}"
+  puts "File path is: #{__FILE__}"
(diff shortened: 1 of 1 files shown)


RX14:
`__LINE__` has a special meaning when used as a default value to a function argument: https://carc.in/#/r/2z97.

I think using a long-form paragraph with examples is going to be a better bet than a table here.

mverzilli:
I agree that a long-form paragraph is better, but the PR adds some valuable info that was missing and maybe we can iterate on this. One thing that I don't like about this is the "Magic" word. Can we move away from that language? Maybe "Predefined constants"?

asterite:
Don't forget about `__END_LINE__`

jhass:
Predefined doesn't cut it for me since they depend on context. How about special purpose constants or pseudo constants (since they're not that constant after all)?

mverzilli:
Context (or contextual) constants sounds great to me!

robacarp:
I rewrote it in paragraph form, and provided an example. I agree this is a better format, good suggestion @RX14. 

@mverzilli, I agree about "Magic" and I think Psuedo Constants conveys the meaning that they're not actually constant better than Contextual Constants. It's also a little less of a mouthful to me. 

Here is the rendered update:

image

robacarp:
Should any of these be documented in this section as well: `ARGF`, `ARGV`, `PROGRAM_NAME`, `STDERR`, `STDIN`, `STDOUT`?

asterite:
@robacarp All of those are "documented" in the API: https://crystal-lang.org/api/0.23.1/toplevel.html

What I mean is that they are there but lack proper documentation.

robacarp:
@asterite yeah, thats where I got the list. I understand the desire to draw the lines well between the std lib and the language reference. Maybe it make sense to have a link in the language reference to the current api docs for those constants?

asterite:
Maybe that makes sense for a tutorial, not for the language reference.

RX14:
I think crystal is really missing a advanced, long, tutorial which brings together the stdlib and language reference.

syntax_and_semantics/constants.md
```diff
@@ -28,3 +28,39 @@ end
 
 TEN #=> 10
 ```
+
+# Psuedo Constants
```
jhass: https://en.m.wiktionary.org/wiki/psuedo ☺

robacarp: hah, good catch

robacarp: I tried to think that I'd typo'd that. "You know how to spell," I told myself. 

And then I found all 6 times I'd typed that. 😅

mverzilli:
Thanks @robacarp!

da99:
Should there be a small mention how these magic constants are also available in macros? Some people coming to Crystal may not have been exposed to a decent macro system and might need reminding this is part of the subset of Crystal available for macros.

robacarp:
@da99 I had the same thought but didn't include anything because [there is a code sample in the macros section](https://crystal-lang.org/docs/syntax_and_semantics/macros.html#constants) which documents this.

da99:
@robacarp Thanks for pointing this out.

Issue #103, opened 2017-06-13
begin..rescue..else..ensure..end


Yay295:
`ensure` must come after `else`, but this is not specified on the [exception handling page](https://crystal-lang.org/docs/syntax_and_semantics/exception_handling.html). In fact, that page actually defines `ensure` before `else`.

robacarp:
This can now be closed

mverzilli:
Thanks!

PR #139, opened 2017-10-27
better document rescue/else/ensure ordering

robacarp:
relevant to #103 

I also toyed with the idea of explicitly stating that the `else` must come before the `ensure` but didn't find anything that pleased me, so I pulled it. 

Comments and suggestions welcome.

### DIFF
--- a/syntax_and_semantics/exception_handling.md
+++ b/syntax_and_semantics/exception_handling.md
@@ -105,6 +105,22 @@ rescue
 end
 ```
 
+## else
+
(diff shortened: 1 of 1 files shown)


RX14:
Perhaps we should include mentioning the new flat `rescue` etc. syntax for `do`/`end` too.

robacarp:
@RX14, it sounds like it's relevant, but I'm not quite up to speed on that. Browsing the backlog conversation on the crystal repo I came up with the changes I just pushed. Is that what you are talking about?

Issue #137, opened 2017-10-25
Integer Literals Should Give Value Ranges, not Bit Size


HCLarsen:
Right now it says "There are four signed integer types: Int8, Int16, Int32 and Int64, being able to represent numbers of 8, 16, 32 and 64 bits respectively."

I think it would be much better to show the range of values that can be represented with those classes. It makes for a far better explanation. I'm happy to write the revision myself, I just want to run it by the core team first.

HCLarsen:
Something like this is what I'm thinking:

There are four signed integer types: [Int8](http://crystal-lang.org/api/Int8.html)(, [Int16](http://crystal-lang.org/api/Int16.html), [Int32](http://crystal-lang.org/api/Int32.html) and [Int64](http://crystal-lang.org/api/Int64.html), being able to represent numbers in the ranges of −128 to 127, −32,768 to 32,767, −2,147,483,648 to 2,147,483,647 and −9,223,372,036,854,775,808 to 9,223,372,036,854,775,807 respectively.

bew:
I think that we should keep the bit size information, but give a hint about the possible value range.

HCLarsen:
The bit size can easily be inferred from the class name. After all, it's right there in the class name. The value range on the other hand, not really. Now maybe some very experienced developers have those ranges memorized, but documentation needs to cater to people of all experience levels. 

When I was young, learning C as my first compiled language, the documentation there did show the difference between an Int and a Long Int, and that was invaluable to me learning how to use them.

HCLarsen:
I'm not against a table, but with the large numbers in the higher ranges, it doesn't get a bit unruly. Take a look:

Length  | Signed | Signed Values | Unsigned | Unsigned Values
 ---------- | -----------: | -----------: |-----------: | -----------: |
8       |   Int8   | -128-127 | UInt8 | 0-255
16       |   Int8   | −32,768-32,767-32,767| UInt8 | 0-65,535
8       |   Int8   | −2,147,483,648-2,147,483,647-2,147,483,647 | UInt8 | 0-4,294,967,295
8       |   Int8   | −9,223,372,036,854,775,808-9,223,372,036,854,775,807 | UInt8 | 0-18,446,744,073,709,551,615

HCLarsen:
Unless we separate the signed and unsigned integers into separate rows or tables.

RX14:
Yes of course we should have each crystal type be one row, not one bit length ( especially when you're saying but length is fairly unimportant)

[...]

PR #140, opened 2017-10-28
Include value range in Int description

HCLarsen:
Solution to issue #137 .

### DIFF
--- a/syntax_and_semantics/comments.md
+++ b/syntax_and_semantics/comments.md
@@ -1,6 +1,6 @@
 # Comments
 
-Comments start with the number (`#`) sign. Only one-line comments are supported for now.
+Comments start with the (`#`) character. Only one-line comments are supported for now.
 
 ```crystal
 # This is a comment
--- a/syntax_and_semantics/comments.md
+++ b/syntax_and_semantics/comments.md
@@ -1,6 +1,6 @@
 # Comments
 
-Comments start with the sharp (`#`) character. Only one-line comments are supported for now.
+Comments start with the number (`#`) sign. Only one-line comments are supported for now.
 
 ```crystal
 # This is a comment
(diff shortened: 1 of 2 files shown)


syntax_and_semantics/literals/integers.md
```diff
@@ -1,8 +1,17 @@
 # Integers
 
-There are four signed integer types: [Int8](http://crystal-lang.org/api/Int8.html), [Int16](http://crystal-lang.org/api/Int16.html), [Int32](http://crystal-lang.org/api/Int32.html) and [Int64](http://crystal-lang.org/api/Int64.html), being able to represent numbers of 8, 16, 32 and 64 bits respectively.
-
-There are four unsigned integer types: [UInt8](http://crystal-lang.org/api/UInt8.html), [UInt16](http://crystal-lang.org/api/UInt16.html), [UInt32](http://crystal-lang.org/api/UInt32.html) and [UInt64](http://crystal-lang.org/api/UInt64.html).
+There are four signed integer types, and four unsigned integer types:
+
+Class | Length  | Minimum Value | Maximum Value
```
mverzilli: Change column `Class` to `Type`. Ints are actually Crystal structs, but at this point in the docs it'd be confusing to point that out.

syntax_and_semantics/comments.md
```diff
@@ -1,6 +1,6 @@
 # Comments
```
mverzilli: Would you remove this file from the PR? Let's keep this PR focused on the integer specs table.

HCLarsen: I don't know how that got in there. I didn't make any changes to it since the previous PR.

mverzilli: I guess it's because you have both things in master. You should create a branch from crystal-lang/master for each PR to avoid this. The 3 commits about comments should be in their own branch, and the 2 commits relevant to this PR should be in their own branch. 

I don't know if it's possible to modify a PR so that it targets a different branch, so you'll probably have to close this PR and open a new one.

PR #141, opened 2017-10-28
Add value range to Ints

HCLarsen:
Solves issue #137.

### DIFF
--- a/syntax_and_semantics/literals/integers.md
+++ b/syntax_and_semantics/literals/integers.md
@@ -1,8 +1,17 @@
 # Integers
 
-There are four signed integer types: [Int8](http://crystal-lang.org/api/Int8.html), [Int16](http://crystal-lang.org/api/Int16.html), [Int32](http://crystal-lang.org/api/Int32.html) and [Int64](http://crystal-lang.org/api/Int64.html), being able to represent numbers of 8, 16, 32 and 64 bits respectively.
+There are four signed integer types, and four unsigned integer types:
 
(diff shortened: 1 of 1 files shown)


syntax_and_semantics/literals/integers.md
```diff
@@ -1,8 +1,17 @@
 # Integers
 
-There are four signed integer types: [Int8](http://crystal-lang.org/api/Int8.html), [Int16](http://crystal-lang.org/api/Int16.html), [Int32](http://crystal-lang.org/api/Int32.html) and [Int64](http://crystal-lang.org/api/Int64.html), being able to represent numbers of 8, 16, 32 and 64 bits respectively.
-
-There are four unsigned integer types: [UInt8](http://crystal-lang.org/api/UInt8.html), [UInt16](http://crystal-lang.org/api/UInt16.html), [UInt32](http://crystal-lang.org/api/UInt32.html) and [UInt64](http://crystal-lang.org/api/UInt64.html).
+There are four signed integer types, and four unsigned integer types:
+
+Class | Length  | Minimum Value | Maximum Value
```
mverzilli: Almost there! `Class` => `Type`. Thanks for your patience! 🙏

HCLarsen: Oh right, in the reset, I forgot about that.

HCLarsen: Actually, thank you for your patience.

HCLarsen:
128bit types? You're planning that soon?

RX14:
It's already merged in master (but they're really unstable so we probably shouldn't document them here until after the next release where they've been tested on multiple platforms)

HCLarsen:
That works for me. I'll update the table after that release.

asterite:
Please no. They are still experimental and might be removed in the future.

HCLarsen:
Then let's leave any discussions about 128 bit integers for another day.

RX14:
@asterite I think they'll be fine to document after a short while. I don't believe we'll have to remove them.

PR #144, opened 2017-11-08
Fix Typo in Char Description

HCLarsen:
Found a small typo in the Char file. It had "an" instead of "a" for the letter *u*.


HCLarsen:
I just found the same typo in the String file. Should I update this PR to fix them both in one go?

RX14:
@HCLarsen yes

mverzilli:
Thank you!

PR #145, opened 2017-11-08
Add gitbook-plugin-offline

lwakefield:
This PR adds the `gitbook-plugin-offline` which uses service workers to make the content available offline.


RX14:
If we're using `gitbook install` we shouldn't add it to the `package.json`, same for the `ga` plugin. Shall we remove the `package.json` entirely?

straight-shoota:
Using `gitbook install` is recommended for this plugin. So you don't need to add it as a dev dependency in `package.json`.

I'm not sure if we should or could remove `package.json` entirely though. because it also contains project metadata.

package.json
```diff
@@ -4,6 +4,7 @@
   "license": "CC0",
   "repository": "github:crystal-lang/crystal-book",
   "devDependencies": {
-    "gitbook-plugin-ga": "^1.0.1"
+    "gitbook-plugin-ga": "^1.0.1",
+    "gitbook-plugin-offline": "0.0.3"
```
straight-shoota: No need to add the plugin here, `gitbook install` will take care of this.

straight-shoota: No need to add the plugin here, `gitbook install` will take care of this.

lwakefield: Addressed in cb159c8

PR #147, opened 2017-11-13
Add instructions for profiling on Linux

nome:
The documentation on profiling just referenced the generic gprof documentation. Figuring out how to apply this to Crystal took a bit of searching and guesswork, which is not an ideal experience for new users.

### DIFF
--- a/guides/performance.md
+++ b/guides/performance.md
@@ -10,7 +10,7 @@ Donald Knuth once said:
 
 However, if you are writing a program and you realize that writing a semantically equivalent, faster version involves just minor changes, you shouldn't miss that opportunity.
 
-And always be sure to profile your program to learn what its bottlenecks are. For profiling, on Mac OSX you can use [Instruments Time Profiler](https://developer.apple.com/library/prerelease/content/documentation/DeveloperTools/Conceptual/InstrumentsUserGuide/Instrument-TimeProfiler.html), which comes with XCode. On Linux, a program that can profile C/C++ programs, like [gprof](https://sourceware.org/binutils/docs/gprof/), should work.
+And always be sure to profile your program to learn what its bottlenecks are. For profiling, on Mac OSX you can use [Instruments Time Profiler](https://developer.apple.com/library/prerelease/content/documentation/DeveloperTools/Conceptual/InstrumentsUserGuide/Instrument-TimeProfiler.html), which comes with XCode. On Linux, any program that can profile C/C++ programs should work. To profile your program with [gprof](https://sourceware.org/binutils/docs/gprof/), build it with `crystal build --release --link-flags -pg` and run it as usual. This will produce a file `gmon.out` in the current directory that can be analyzed with `gprof ./my-program ./gmon.out`.
 
 Make sure to always profile programs by compiling or running them with the `--release` flag, which turns on optimizations.
 


guides/performance.md
```diff
@@ -10,7 +10,7 @@ Donald Knuth once said:
...
 
-And always be sure to profile your program to learn what its bottlenecks are. For profiling, on Mac OSX you can use [Instruments Time Profiler](https://developer.apple.com/library/prerelease/content/documentation/DeveloperTools/Conceptual/InstrumentsUserGuide/Instrument-TimeProfiler.html), which comes with XCode. On Linux, a program that can profile C/C++ programs, like [gprof](https://sourceware.org/binutils/docs/gprof/), should work.
+And always be sure to profile your program to learn what its bottlenecks are. For profiling, on Mac OSX you can use [Instruments Time Profiler](https://developer.apple.com/library/prerelease/content/documentation/DeveloperTools/Conceptual/InstrumentsUserGuide/Instrument-TimeProfiler.html), which comes with XCode. On Linux, any program that can profile C/C++ programs should work. To profile your program with [gprof](https://sourceware.org/binutils/docs/gprof/), build it with `crystal build --release -D -pg --link-flags -pg` and run it as usual. This will produce a file `gmon.out` in the current directory that can be analyzed with `gprof ./my-program ./gmon.out`.
```
bew: What is `-D -pg` for?

nome: @bew According to [the gprof docs](https://sourceware.org/binutils/docs/gprof/Compiling.html#Compiling), the `-pg` option has to be given both as a compiler and as a linker option. `-D -pg` tells Crystal to apply it as a compiler flag.

nome:
Incidentally, the build cache apparently doesn't take these flags into account (i.e. after doing a regular build of a file, one either has to change the input file or clear the cache in order to get a profiling build). Not sure whether this behaviour should be documented or changed in the compiler.

guides/performance.md
RX14: That's incorrect. `-pg` is a flag for the C compiler. The crystal compiler has no analog and it should be left out. I'm not sure the linker flag does anything either, without the corresponding `-pg` compiler flag.

`--release --debug` are the flags I use when debugging, and produce adequate debug information for both `perf` and callgrind. I haven't tried gprof, but it's probably the same. Perhaps better gprof-specific profiling info could be generated, but i've never felt wanting when using callgrind/perf.

RX14: I've just tested it out, seems the linker `-pg` flag causes the program to be wrapped in such a way that it produces a `gmon.out`, but obviously as none of the object files have been compiled with gprof data, the file is empty and useless.

A much more useful tutorial would guide the user through how to use perf and callgrind. gprof will likely never work with crystal because it requires compiler modifications, while perf and callgrind work on the already-existing generic DWARF debug data built into the program.

nome: I do get some profiling data with the procedure I described, provided that the program in question is sufficiently non-trivial. If it just contains some `puts` or `sleep`, I also get an empty `gmon.out`. However, after some more testing, it seems the `-D -pg` option indeed doesn't do anything. I've fixed this.

I'll also try out `perf` and `callgrind`. Obviously, if these are the preferred profiling tools for Crystal, they should be documented instead of `gprof`.

RX14: I agree that the current docs are wrong too. I assume these docs were written by people at manas, who use OSX mainly. `gprof` is the wrong tool for the job here.

nome:
The changes requested end up making a rather different modification to the docs, so I created a new PR (#148). Obviously, the idea is that only one of the PRs gets merged.

PR #148, opened 2017-11-14
Suggest perf/callgrind for profiling on Linux

nome:
gprof requires compiling code with an extra flag, which is only indirectly supported by crystal. Also, the gmon.out file produced sometimes ends up being empty. perf and callgrind are usually both more straightforward to use for new users and more powerful.

This PR is intended as an alternative to #147 (see discussion there for background).


mverzilli:
Thank you @nome!

PR #152, opened 2017-11-23
Clarify require docs and change libs to lib, fixes #151

### LINKED ISSUE #151: Require documentation should use `lib` instead of `libs`
The documentation on requiring files need to be changed to `lib` instead of `libs`.

I have a push request for this in the works that also clarifies how to get `require` to use the right path since it took me a while to get it right.

For reference, see crystal-lang/crystal#321 and crystal-lang/crystal/pull/2788.

GloverDonovan:
This fixes #151 and clarifies the importance of the working directory when requiring files.

### DIFF
--- a/syntax_and_semantics/requiring_files.md
+++ b/syntax_and_semantics/requiring_files.md
@@ -10,7 +10,7 @@ Once a file is required, the compiler remembers its absolute path and later `req
 
 This looks up "filename" in the require path.
 
-By default the require path is the location of the standard library that comes with the compiler, and the "libs" directory relative to the current working directory (given by `pwd` in a Unix shell). These are the only places that are looked up.
+By default the require path is the location of the standard library that comes with the compiler, and the "lib" directory relative to the current working directory (given by `pwd` in a Unix shell). These are the only places that are looked up.
 
 The lookup goes like this:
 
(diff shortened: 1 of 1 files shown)


syntax_and_semantics/requiring_files.md
```diff
@@ -33,6 +33,8 @@ The second rule is very convenient because of the typical directory structure of
     - project_spec.cr
 ```
 
+For example, if you want to `require "foo"` in "project.cr", then you should run crystal in the `project` directory with `crystal src/project.cr`. This ensures that the `lib` folder is one of the require paths.
```
straight-shoota: This sounds to bulky, and it should better describe *how* it behaves if used as intended and how you break it.
```
For example, if you put `require "foo"` in `project.cr` and run `crystal src/project.cr` in the
project's root directory, it will find `foo` in `lib/foo/foo.cr`.
If you run the compiler from somewhere else, say the `src` folder,
`lib` will not be in the path and `require "foo"` can't be resolved.
```

GloverDonovan: I agree with your suggestion and will update my pull request with the new text. Hope everything is better now!

mverzilli:
Thank you @GloverDonovan!

PR #156, opened 2017-12-13
Add examples of capturing blocks

marksiemers:


marksiemers:
I may have added too many examples, what do you think?

mverzilli:
Hi! Thanks for this! I agree, I think you can reduce the number of examples by showing just one example of each case (without combining them): one for any return type, one for a specific return type, one with two arguments.

Also, it'd get much more readable if you put each example in its own code snippet and describe them outside the snippet instead of using comments.

marksiemers:
@mverzilli - I made an update, adding sections and explanations for each code example. Let me know if you still think it is too much volume. In-line comments about extraneous examples would be extra helpful.

Thanks.

syntax_and_semantics/capturing_blocks.md
```diff
@@ -2,7 +2,9 @@
 
 A block can be captured and turned into a `Proc`, which represents a block of code with an associated context: the closured data.
 
-To capture a block you must specify it as a method's block argument, give it a name and specify the input and output types. For example:
+To capture a block you must specify it as a method's block argument, and return that block from the method.
```
straight-shoota: A method doesn't have to return a block to capture it. It can be captured even if it is not used outside the method.
You should keep the mention of input and output types.

syntax_and_semantics/capturing_blocks.md
```diff
@@ -37,29 +41,91 @@ model.save # prints "Saved!"
 
 In the above example the type of `&block` wasn't specified: this just means that the captured block doesn't have arguments and doesn't return anything.
 
-Note that if the return type is not specified, nothing gets returned from the proc call:
+## Options for capturing blocks
```
straight-shoota: Better: Type restrictions for blocks

syntax_and_semantics/capturing_blocks.md
```diff
@@ -37,29 +41,91 @@ model.save # prints "Saved!"
...
+
+Capturing is not limited to requiring one block argument and a specific return type.
+A block can be captured in many ways. A captured block can:
```
straight-shoota: I think these two lines should be rephrased, because they are not really precise and easy to understand. These type restrictions do not only apply for capturing blocks, they also apply for non-captured blocks (the only difference is that with non-captured you can omit them).
So I'm not sure if this entire section fits here.

syntax_and_semantics/capturing_blocks.md
marksiemers: @straight-shoota - Understood about the "return that block" statement being incorrect.
The statement "specify the input and output types" - that reads to me that you can't do any of these:
`Int32 ->` - no output
`-> String` - no input
`Int32 -> _` - generic output

It may need to stay, but I think it needs a rewording, what do you think of:
> ... specify the inputs and outputs. You can also specify that there are no inputs and/or there is no output. Input types must be specific, the output can be specific or generic.

syntax_and_semantics/capturing_blocks.md
marksiemers: Is there a better spot for it?

If it belongs somewhere else, I would want a link to it from this page - so you can figure out the different ways that are permissible to capture blocks from this page (even if through a link)

syntax_and_semantics/capturing_blocks.md
marksiemers: @straight-shoota - Would this be a true statement?
> To capture a block you must specify it as a method's block argument, then store it in a variable or return it from the method.

If it isn't stored in a variable (instance, class, or global) and it isn't returned from the method - is there any other way to "capture" it?

straight-shoota: Hm. Specifying input and output types IMHO also includes specifying none.
But I understand that the original wording might be confusing. Maybe it's better to use `type declaration` instead of `type`? Making it clear that it doesn't have to be a literal type but you can also declare that there is none. For example with your last suggestion: `... specify it as a method's block argument including input and output type declarations`.

I would leave out the part `or return it from the method` because to be able to return it, you need to have it in a variable (including local) and therefore it get's captured.

In this example, the block is only inlined. Still, it technically is captured in the variable `block` and does qualify as a captured block, though it is not *used* as such:
```
def capture(&block : -> Int32)
  yield
end
```

syntax_and_semantics/capturing_blocks.md
straight-shoota: I guess there is no place for "blocks in general". So it probably might as well stay here...

syntax_and_semantics/capturing_blocks.md
asterite: No, that's not a captured block. A captured block is only one where it's mentioned in the body of the method.

marksiemers:
Updates made based on comments and also expanded some sections to make connections between stated requirements for capturing blocks and code examples.

marksiemers:
@straight-shoota @RX14 - LMK if any further changes are needed. Thanks.

syntax_and_semantics/capturing_blocks.md
```diff
@@ -2,7 +2,14 @@
 
 A block can be captured and turned into a `Proc`, which represents a block of code with an associated context: the closured data.
 
-To capture a block you must specify it as a method's block argument, give it a name and specify the input and output types. For example:
+To capture a block you must:
```
straight-shoota: I really dislike direct address in a formal specification. Could you rephrase this? For example here: `A block is captured if these conditions are met:` (list items should be in 3rd person)

syntax_and_semantics/capturing_blocks.md
```diff
@@ -15,7 +22,14 @@ proc.call(1) #=> 2
 
 The above code captures the block of code passed to `int_to_int` in the `block` variable, and returns it from the method. The type of `proc` is [Proc(Int32, Int32)](http://crystal-lang.org/api/Proc.html), a function that accepts a single `Int32` argument and returns an `Int32`.
 
-In this way a block can be saved as a callback:
+The simple example covers all requirements for capturing a block:
+- def int_to_int(__&block__ : Int32 -> Int32) - specify it as a method's block argument
```
straight-shoota: Code shoud be wrapped in backticks and you shouldn't need to highlight specific parts.

syntax_and_semantics/capturing_blocks.md
marksiemers: LMK what you think of the new wording.

marksiemers:
@straight-shoota @RX14 - Any further changes?

syntax_and_semantics/capturing_blocks.md
```diff
@@ -37,29 +51,101 @@ model.save # prints "Saved!"
...
+
+Note that if the return type is not specified, `nil` is returned from the proc call:
+
+```crystal
+def some_proc(&block : Int32 ->)
+  block
+end
+
+proc = some_proc { |x| x + 1 }
+proc.call(1) # nil
```
straight-shoota: There's a `=>` missing in the comment

marksiemers:
@straight-shoota - Thanks for your patience on this and detailed reviews.

How does everything look now?

marksiemers:
@RX14 - If you don't have any issues with the PR, could you update your review?

RX14:
This is just overly verbose and confusing. It seems to me that the type syntax for `Proc`s and block capturing are being conflated. All you need to do is link to the already-documented `Proc` type grammar, and document that `_` can be used for multiple return types. Then mention you need to use `block` in the method body for the block to be captured - otherwise you're just documenting the type of `yield`. Then you don't need an example for every single possible permuation of syntax.

Regarding `_`: can we just remove `def foo(&block : String -> _)` entirely and replace it with `def foo(&block : String -> T) forall T`? The latter works *right now*, and `_` isn't generic since: https://carc.in/#/r/3i62

cc @asterite

syntax_and_semantics/capturing_blocks.md
```diff
@@ -2,20 +2,34 @@
 
 A block can be captured and turned into a `Proc`, which represents a block of code with an associated context: the closured data.
 
-To capture a block you must specify it as a method's block argument, give it a name and specify the input and output types. For example:
+A block is captured if these conditions are met:
+- The block is specified as a method's block argument (e.g. `def method(&block)`)
+- The input(s) and output are specified (e.g. `def method(&block : Int32 -> Int32)`)
```
asterite: This isn't necessary, just the first and third points are true.

marksiemers:
Yeah, it seems `alias` makes the difference, `_` works with all these types: https://carc.in/#/r/3i6e

syntax_and_semantics/capturing_blocks.md
```diff
@@ -2,20 +2,34 @@
...
-proc.call(1) #=> 2
+proc.call(1) # => 2
 ```
 
 The above code captures the block of code passed to `int_to_int` in the `block` variable, and returns it from the method. The type of `proc` is [Proc(Int32, Int32)](http://crystal-lang.org/api/Proc.html), a function that accepts a single `Int32` argument and returns an `Int32`.
 
-In this way a block can be saved as a callback:
+The simple example covers all requirements for capturing a block:
+- `def int_to_int(&block : Int32 -> Int32)` - specify it as a method's block argument
+- `def int_to_int(&block : Int32 -> Int32)` - specify the input(s) and output
```
asterite: As I said, this isn't needed. If you do:

```crystal
def foo(&block)
  block
end
```

that defines a method that captures a block that has no inputs and returns `nil`.

syntax_and_semantics/capturing_blocks.md
```diff
@@ -37,29 +51,101 @@ model.save # prints "Saved!"
...
+  - Many arguments: `Int32, String, Array ->`
+- Return a specific type
+  - With no inputs: `-> Int32`, `-> String`
+  - With input(s): `Int32 -> String`
+- Return anything
+  - With no inputs: `-> _`
+  - With input(s): `String -> _`
+- Return nothing (returns `nil`)
+  - With no inputs: `->` (the `->` can be omitted - `def method(&block)` )
+  - With input(s): `Int32 ->`
```
asterite: This is too verbose. Just saying "Zero or many inputs can be specified, separates by comma. One output can be specified: if it's omitted, it means `Nil`".

syntax_and_semantics/capturing_blocks.md
marksiemers: "requirements" is probably the wrong word, but I want to show that all those things are happening either explicitly or implicitly when a block is captured.

In your example, the "no inputs" and `nil` return are implied. So I need to find a better way to express that in the docs.

RX14:
@asterite but why have `_` when it can always be replaced with `T` and `forall T`? Isn't it just more code for more confusion for no reason? I personally think that unifying every kind of "type templating" using `forall` is a good idea.

asterite:
Yeah, it's something from the past. But also `_` usually means "anything, I don't care" and it's shorter than writing `forall`.

marksiemers:
I can make the bulleted lists more concise and incorporate the comments, should the code examples near the bottom be gutted as well?

FWIW, I like the `_` syntax - it doesn't get more concise and it seems intuitive to me.

It has the same reason to be around as the unnecessary block shorthand: `array.map &.downcase` instead of `array.map { |str| str.downcase }`

PR #157, opened 2017-12-18
Improve subtitle phrasing

lukeholder:
The 2 subtitles in their original form seem like declarative instructive statements that contradict each other, rather than being just 2 different ways of doing something.

### DIFF
--- a/syntax_and_semantics/type_inference.md
+++ b/syntax_and_semantics/type_inference.md
@@ -23,7 +23,7 @@ For this reason, Crystal needs to know, in an obvious way (as obvious as to a hu
 
 There are several ways to let Crystal know this.
 
-## Using an explicit type annotation
+## With type annotations
 
 The easiest, but probably most tedious, way is to use explicit type annotations.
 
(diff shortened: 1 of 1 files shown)


RX14:
I actually prefer the old wording.

mverzilli:
It does sound better to me in -ing form, but I'm not a native speaker...

straight-shoota:
To me it currently sounds like an instruction like "you shouldn't use an explicit type annotation!"

But I think the wording can be further improved in general. "use/-ing" in these section titles does not make a lot of sense, at least it doesn't really add any value. 
We could just shorten the first title to `Explicit type annotation` and a more descriptive variant for the second would be `Implicit type annotation`.

mverzilli:
But "implicit annotation" is a contradiction in terms, annotations are always explicit... unless maybe you annotate with invisible ink :D. Sorry for the bad joke!

straight-shoota:
True, that should be `Implicit type`.

`annotation` was supposed to be written with invisible ink ;)

straight-shoota:
Or any other suggestion...

mverzilli:
How about `With type annotations` and `Without type annotations`?

RX14:
Yup, I vote for what @mverzilli said.

straight-shoota:
I'd say the type *is* annotated, just not explicitly.

But It's fine by me.

lukeholder:
@RX14 I also agree with @mverzilli's suggestion. Should I go ahead and amend PR? @straight-shoota ?

PR #163, opened 2017-12-19
Fix errors in string literal

straight-shoota:
It feels like #160 was merged way too fast... This fixes some mistakes that have been unnoticed.

I don't know... it's nice to get merged right away but maybe it would be better to give such PR's a few days time before merging (even if two core members approve) to give more time for others to review?

### DIFF
--- a/syntax_and_semantics/literals/string.md
+++ b/syntax_and_semantics/literals/string.md
@@ -10,7 +10,7 @@ A String is typically created with a string literal enclosing UTF-8 characters i
 
 ## Escaping
 
-A backslash denotes a special character inside a string, which can either be a named escape sequence or a numerical representation of a unicode character.
+A backslash denotes a special character inside a string, which can either be a named escape sequence or a numerical representation of a unicode codepoint.
 
 Available escape sequences:
 ```crystal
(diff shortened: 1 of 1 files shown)


mverzilli:
I understand the suggestion, but that would be a project tracking nightmare. It naturally happens that way for big stuff, design decisions, breaking changes, etc. But things like #160 are clearly an enhancement, are small and are self contained. Ensuring two core team members approve of the change is more than enough, otherwise we'll let even more things fall through the cracks.

If you want to treat a PR as WIP for which you want comments, I suggest you state that in the title or description.

straight-shoota:
@mverzilli The fixup commits should have been squashed before merging. I'm not sure if there is any policy about that for this repo... I could have done it myself  but I think it's easier for reviewers to follow the progress of the PR to have individual commits and only squash before merge.

RX14:
@straight-shoota completely agree. I was thinking of making a bot which handled all this when I got CI all sorted out...

mverzilli:
I usually prefer to squash before merging as you suggest, but in this case the commits were self-contained and made sense semantically, that's why I chose to leave them like that.

PR #164, opened 2017-12-19
Overhaul Char literal docs

straight-shoota:


syntax_and_semantics/literals/char.md
```diff
@@ -13,25 +12,27 @@ It is created by enclosing an UTF-8 character in single quotes.
...
 '\\' # backslash
+'\b' # backspace
 '\e' # escape
 '\f' # form feed
 '\n' # newline
 '\r' # carriage return
 '\t' # tab
 '\v' # vertical tab
+'\uNNNN' # hexadecimal unicode character
+'\u{NNNN...}' # hexadecimal unicode character
```
bew: what is `...` for? I can understand its use for strings, where you can chain unicode chars like `\u{12 13 14}` But I don't think it's useful for the `CHar` doc.

straight-shoota: True, it should perhaps be described differently. For char literals, it can still be one to six digits. So maybe `\u{N+}` to mimic some sort of regex quantifier?

PR #167, opened 2017-12-19
Build PRs

matiasgarciaisaia:
- Fixed missing plugins during build
- Build PRs but only deploy branches

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,10 +6,7 @@ before_install: |
   npm install -g gitbook-cli@2.3.0
   npm install
   gitbook install
-script: test $TRAVIS_PULL_REQUEST != "false" || gitbook build --gitbook=3.2.2
-branches:
-  only:
-    - master
+script: gitbook build --gitbook=3.2.2
 deploy:
   provider: s3
   access_key_id: AKIAIRL4BWFPN7P54TBQ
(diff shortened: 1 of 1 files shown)


matiasgarciaisaia:
@bcardiff you won't probably recall, but was there any good reason we weren't building PRs?

.travis.yml
```diff
@@ -18,3 +16,6 @@ deploy:
   local_dir: _book
   upload-dir: docs
   skip_cleanup: true
+  on:
+    branch: master
+    condition: "$TRAVIS_PULL_REQUEST = true"
```
RX14: why? I'm confused.

RX14:
Isn't the correct fix to add `gitbook-plugin-edit-link` to our `package.json` similar to the `ga` plugin?

straight-shoota:
I've just created #168 for adding `gitbook install`. I think this should better be in a separate PR because it is not enabling builds for pull requests.

straight-shoota:
@RX14 The recommended installation method for the [`edit-link` plugin](https://plugins.gitbook.com/plugin/edit-link) is to run `gitbook install`. For [`ga`](https://plugins.gitbook.com/plugin/ga) it's using `npm install gitbook-plugin-ga` which is equivalent to adding it to `package.json`. I don't know the difference between both methods.

.travis.yml
matiasgarciaisaia: Well spotted. I've mixed the previous build condition with the `||` operator - it should deploy when PR is `false` 👍

RX14:
If that's the case then we should remove the package.json. Personally I prefer to use 1 package manager not 2.

RX14 (approved):

PR #169, opened 2017-12-19
Overhaul symbol literal docs

straight-shoota:
Add some explanations about symbol literals.

### DIFF
--- a/syntax_and_semantics/literals/symbol.md
+++ b/syntax_and_semantics/literals/symbol.md
@@ -12,7 +12,7 @@ Symbols are interpreted at compile time and cannot be created dynamically. The o
 
 A double-quoted identifier can contain any unicode character including white spaces and accepts the same escape sequences as a [string literal](./string.html), yet no interpolation.
 
-For an unquoted identifier the same naming rules apply as for methids. It can contain alphanumeric characters, underscore (`_`) or characters with a code point greater than `159`(`0x9F`). It must not start with a number and may end with an exclamation mark (`!`) or question mark (`?`).
+For an unquoted identifier the same naming rules apply as for methods. It can contain alphanumeric characters, underscore (`_`) or characters with a code point greater than `159`(`0x9F`). It must not start with a number and may end with an exclamation mark (`!`) or question mark (`?`).
 
 ```crystal
 :question?
--- a/syntax_and_semantics/literals/symbol.md
+++ b/syntax_and_semantics/literals/symbol.md
@@ -1,6 +1,14 @@
 # Symbol
 
-A [Symbol](http://crystal-lang.org/api/Symbol.html) is a constant that is identified by a name without you having to give it a numeric value.
+A [Symbol](http://crystal-lang.org/api/Symbol.html) represents a unique name inside the entire source code.
+It is internally implemented as a constant with a numeric value of type `Int32`.
+
+Symbols are interpreted at compile time and cannot be created dynamically. The only way to create a Symbol is by using a symbol literal, denoted by a colon (`:`) followed by an identifier. The identifier may be enclosed in double quotes (`"`).
+
+An unquoted identifier can contain alphanumeric characters, underscore (`_`) or characters with a code point greater than `159`(`0x9F`). It must not start with a number and may end with an exclamation mark (`!`) or question mark (`?`).
+
+A double-quoted identifier can contain any unicode character including white spaces and accepts the same escape sequences as a [string literal](./string.html), yet no interpolation.
+
 
 ```crystal
 :hello
@@ -41,5 +49,3 @@ A [Symbol](http://crystal-lang.org/api/Symbol.html) is a constant that is identi
 :<=>
 :===
 ```
-
-Internally a symbol is represented as an `Int32`.
--- a/syntax_and_semantics/literals/symbol.md
+++ b/syntax_and_semantics/literals/symbol.md
@@ -49,3 +49,15 @@ A double-quoted identifier can contain any unicode character including white spa
 :<=>
 :===
 ```
+
+## Percent symbol array literal
+
+Besides the single symbol literal, there is also a percent literal to create an [Array](https://crystal-lang.org/api/Array.html) of symbols. It is indicated by `%i` and a pair of delimiters. Valid delimiters are parenthesis `()`, square brackets `[]`, curly braces `{}`, angles `<>` and pipes `||`. Except for the pipes, all delimiters can be nested meaning a start delimiter inside the string escapes the next end delimiter.
+
+Identifiers may contain any unicode characters. Individual symbols are separated by a single space character (` `) which must be escaped to use it as a part of an identifier.
+
+```crystal
+%i(foo bar baz) # => [:foo, :bar, :baz]
+%i(foo\nbar baz) # => [:"foo\nbar", :baz]
+%i(foo\ bar baz) # => [:"foo bar", :baz]
+```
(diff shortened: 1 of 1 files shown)


straight-shoota:
The section about percent literals describes how it *should* be (IMHO) but the compiler currently does not support nested delimters and escaped spaces (see crystal-lang/crystal#5418).

bew:
I think you should add more examples, like for the quoted symbols like `:"abc"` for example

straight-shoota:
The third example demonstrates a quoted symbol. But I'll look over it again tomorrow.
I'm not to keen of including examples of *invalid code*.

I also noticed that the description of unquoted literal misses a mention of the operator symbols.

bew:
> The third example demonstrates a quoted symbol

It looks far away from the description, I was thinking about a very simple example showing the 2 forms, right after the initial description, like:
```cr
:unquoted_symbol
:"quoted symbol"
```

syntax_and_semantics/literals/symbol.md
```diff
@@ -1,45 +1,62 @@
 # Symbol
 
-A [Symbol](http://crystal-lang.org/api/Symbol.html) is a constant that is identified by a name without you having to give it a numeric value.
+A [Symbol](http://crystal-lang.org/api/Symbol.html) represents a unique name inside the entire source code.
+It is internally implemented as a constant with a numeric value of type `Int32`.
```
bcardiff: I think that going into implementation detail before any sample is not the best order.

syntax_and_semantics/literals/symbol.md
```diff
@@ -1,45 +1,62 @@
...
+
+Symbols are interpreted at compile time and cannot be created dynamically. The only way to create a Symbol is by using a symbol literal, denoted by a colon (`:`) followed by an identifier. The identifier may be enclosed in double quotes (`"`).
 
 ```crystal
-:hello
-:good_bye
+:unquoted_symbol
+:"quoted symbol"
```
bcardiff: Maybe showing that `:a == :"a"` is worth to mention here.

syntax_and_semantics/literals/symbol.md
```diff
@@ -1,45 +1,62 @@
...
+```
+
+A double-quoted identifier can contain any unicode character including white spaces and accepts the same escape sequences as a [string literal](./string.html), yet no interpolation.
 
-# With spaces and symbols
-:"symbol with spaces"
+An unquoted identifier can contain alphanumeric characters, underscore (`_`) or characters with a code point greater than `159`(`0x9F`). It must not start with a number and may end with an exclamation mark (`!`) or question mark (`?`).
```
bcardiff: I would mention that these are the same rule for naming methods. (either before or after the detailed explanation)

syntax_and_semantics/literals/symbol.md
```diff
@@ -1,45 +1,62 @@
...
+:=~
 ```
 
-Internally a symbol is represented as an `Int32`.
+## Percent symbol array literal
+
+Besides the single symbol literal, there is also a percent literal to create an [Array](https://crystal-lang.org/api/Array.html) of symbols. It is indicated by `%i` and a pair of delimiters. Valid delimiters are parenthesis `()`, square brackets `[]`, curly braces `{}`, angles `<>` and pipes `||`. Except for the pipes, all delimiters can be nested meaning a start delimiter inside the string escapes the next end delimiter.
+
+Identifiers may contain any unicode characters. Individual symbols are separated by a single space character (` `) which must be escaped to use it as a part of an identifier.
+
```
bcardiff: I would split the escaping example after the initial ones.

And also (before or after the escaping example) I would explain the nested usage with a sample. This way it kind of follows the same pattern as before.

syntax_and_semantics/literals/symbol.md
```diff
@@ -1,45 +1,68 @@
...
+Internally, symbols are implemented as constants with a numeric value of type `Int32`.
+
+## Percent symbol array literal
+
+Besides the single symbol literal, there is also a percent literal to create an [Array](https://crystal-lang.org/api/Array.html) of symbols. It is indicated by `%i` and a pair of delimiters. Valid delimiters are parenthesis `()`, square brackets `[]`, curly braces `{}`, angles `<>` and pipes `||`. Except for the pipes, all delimiters can be nested meaning a start delimiter inside the string escapes the next end delimiter.
+
+```crystal
+%i(foo bar baz) # => [:foo, :bar, :baz]
+%i(foo\nbar baz) # => [:"foo\nbar", :baz]
+%i(foo(bar) baz) # => [:"foo(bar)", :baz]
```
bcardiff: I double check what happen with `%i[a b c(d e)]` and `%i(a b c(d e))` to see how the nesting is working with spaces and different symbols. From the proposed description I thought that was a corner case.

`%i[a b c(d e)] # => [:a, :b, :"c(d", :"e)"]` as expected since the enclosing delimiter and the inner one are not the same, but

`%i(a b c(d e)) => unexpected token: )`

The example `%i(foo(bar) baz)` also complains with `unexpected token: baz`.

Regarding `%i(foo\nbar baz)` the value of the expression is actually `[:"foo\\nbar", :baz]`

I checked with Crystal 0.24.1.

syntax_and_semantics/literals/symbol.md
```diff
@@ -1,45 +1,68 @@
...
 ```
 
-Internally a symbol is represented as an `Int32`.
+Identifiers may contain any unicode characters. Individual symbols are separated by a single space character (` `) which must be escaped to use it as a part of an identifier.
+
+```crystal
+%i(foo\ bar baz) # => [:"foo bar", :baz]
```
bcardiff: ` %i(foo\ bar baz) # => [:"foo\\", :bar, :baz]`

Again, checked with Crystal 0.24.1.

straight-shoota:
@bcardiff This describes how it *should* be, but the compiler actually has some issues with parsing array literals, see my second comment in this thread and crystal-lang/crystal#5418

straight-shoota:
I've opened crystal-lang/crystal#5667 to fix nested delimters and escaped whitespace. Let's discuss the intended behavior there and come back here once that's settled.

syntax_and_semantics/literals/symbol.md
```diff
@@ -1,45 +1,68 @@
...
+:unquoted_symbol
+:"quoted symbol"
+:"a" # identical to :a
+```
+
+A double-quoted identifier can contain any unicode character including white spaces and accepts the same escape sequences as a [string literal](./string.html), yet no interpolation.
 
-# With spaces and symbols
-:"symbol with spaces"
+For an unquoted identifier the same naming rules apply as for methods. It can contain alphanumeric characters, underscore (`_`) or characters with a code point greater than `159`(`0x9F`). It must not start with a number and may end with an exclamation mark (`!`) or question mark (`?`).
```
straight-shoota: crystal/crystal-lang#5956

straight-shoota: Fixed in crystal-lang/crystal#5969

straight-shoota:
Can this get a second approval?

syntax_and_semantics/literals/symbol.md
straight-shoota: These are now all fixed in master.

syntax_and_semantics/literals/symbol.md
straight-shoota: This works in master.

PR #170, opened 2017-12-20
Overhaul Array literal docs

straight-shoota:


syntax_and_semantics/literals/array.md
```diff
@@ -1,67 +1,82 @@
...
+```crystal
+Array{1, 2, 3}
+```
 
-You can use a special array literal syntax with other types too, as long as they define an argless `new` method and a `<<` method:
+This literal can be used with any type as long as it has an argless constructor and responds to `<<`.
 
 ```crystal
-MyType{1, 2, 3}
+IO::Memory{1, 2, 3}
```
RX14: Not sure this is a good example... Not sure there's a better one though.

syntax_and_semantics/literals/array.md
```diff
@@ -1,67 +1,82 @@
...
 ```crystal
-tmp = MyType.new
-tmp << 1
-tmp << 2
-tmp << 3
-tmp
+array_like = IO::Memory.new
+array_like << 1
+array_like << 2
+array_like << 3
```
RX14: How about just use `IO::Memory.new << 1 << 2 << 3`

syntax_and_semantics/literals/array.md
```diff
@@ -1,67 +1,82 @@
...
 ```crystal
-tmp = MyType(typeof(1, 2, 3)).new
-tmp << 1
-tmp << 2
-tmp << 3
-tmp
+array_like = Set(typeof(1, 2, 3)).new
+array_like << 1
+array_like << 2
+array_like << 3
```
RX14: ditto

syntax_and_semantics/literals/array.md
straight-shoota: Agreed. I think it is better to use a type from standard lib instead of an unspecified custom type.
There aren't much options, though, espacially non-generic ones.

syntax_and_semantics/literals/array.md
straight-shoota: I considered that, but I think it's preferable to use the extensive style to make it easier. After all, the language reference should be understandable for beginners and not everyone might be familiar with using operators in method chains.

syntax_and_semantics/literals/array.md
```diff
@@ -1,6 +1,6 @@
 # Array
 
-An [Array](http://crystal-lang.org/api/Array.html) is a generic, ordered and integer-indexed collection of elements of a specifc type `T`.
+An [Array](http://crystal-lang.org/api/Array.html) is a ordered and integer-indexed generic collection of elements of a specifc type `T`.
```
bew: Is a ordered => is an ordered ?

syntax_and_semantics/literals/array.md
```diff
@@ -1,67 +1,82 @@
 # Array
 
-An [Array](http://crystal-lang.org/api/Array.html) is a generic type containing elements of a type `T`. It is typically created with an array literal:
+An [Array](http://crystal-lang.org/api/Array.html) is an ordered and integer-indexed generic collection of elements of a specifc type `T`.
```
mverzilli: typo: "specifc" => "specific"

mverzilli:
Thank you @straight-shoota!

PR #177, opened 2017-12-27
Improve cross-compilation documentation.

barisbalic:
I'm proposing some improvements to the pages on cross-compilation, based on my recent experience (and lack of experience), I think these changes will make it easier for others.  Happy to refactor to feedback.

### DIFF
--- a/syntax_and_semantics/cross-compilation.md
+++ b/syntax_and_semantics/cross-compilation.md
@@ -2,7 +2,20 @@
 
 Crystal supports a basic form of [cross compilation](http://en.wikipedia.org/wiki/Cross_compiler).
 
-In order to achieve this, the compiler executable provides two flags:
+## Preparation
+
+Cross-compilation will depend on `libcrystal` being present on the target system, so we need to build it by:
+* Downloading Crystal source code on the target machine.
+* Installing the LLVM/compiler toolchain. (Eg. `sudo apt-get install clang` on Ubuntu)
+* Installing the [required libraries](https://github.com/crystal-lang/crystal/wiki/All-required-libraries) for building Crystal.
+* Running `make deps` in the source directory.
+
+Now when we compile Crystal programs on the target system we can refer to `crystal/src/ext/libcrystal.a`.
+
+
+## Compiling
+
+The compiler executable provides two flags:
 
 * `--cross-compile`: When given enables cross compilation mode
 * `--target`: the [LLVM Target Triple](http://llvm.org/docs/LangRef.html#target-triple) to use and set the default [compile-time flags](compile_time_flags.html) from
@@ -20,10 +33,10 @@ crystal build your_program.cr --cross-compile --target "x86_64-unknown-linux-gnu
 This will generate a `.o` ([Object file](http://en.wikipedia.org/wiki/Object_file)) and will print a line with a command to execute on the system we are trying to cross-compile to. For example:
 
 ```bash
-cc your_program.o -o your_program -lpcre -lrt -lm -lgc -lunwind
+cc your_program.o -o your_program -lpcre -lgc /local/crystal/src/ext/libcrystal.a
 ```
 
-You must copy this `.o` file to that system and execute those commands. Once you do this the executable will be available in that target system.
+You must copy this `.o` file to the target system and execute the command above, replacing the local `libcrystal` reference with file we built in the preparation step.  Once you do this the executable will be available in that target system.
 
 This procedure is usually done with the compiler itself to port it to new platforms where a compiler is not yet available. Because in order to compile a Crystal compiler we need an older Crystal compiler, the only two ways to generate a compiler for a system where there isn't a compiler yet are:
 * We checkout the latest version of the compiler written in Ruby, and from that compiler we compile the next versions until the current one.
@@ -32,3 +45,17 @@ This procedure is usually done with the compiler itself to port it to new platfo
 The first alternative is long and cumbersome, while the second one is much easier.
 
 Cross-compiling can be done for other executables, but its main target is the compiler. If Crystal isn't available in some system you can try cross-compiling it there.
+
+
+## Troubleshooting libraries
+
+The cross-compilation step will generate a command with numerous flags, many of these refer to the libraries that need to be included when compiling your program.
+If the libraries are not present on the target system, the compilation will fail and complain about the respective flags, which are not always descriptive.
+
+When you need to work out what libraries a flag refers to you can do the following:
+
+```bash
+ld -lgc --verbose
+```
+
+The output will give you more descriptive names (and even paths) for the libraries that cannot be loaded.  You can then install these and try compiling your program again.
--- a/syntax_and_semantics/cross-compilation.md
+++ b/syntax_and_semantics/cross-compilation.md
@@ -4,13 +4,13 @@ Crystal supports a basic form of [cross compilation](http://en.wikipedia.org/wik
 
 ## Preparation
 
-Cross-compilation will depend on `libcrystal` being present on the target system, so we need to build it by:
-* Downloading Crystal source code on the target machine.
-* Installing the LLVM/compiler toolchain. (Eg. `sudo apt-get install clang` on Ubuntu)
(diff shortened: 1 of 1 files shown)


straight-shoota:
@bew please do not comment on the commit but on the PR diff.

syntax_and_semantics/cross-compilation.md
```diff
@@ -2,7 +2,20 @@
 
 Crystal supports a basic form of [cross compilation](http://en.wikipedia.org/wiki/Cross_compiler).
 
-In order to achieve this, the compiler executable provides two flags:
+## Preparation
+
+Cross-compilation will depend on `libcrystal` being present on the target system, so we need to build it by:
```
straight-shoota: Direct address ("we") should be avoided. I know, it's used all over the place but let's try to remove it.
Instead of `, so we need to build it by:` you could write `. It can ne built following these steps:`

syntax_and_semantics/cross-compilation.md
```diff
@@ -2,7 +2,20 @@
...
+* Downloading Crystal source code on the target machine.
```
straight-shoota: I'd suggest to use imperatives instead of gerunds. This better expresses intention and purpose.

syntax_and_semantics/cross-compilation.md
barisbalic: I mixed style because it was mixed in this document, will adjust as you suggested though!

bew:
@straight-shoota yeah sorry for the commit comment, it's not intuitive on the mobile version of github :worried: I'll try to be careful (no idea how/where to report a UX issue for github..)

barisbalic:
I've applied the feedback, and took the liberty of changing the form of address throughout that page so that it's uniform.  Please review the whole page again just in case I've made a mistake!

syntax_and_semantics/cross-compilation.md
```diff
@@ -2,33 +2,60 @@
 
 Crystal supports a basic form of [cross compilation](http://en.wikipedia.org/wiki/Cross_compiler).
 
-In order to achieve this, the compiler executable provides two flags:
+## Preparation
+
+Cross-compilation will depend on `libcrystal` being present on the target system.  It can be built by following these steps:
+* Download the Crystal source code on the target machine.
+* Install the LLVM/compiler toolchain. (e.g. `sudo apt-get install clang` on Ubuntu)
```
RX14: This isn't required to build libcrystal. It is however required for cross-compiling the compiler. Installing clang is a terrible way to install llvm.

I suggest this section is changed to make it clear installing LLVM/required libraries is only neccesary when cross-compiling the compiler. Libcrystal is neccesary regardless.

syntax_and_semantics/cross-compilation.md
```diff
@@ -2,33 +2,60 @@
...
+
+## Compiling
+
+The compiler executable provides two flags:
 
 * `--cross-compile`: When given enables cross compilation mode
 * `--target`: the [LLVM Target Triple](http://llvm.org/docs/LangRef.html#target-triple) to use and set the default [compile-time flags](compile_time_flags.html) from
 
-To get the `--target` flags you can execute `llvm-config --host-target` using an installed LLVM 3.5. For example on a Linux it could say "x86_64-unknown-linux-gnu".
+The `--target` flags can be determined by executing `llvm-config --host-target` on the target system. (e.g. 'x86_64-unknown-linux-gnu')
```
RX14: I suggest using `gcc -dumpmachine` instead, as `gcc` is far more common than `llvm-config`.

syntax_and_semantics/cross-compilation.md
```diff
@@ -2,33 +2,60 @@
...
+
+## Troubleshooting libraries
+
+The cross-compilation step will generate a command with numerous flags, many of these refer to the libraries that need to be included when compiling a program.
+If the libraries are not present on the target system, the compilation will fail and complain about the respective flags, which are not always descriptive.
+
+The following command may be helpful:
+
+```bash
+ld -lgc --verbose
```
RX14: This command prints a huge load of stuff that isn't very useful, perhaps a better method is required.

syntax_and_semantics/cross-compilation.md
barisbalic: Happy to change this to installing `llvm-3.x` directly.  Not sure I agree about the second part, given that this page is specifically in the context of cross-compiling already.  I think this speaks to the fact that the documentation on this page is a cross between general information about cross-compiling mixed with the specifics of cross-compiling the compiler.  Potentially worth separating these a little?

syntax_and_semantics/cross-compilation.md
barisbalic: Happy to switch this out, although the output seems less specific...

```bash
$ gcc -dumpmachine
arm-linux-gnueabihf
$ llvm-config --host-target
armv6-linux-gnueabihf
```

Admittedly I don't know how important this is.

syntax_and_semantics/cross-compilation.md
barisbalic: It does, but the volume doesn't matter that much right now, the useful output is right at the end, and I don't know of a better alternative, happy to change if you have any suggestions?

barisbalic:
@RX14 I've left some comments, let me know what you think and then I'll make any adjustments that remain.

syntax_and_semantics/cross-compilation.md
RX14: That was largely what I was getting at: instructions for cross-compiling hello world should be covered completely *then* we should cover cross-compiling the compiler.

syntax_and_semantics/cross-compilation.md
RX14: If it works as a target which you can cross-compile with, go for it.

syntax_and_semantics/cross-compilation.md
RX14: Well personally I think that having the paths doesn't really help you: you can either search for packages that provide `libfoo.so`, or you have to google `libfoo`. This command only really tells you the linker path, which is knowledge contained in the package already.

barisbalic: You say this but I would suggest you have so much integral knowledge by now it's hard to judge.  I had no idea what any of the flags were, seeing the paths was the only thing that allowed me to work out what libraries I needed to install.

syntax_and_semantics/cross-compilation.md
barisbalic: :+1:

syntax_and_semantics/cross-compilation.md
asterite: I wish someone could port libcrystal to c bindings. It's just one function (though a difficult one because of the bindings)

RX14: Luckily its not required for windows.

barisbalic: Definitely agree it should be separated, I was hoping to improve it incrementally rather than all in one go.

barisbalic:
@RX14 @straight-shoota, as much as I would like to, I don't see myself having any time to spend breaking this down into multiple pages any time soon.  This feels like an incremental improvement as it stands, but if you feel it's better to close this PR, I won't be offended.

straight-shoota:
@barisbalic Would you be able to apply the suggested changes?

barisbalic:
> @barisbalic Would you be able to apply the suggested changes?

I did for the most part, I left a comment that no-one replied to, aside from that it's just splitting the page into two sections, which I'm happy to do.

straight-shoota:
Yeah, splitting out the instructions for cross-compiling the compiler would be great. Also `llvm-config --host-target` can be replaced by `gcc -dumpmachine` (or maybe mention both?).

I'm not sure about what to do with the troubleshooting section, though. Maybe you could move that to a subsequent PR and we can get this merged soon?

barisbalic:
Okay cool, I'm in the process of moving home and starting a new work contract so I probably won't get around to this until the weekend, but very happy to do it.

kalinon:
@barisbalic do we still need this?

straight-shoota:
I merged this into a feature branch to combine with #424

PR #180, opened 2017-12-28
Add a section about binary size reduction

j8r:
~~This adds instructions for static linking.
I've also included a little change to the link explaining heap/stack memory.~~
- Improve the title to reflect the dynamically-linked nature of the compiled executable 
- Create a section about binary size reduction

### DIFF
--- a/using_the_compiler/README.md
+++ b/using_the_compiler/README.md
@@ -20,7 +20,7 @@ Alternatively you can use the `run` command:
 $ crystal run some_program.cr
 ```
 
-## Creating an executable
+## Creating a dynamically-linked executable
 
 To create an executable use the `build` command:
 
@@ -43,6 +43,18 @@ $ crystal build some_program.cr --release
 Make sure to always use `--release` for production-ready executables and when performing benchmarks.
 
 The reason for this is that performance without full optimizations is still pretty good and provides fast compile times, so you can use the `crystal` command almost as if it were an interpreter.
+
+To reduce the binary size, you can add the `--no-debug` flag and use the `strip` command. Debug symbols will be removed, use this option if only size is an issue and you won't need to debug the program.
+
+## Creating a standalone executable
+
+To build a standalone executable of your program:
+
+`
+$ crystal build some_program.cr --release --static
+```
+
+More informations about statically linking [can be found on the wiki](https://github.com/crystal-lang/crystal/wiki/Static-Linking)
 
 ## Creating a project or library
 


straight-shoota:
I'm not sure that experimental behaviour (even if labeled as such) should be in the language reference. Of course, Crystal itself is still somewhat experimental and many things might get changed before 1.0 is released. But there is still a difference to highly experimental stuff.

Maybe this could better be detailed in the Github Wiki for the time being.

The unrelated changes are good. The updated compiler output is already in #178 . The SO link should probably be in a separate PR.

j8r:
Ok I will rework this PR. I agree, we can keep the static linking section, explaining the `--static` flag, and add a link to a FAQ in the GitHub wiki to answer differents issues that can be faced.

using_the_compiler/README.md
```diff
@@ -44,6 +44,25 @@ Make sure to always use `--release` for production-ready executables and when pe
...
 
+To reduce the binary size, you can add the `--no-debug` flag and use the `strip` command. Debug symbols will be removed, use this options if only the size is an issue and you won't need to debug the program.
+
+## Creating a statically-linked executable
+
+This following libraries need to be required in the programm to force the linking against `libstdc++`:
+
+```
+require "llvm/lib_llvm"
+require "llvm/enums"
```
RX14: Very heavily -1 on documenting a bug.

j8r: I guess we will add this to the wiki, I remove it.

j8r:
I've removed the dirty hacks related to the static compilation @RX14 . I have kept them in another place temporaly on my side. Do we have to create a new Wiki page, or adding this informations to the FAQ?

using_the_compiler/README.md
```diff
@@ -44,6 +44,18 @@ Make sure to always use `--release` for production-ready executables and when pe
 
 The reason for this is that performance without full optimizations is still pretty good and provides fast compile times, so you can use the `crystal` command almost as if it were an interpreter.
 
+To reduce the binary size, you can add the `--no-debug` flag and use the `strip` command. Debug symbols will be removed, use this options if only the size is an issue and you won't need to debug the program.
```
RX14: *How* is the `strip` command used?

using_the_compiler/README.md
```diff
@@ -44,6 +44,18 @@ Make sure to always use `--release` for production-ready executables and when pe
...
+
+## Creating a standalone executable
```
RX14: Please call it "statically-linked".

using_the_compiler/README.md
```diff
@@ -44,6 +44,18 @@ Make sure to always use `--release` for production-ready executables and when pe
...
+
+To build a standalone executable of your program:
+
+`
+$ crystal build some_program.cr --release --static
+```
+
+More informations about statically linking [can be found on the wiki](https://github.com/crystal-lang/crystal/wiki/Static-Linking)
```
RX14: I think we shouldn't include this link.

j8r: But if we don't provide further informations about known issues the users will face, they will lose their time searching answers in existing issues, or even create new ones.

RX14: Honestly the `--static` flag won't do what 99% of users think it does. They'll think static compilation -> go, and our static compilation is nothing like go's because we still link to non-crystal C libraries, whereas go doesn't. So I strongly vote to just leave the flag out until we have an acceptable static linking story on any target that *isn't* musl.

j8r:
After discussion, the nature of this PR has quite changed and is now more about the binary size reduction (see the updated description)

asterite:
Does `strip` exist for windows?

`strip` seems like a general utility. I'm not sure it's a good idea to talk about this in a programming language. Searching "how to reduce binary size" in Google will probably point to strip or other utilities.

j8r:
No problem @asterite , I can just mention.

using_the_compiler/README.md
```diff
@@ -20,7 +20,7 @@ Alternatively you can use the `run` command:
 $ crystal run some_program.cr
 ```
 
-## Creating an executable
+## Creating a dynamically-linked executable
```
mverzilli: This change is unrelated to binary size reduction, please send it in a separate PR.

using_the_compiler/README.md
```diff
@@ -44,6 +44,12 @@ Make sure to always use `--release` for production-ready executables and when pe
 
 The reason for this is that performance without full optimizations is still pretty good and provides fast compile times, so you can use the `crystal` command almost as if it were an interpreter.
 
+## Reducing the executable's binary size
+
+This will remove debug symbols. Use this options only if the size is an issue and debugging isn't planned.
```
mverzilli: Can you send this paragraph below the one that starts with "To reduce...", it's often best to start with the core of the section ("to reduce binary size you can you the `--no-debug` flag") and then explain further ("you get a smaller binary file because you're effectively removing debug symbols").

mverzilli: "this options" => "these options"

using_the_compiler/README.md
```diff
@@ -44,6 +44,12 @@ Make sure to always use `--release` for production-ready executables and when pe
...
+
+To reduce the binary size, the `--no-debug` flag can be added. Furthermore, commands like `strip` can also be used.
```
mverzilli: The reference to strip is a bit too loose. Is it a flag for the compiler? Is it the Unix `strip` tool? Either remove the reference or add a bit more context assuming the reader has never been exposed to `strip`.

j8r:
I've made some changements according to your reviews @mverzilli

using_the_compiler/README.md
```diff
@@ -44,6 +44,16 @@ Make sure to always use `--release` for production-ready executables and when pe
...
 
+## Reducing the executable's binary size
+
+To reduce the binary size by removing debug symbols, the `--no-debug` flag can be added.
+
+```                                                                                                                                      
+$ crystal build some_program.cr --no-debug
+```
+
+Use this options only if the size is an issue and debugging isn't planned.
```
mverzilli: this options => this option

j8r: Thanks, the corrections are now applied

j8r:
Nothing more to do @mverzilli ?

mverzilli:
Thank you!

PR #185, opened 2017-12-30
Improve wording of Nil explaination

thefliik:
"not unlike" (a double negative) is harder for my brain to process than simply "similar to"


mverzilli:
Thank you!

PR #189, opened 2018-01-11
Various changes to compile_time_flags.md

GloverDonovan:
When I was learning about flags, I had to manually figure out how to define them. This pull request adds a new section on how to define flags and changes some wording.  It should be easier to follow the examples now.

### DIFF
--- a/syntax_and_semantics/compile_time_flags.md
+++ b/syntax_and_semantics/compile_time_flags.md
@@ -9,9 +9,15 @@ x86_64-unknown-linux-gnu
 # so the flags are: x86_64, unknown, linux, gnu
 ```
 
+To define a flag, simply use the `--define` or `-D` option, like so:
+
+```bash
+$ crystal some_program.cr -D my_flag
+```
+
 Additionally, if a program is compiled with `--release`, the `release` flag will be set.
 
-You can test these flags with the `flag?` macro method:
+You can check if a flag is defined with the `flag?` macro method:
 
 ```crystal
 {% if flag?(:x86_64) %}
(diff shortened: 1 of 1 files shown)


syntax_and_semantics/compile_time_flags.md
```diff
@@ -9,9 +9,15 @@ x86_64-unknown-linux-gnu
 # so the flags are: x86_64, unknown, linux, gnu
 ```
 
+To define a flag, simply use the `--define` or `-D` option, like so:
+
+```bash
+$ crystal some_program.cr -D my_flag
```
straight-shoota: Maybe add `crystal some_program.cr --debug my_flag` and  `crystal some_program.cr -Dmy_flag` as alternatives?

GloverDonovan: Can you clarify what I need to change? `-D` to `-Dmy_flag` or `--define my_flag`? I'd prefer using `--define my_flag` since it distinguishes the option from the value, but `-Dx86_64` could also be an option (feels  better than `my_flag` too).

Also when you said `--debug` did you mean `--define`?

straight-shoota: Oh, it was late... of course I meant `--define`.
You don't *need* to change anything. I would suggest to show the usage at least of using `-D` without whitespace. But I don't think `-Dx86_64` is a good idea because this flag is actually set automatically based on the target triple.

RX14: Personally I use `-Dflag`.

PR #196, opened 2018-01-28
Added help for missing libevent on MacOS + Homebrew installation

zerolith:


asterite:
The formula [already depends on libevent](https://github.com/Homebrew/homebrew-core/blob/master/Formula/crystal-lang.rb#L35)

mverzilli:
Hi @zerolith, as Ary mentioned the formula already states its dependency on libevent, I wonder what prompted you to add this. Maybe your brew environment wasn't clean or up to date?

zerolith:
Hi, sorry for the long wait. 
I restored an old backup and retried again. Seems that due to the upgrade from Sierra to High-Sierra my homebrew didn't upgrade very well and the libevent wasn't properly linked. Installed yes but not (re-) linked during the crystal setup.

PR #200, opened 2018-02-06
Fix String::Builder example

sudo-nice:
The former one gave me the error: `String::Builder.new' is not expected to be invoked with a block, but a block was given`


syntax_and_semantics/literals/string.md
```diff
@@ -76,7 +76,7 @@ Interpolation can be disabled by escaping the `#` character with a backslash or
 Interpolation is implemented using a [`String::Builder`](http://crystal-lang.org/api/String/Builder.html) and invoking `Object#to_s(IO)` on each expression enclosed by `#{...}`. The expression `"sum: #{a} + #{b} = #{a + b}"` is equivalent to:
 
 ```crystal
-String::Builder.new do |io|
+String::Builder.build do |io|
```
asterite: This should be just `String.build`

asterite: Oh, nevermind, it's for the docs of how it works... hmmm...

mverzilli:
Thank you!

PR #201, opened 2018-02-08
Add Unsigned Integers explanation

r00ster91:
Add an explanation when and why Unsigned Integers are good.

### DIFF
--- a/syntax_and_semantics/literals/integers.md
+++ b/syntax_and_semantics/literals/integers.md
@@ -64,3 +64,8 @@ Hexadecimal numbers start with `0x`:
 0xFE012D # == 16646445
 0xfe012d # == 16646445
 ```
+
+## When and why Unsigned Integers?
+
+Unsigned Integers are a good idea when your value will only be positive.
+Unsigned integers are more efficient and faster than normal integers.


mverzilli:
Hi @r00ster91! Thanks for contributing ^_^. I'm not sure about this PR for the reasons below:

1. The new section is getting too much weight in the page while its content is not central (e.g: why having a section on when to use "unsigned" and not when to pick 32bits vs 64bits integers?)
2. The range of supported integers is already in the table, so "Unsigned Integers are a good idea when your value will only be positive" adds no new information.
3. "Unsigned integers are more efficient and faster than normal integers.". I'm curious, do you have any references or benchmarks to back this claim? I wouldn't have expected there to be a difference.

straight-shoota:
To add to what @mverzilli  said: It might be worth adding a few notes about signed vs. unsigned integers, but what you're suggesting is not ideal I think.

Since when are unsigned integers faster than signed ones and why are signed declared as *normal integers*?

The advice to use an unsigned integer when the value is positive is at best very naive. Which type to use depends mostly on the context, for example comparisons with other integers of different types or operations that might potentially involve negative values.

I don't think the Crystal reference is the right place to discuss the details of integer data types.

mverzilli:
Yeah, in the end it will really depend on how and what you benchmark. Thanks for understanding!

PR #206, opened 2018-02-24
Fix wrong examples and texts in concurrency.md

r00ster91:
Some examples are wrong and some examples dont really do what the text actually says.

In the first example `socket.gets` isn't working. It's giving:
```
Error in weather.cr:3: undefined local variable or method 'socket'

  socket.gets
  ^~~~~~
```
(requiring "socket" doesn't fixes it too)
And also the text of it says "one reads from a socket". But shouldnt this be changed to "one reads from STDIN"?

Another example is this:
```crystal
spawn do
  loop do
    puts "Hello!"
  end
end

sleep 1.second
```
The text of this says that this will print "Hello!" for one second and then exit. But thats not true. It just prints "Hello!" until you press CTRL+C.

This PR is probably not perfect yet but im sure that theres something wrong with this Concurrency doc. It's probably a bit outdated.

### DIFF
--- a/guides/concurrency.md
+++ b/guides/concurrency.md
@@ -63,7 +63,7 @@ To spawn a fiber you use `spawn` with a block:
 ```crystal
 spawn do
   # ...
-  socket.gets
+  gets
   # ...
 end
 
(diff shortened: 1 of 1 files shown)


guides/concurrency.md
```diff
@@ -63,7 +63,7 @@ To spawn a fiber you use `spawn` with a block:
 ```crystal
 spawn do
   # ...
-  socket.gets
```
z64: The purpose of `socket.gets` is to show an abstract example of something you would *actually* do in another fiber; reading data from a socket (any socket). It isn't meant to be a compile-able example, and actually declaring a specific socket type and illustrating that is out of scope of the discussion. If desired, we could preface this example that explains that `socket` is abstract.

Moreover, reading input from `STDIN` is *not* typically something you would do in another fiber; for some kind of CLI application this is almost always going to be the "main" fiber of the application, so I don't think top level `gets` makes sense here.

PR #209, opened 2018-03-08
Remove reference of bare array creation

straight-shoota:
This feature has been removed in 0.24.0  (see crystal-lang/crystal#4824)

Fixes crystal-lang/crystal#5793


mverzilli:
Good catch, thanks!

Issue #213, opened 2018-03-09
Syntax and Semantics page barren


codyjb:
Opening page of this section is empty, suggesting adding a table of contents for its containing headers, as its the largest section anyways.

PR #214, opened 2018-03-09
Added top level syntax & semantics ToC

codyjb:
Fixes https://github.com/crystal-lang/crystal-book/issues/213

### DIFF
--- a/syntax_and_semantics/README.md
+++ b/syntax_and_semantics/README.md
@@ -1,3 +1,20 @@
 # Syntax and Semantics
 
 The program’s source code must be encoded in UTF-8.
+
+* [Comments](comments.html)
+* [Literals](literals.html)
+* [Assignment](assignment.html)
+* [Local Variabls](local_variables.html)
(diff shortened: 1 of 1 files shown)


RX14:
If we do this, shouldn't we do this for all sections?

codyjb:
Yeah, I think so, this was just the most egregious offender.

syntax_and_semantics/README.md
```diff
@@ -1,3 +1,20 @@
 # Syntax and Semantics
 
 The program’s source code must be encoded in UTF-8.
+
+* [Comments](comments.html)
+* [Literals](literals.html)
+* [Assignment](assignment.html)
+* [Local Variabls](local_variables.html)
```
mverzilli: Local Variables

mverzilli:
I think it's alright if we add the TOC pages little by little, or at least I'd rather have a few of them than none just for the sake of consistency.

PR #215, opened 2018-03-10
Add the description of StaticArray

sudo-nice:
Any remarks are wellcome.


sudo-nice:
This doesn't belong to Literals, but there's no good place for it neither.

straight-shoota:
Yeah, that's what I thought about this, too. `StaticArray` is mostly part of the standard library, maybe you can help improve the API docs? It's still very basic and your additions would certainly be welcome!

On the other hand, `StaticArray` is not entirely implemented in the stdlib but backed by some specific compiler features like having a number as generic argument, initialization of buffer (IIRC). This should also be covered in the language reference.

PR #216, opened 2018-03-16
Adding libxml2 and libxml2-dev dependencies

kazzkiq:
Ubuntu often throws errors when trying to running Crystal before installing `libxml2` and `libxml2-dev` dependencies. Its wise to document the need to install those dependencies them in the docs.

Related to: https://github.com/crystal-lang/crystal/issues/4342

### DIFF
--- a/installation/on_debian_and_ubuntu.md
+++ b/installation/on_debian_and_ubuntu.md
@@ -25,10 +25,10 @@ Once the repository is configured you're ready to install Crystal:
 sudo apt-get install crystal
 ```
 
-Sometimes [you will need](https://github.com/crystal-lang/crystal/issues/4342) to install the package `build-essential` in order to run/build Crystal programs. You can install it with the command:
+Sometimes [you will need](https://github.com/crystal-lang/crystal/issues/4342) to install the packages `build-essential`, `libxml2` and `libxml2-dev` in order to run/build Crystal programs. You can install them with the following command:
 
 ```
-sudo apt-get install build-essential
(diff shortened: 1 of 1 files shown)


straight-shoota:
I don't think this is the right way to do this. You will always require to install some libraries when using Crystal bindings for these libraries. If you're using XML, you will need libxml, for big numbers libgmp, for readline libreadline etc.

I think it would make sense to list all libraries used in the stdlib with installation instructions for each operating system.

It should probably also be mentioned in the API docs that in order to use `XML` you need to have libxml available (etc.).

RX14:
0.25.0 will prompt people without dependencies installed, see https://github.com/crystal-lang/crystal/pull/5486

straight-shoota:
IMO it should still be documented in the installation instructions and API docs.

RX14:
Give a person a list of dependencies, and they're fine until they start using shards with C bindings. Teach a person to install dependencies, and they're fine for a liftime.

straight-shoota:
Every shard usually documents what libraries it requires and how to install them. That should also be the case for the stdlib.

Having the error message point you to the libraries you're missing is a great improvement and has been duely missing so far. But if you know what libraries you need, you still need to know how to install them. And this can be different depending on your operating system/package manager. It would be great if  Crystal's documentation had installation instructions for the libraries used in the stdlib.

RX14:
@straight-shoota how to install each dependency is highly system specific, and the documentation for the dependencies of each module (yaml, xml, etc) should be inside the module's docs page (i.e. the API docs)

PR #220, opened 2018-03-21
Add example about numerical for iterating in macro

Heaven31415:
I wasn't able to find information about how to use "macro for" to iterate a specific amount of times, so I added a simple example.


mverzilli:
Thank you!

PR #232, opened 2018-04-05
Document the process of creating static builds

rzane:
I'm definitely not an expert in this sort of thing, so it is likely that I'm using the wrong terminology.

I didn't find a concise guide to creating static binaries anywhere, but I have seen a lot of people asking about how to create them.

The idea of creating pre-compiled binaries was the main motivation for me to use crystal in the first place. I assumed it was possible, but I was extremely frustrated when I couldn't find an explanation of how to create precompiled binaries for macs.

### DIFF
--- a/SUMMARY.md
+++ b/SUMMARY.md
@@ -108,6 +108,7 @@
        * [Uninitialized variable declaration](syntax_and_semantics/declare_var.md)
    * [Compile-time flags](syntax_and_semantics/compile_time_flags.md)
        * [Cross-compilation](syntax_and_semantics/cross-compilation.md)
+       * [Static builds](syntax_and_semantics/static_builds.md)
    * [C bindings](syntax_and_semantics/c_bindings/README.md)
        * [lib](syntax_and_semantics/c_bindings/lib.md)
        * [fun](syntax_and_semantics/c_bindings/fun.md)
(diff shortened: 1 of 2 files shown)


RX14:
Dynamically linked libraries are precompiled. I'm a bit reluctant to document static linking because it's really hard to do and hacky - and crystal doesn't *really* support it apart from on alpine linux.

j8r:
Maybe we can only document _the Docker Way_ to build statically-linked libraries. Is there an official container for?

Note that for now some libraries miss static archives required for static linking in Alpline, like `zlib-dev`. We have to compile them ourselves.

rzane:
Based on the numerous github issues that I read to get to this point, it definitely seems like a lot of people are interested in doing this sort of thing.

Perhaps this guide should explain the procedure, document the risks, and explain why it is not officially supported at this time.

So far, this is the only documentation that I've found on static linking:
* https://crystal-lang.org/docs/using_the_compiler/#creating-a-standalone-executable
* https://github.com/crystal-lang/crystal/wiki/Static-Linking

mverzilli:
How about we add this to an "experimental" section below "Guides"? That way we can centralize advice while clearly signaling that it is currently a "use at your own risk" feature.

straight-shoota:
I agree this should be added as a guide. It's already written that way.

But I don't think it should be labeled "experimental". Static linking works and is even an integral part of the compiler's own build process. I don't think there is anything experimental about it.
It just doesn't work on all systems. But that's not our fault.

kalinon:
If we dont need this as experimental, should be good to merge then right?

straight-shoota:
This was overtaken by #469
Nevertheless, thanks @rzane

PR #234, opened 2018-04-09
Fix c_bindings/fun can have upper case name

straight-shoota:
As of 0.24.0 function names in lib bindings can use upper case names (see crystal-lang/crystal#4934).

I also added two headings to structure the content of the page.

### DIFF
--- a/syntax_and_semantics/c_bindings/fun.md
+++ b/syntax_and_semantics/c_bindings/fun.md
@@ -59,7 +59,7 @@ lib LibSDL
 end
 ```
 
-The C name can be put in a string to be able to write a name that is not a valid identifier:
+The C name can be put in quotes to be able to write a name that is not a valid identifier:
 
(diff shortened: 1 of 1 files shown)


syntax_and_semantics/c_bindings/fun.md
```diff
@@ -47,15 +47,19 @@ X.variadic(1, 2, 3, 4)
...
+Function names in Crystal can be different from the C name. The following example shows how to bind the C function name `SDL_Init` as `LibSDL.init` in Crystal.
 
 ```crystal
 lib LibSDL
   fun init = SDL_Init(flags : UInt32) : Int32
 end
 ```
 
-You can also use a string as a name if the name is not a valid identifier or type name:
+The C name can be put in a string to be able to write a name that is not a valid identifier:
```
RX14: >put in quotes

PR #235, opened 2018-04-09
Add documentation for NoReturn

straight-shoota:


mverzilli:
Hi @straight-shoota! It's great to have `NoReturn` somehow documented but I'm not sure about the placement. Users of the language shouldn't need to know about `NoReturn`, although it's definitely useful for people who want to contribute to the compiler. Maybe we can find a better "home" for this? (cc: @bcardiff)

straight-shoota:
No, I think this should be in the language documentation. Users will be confronted with `NoReturn` when evaluating `typeof` of an expression wich does not return or when reading/writing C bindings. It should be documented in a way that language users can find it.

syntax_and_semantics/return_types.md
```diff
@@ -35,3 +35,23 @@ This is useful for two reasons:
...
+
+Some expressions won't return to the current scope and therefore have no return type. This is expressed as the special return type `NoReturn`.
+
+Typical examples for non-returning methods and keywords are `return`, `exit`, `raise`, `next`, and `break`.
+
+This is for example useful for deconstructing union types:
+
+```
+string = STDIN.gets
+typeof(string)                        # => String?
```
RX14: add a `typeof(raise "Empty input")` to make this clear

PR #236, opened 2018-04-15
Clarify protected usage

philxyz:
You can't do the following, but the existing comment implied that it could be done
```cr
class Person
  protected def self.sayit
    puts "Hi!"
  end
end

class Cat
  def say_hello
    Person.sayit
  end
end

Cat.new.say_hello
```

### DIFF
--- a/syntax_and_semantics/visibility.md
+++ b/syntax_and_semantics/visibility.md
@@ -118,7 +118,7 @@ end
 Namespace::Bar.new.bar
 ```
 
-A `protected` class method can be invoked from an instance method and the other way around:
+A `protected` class method can be invoked from an instance method (of the same class) and the other way around:
 
 ```crystal
 class Person


syntax_and_semantics/visibility.md
```diff
@@ -118,7 +118,7 @@ end
 Namespace::Bar.new.bar
 ```
 
-A `protected` class method can be invoked from an instance method and the other way around:
+A `protected` class method can be invoked from an instance method (of the same class) and the other way around:
```
straight-shoota: Adding parenthesis is unnecessary.

philxyz: Agreed. It seemed unnecessary to go back and change it however since there might also be a better way of wording the "and the other way around" part.

Something like: "A `protected` class or instance method can be invoked from any other one of the same class.

straight-shoota:
Fixed in  #314

PR #238, opened 2018-05-10
Add the other string interpolations in string.md

r00ster91:
And some other little improvements.

### DIFF
--- a/syntax_and_semantics/literals/string.md
+++ b/syntax_and_semantics/literals/string.md
@@ -43,8 +43,8 @@ A backslash followed by at most three digits denotes a code point written in oct
 A backslash followed by a `u` denotes a unicode codepoint. It can either be followed by exactly four hexadecimal characters representing the unicode bytes (`\u0000` to `\uFFFF`) or a number of one to six hexadecimal characters wrapped in curly braces (`\u{0}` to `\u{10FFFF}`.
 
 ```crystal
-"\u0041" # => "A"
-"\u{41}" # => "A"
+"\u0041"    # => "A"
+"\u{41}"    # => "A"
 "\u{1F52E}" # => "&#x1F52E;"
 ```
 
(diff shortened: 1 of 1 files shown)


r00ster91:
@RX14 Why are you confused? :/
I removed the ``` ` ``` at the String::Builder link so it's better to see that it's clickable. And in other docs it's also not surrounded by ``` ` ```

r00ster91:
But I can't really find any example of the Tuple string interpolation in the String API
```Crystal
"sum: %{one} + %{two} = %{three}" % {one: a, two: b, three: a + b}
```
Or that you need to use an Array when you want to embed for example multiple `%s` in an String
```Crystal
"sum: %s + %s = %s" % [a, b, a + b] 
```


So you mean I should add a link to https://crystal-lang.org/api/master/String.html#%25(other)-instance-method ?

RX14:
@r00ster91 yeah - and improving the `sprintf` docs to mention the tuple interpolation stuff would be good

r00ster91:
Maybe it's better to split these two interpolations from the compiler-interpolation so people better understand that the sprintf-interpolations are not expanded at compile time.

syntax_and_semantics/literals/string.md
```diff
@@ -61,7 +61,14 @@ A string literal with interpolation allows to embed expressions into the string
 ```crystal
 a = 1
 b = 2
-"sum: #{a} + #{b} = #{a + b}"  # => "sum: 1 + 2 = 3"
+"sum: #{a} + #{b} = #{a + b}" # => "sum: 1 + 2 = 3"
+```
+
+There are also two other types of interpolation using `Kernel#sprintf`. They are not expanded at runtime like the other one. See [String#%](https://crystal-lang.org/api/master/String.html#%25(other)-instance-method).
```
RX14: I think you mean the opposite: these *are* expanded at runtime, the other one isn't.

straight-shoota: But technically, both `Kernel.sprintf` and `String#%` are just API methods provided by the stdlib and not really relevant in this section about string literals.
It might be helpful to add a short note about this, but I wouldn't expand it to much.

r00ster91: You are right. I think I just take out the sprintf-interpolations and just add the examples to https://crystal-lang.org/api/master/String.html#%25(other)-instance-method and we just merge the other little improvements.

r00ster91:
Ok I made the PR in https://github.com/crystal-lang/crystal/pull/6095 now.

r00ster91:
[#6095](https://github.com/crystal-lang/crystal/pull/6095) is merged so I think this can be merged too. Or is there still something that should be changed?

PR #243, opened 2018-06-15
Update HTTP Server example

pudarida:
Current HTTP Server example code breaks on `HTTP::Server.new(8080)` as it is no longer a valid overload.

New example code is based on [crystal-lang/crystal/blob/master/samples/http_server.cr](https://github.com/crystal-lang/crystal/blob/master/samples/http_server.cr)

### DIFF
--- a/overview/http_server.md
+++ b/overview/http_server.md
@@ -3,15 +3,16 @@
 A slightly more interesting example is an HTTP Server:
 
 ```crystal
+
 require "http/server"
 
-server = HTTP::Server.new(8080) do |context|
-  context.response.content_type = "text/plain"
-  context.response.print "Hello world! The time is #{Time.now}"
+server = HTTP::Server.new do |context|
+  context.response.headers["Content-Type"] = "text/plain"
+  context.response.print("Hello world! The time is #{Time.now}")
 end
 
-puts "Listening on http://127.0.0.1:8080"
-server.listen
+puts "Listening on http://0.0.0.0:8080"
+server.listen "0.0.0.0", 8080
 ```
 
 The above code will make sense once you read the whole language reference, but we can already learn some things.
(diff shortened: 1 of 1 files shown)


overview/http_server.md
```diff
@@ -3,15 +3,16 @@
...
-  context.response.content_type = "text/plain"
-  context.response.print "Hello world! The time is #{Time.now}"
+server = HTTP::Server.new do |context|
+  context.response.headers["Content-Type"] = "text/plain"
+  context.response.print("Hello world! The time is #{Time.now}")
 end
 
-puts "Listening on http://127.0.0.1:8080"
-server.listen
+puts "Listening on http://0.0.0.0:8080"
```
straight-shoota: If it is supposed to print the listening address, I'd rather do this:
```cr
address = server.bind_tcp "0.0.0.0", 8080
puts "Listening on #{address}"
server.listen
```
This is more DRY and in some cases the actual address might be different from the arguments delivered to `listen`.

pudarida: That's preferrable.

overview/http_server.md
```diff
@@ -3,14 +3,16 @@
 A slightly more interesting example is an HTTP Server:
 
 ```crystal
+
```
RX14: please remove this line

overview/http_server.md
```diff
@@ -3,14 +3,16 @@
...
 require "http/server"
 
-server = HTTP::Server.new(8080) do |context|
-  context.response.content_type = "text/plain"
-  context.response.print "Hello world! The time is #{Time.now}"
+server = HTTP::Server.new do |context|
+  context.response.headers["Content-Type"] = "text/plain"
```
RX14: why this change?

overview/http_server.md
```diff
@@ -3,14 +3,16 @@
...
+  context.response.print("Hello world! The time is #{Time.now}")
```
RX14: ditto

RX14: This was more idiomatic as-is. That example is pretty old, and I think it predates `HTTP::Response#content_type`. The example was fine as-is anyway, no need to inflate the diff.

straight-shoota:
This can be closed. #256 has already been merged.

PR #245, opened 2018-06-15
Create OpenSUSE Instructions

sirredbeard:
I created some installation instructions to install Crystal with zypper on OpenSUSE.

This is easy because zypper supports official project rpm repos for CentOS and RHEL.

### DIFF
--- a/installation/on_opensuse.html
+++ b/installation/on_opensuse.html
@@ -1,0 +1,32 @@
+# On OpenSUSE
+
+In OpenSUSE, you can use the official Crystal repository.
+
+## Setup repository
+
+First you have to add the signing key:
+
+```
+rpm --import https://dist.crystal-lang.org/rpm/RPM-GPG-KEY
+```
+
+Next you have to configure the repository in Zypper. You can do this with the following command:
+
+```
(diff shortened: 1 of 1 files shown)


straight-shoota:
@sirredbeard Could you change the file name to `.md` and append the commit from #246?

installation/on_opensuse.html
```diff
@@ -0,0 +1,32 @@
...
+First you have to add the signing key:
+
+```
+rpm --import https://dist.crystal-lang.org/rpm/RPM-GPG-KEY
+```
+
+Next you have to configure the repository in Zypper. You can do this with the following command:
+
+```
+sudo zypper ar -e -f -t rpm-md https://dist.crystal-lang.org/rpm/ Crystal```
```
straight-shoota: Those backticks at the end are probably an error.

installation/on_opensuse.html
```diff
@@ -0,0 +1,32 @@
+# On OpenSUSE
+
+In OpenSUSE, you can use the official Crystal repository.
```
straight-shoota: I'd prefer a more descriptive writing style instead of directly addressing the reader in every sentence.
Example:
```
On OpenSUSE, Crystal can be installed from the official rpm package using Zypper.
```

PR #249, opened 2018-06-19
Troubleshooting on WSL

marcelobarreto:
This is how I solved a problem trying to install Crystal on WSL.


straight-shoota:
I'm not sure if this is a common issue that would warrant putting it in the reference... Maybe this is better for StackOverflow?

mverzilli:
Hi @marcelobarreto, thank you so much for collaborating! I agree with @straight-shoota, StackOverflow is the way to go here. It's not widely advertised, but you can answer your own questions in SO. So I'd create a question on SO, tag it `crystal-lang` (https://stackoverflow.com/questions/tagged/crystal-lang) and answer it right away.

That way it has much higher chances of appearing in Google searches from people who stumbled upon your same issue.

PR #260, opened 2018-07-16
Note that statically-linked executables can only be built on Alpine Linux

RX14:


straight-shoota:
Great to see you also improved the references wiki page (at least a little bit). It would great to explain some of the background there. I'd do it myself but don't know enough about it...

mverzilli (approved):

PR #262, opened 2018-07-16
Remove redundancy in WSL instructions and improve for Debian/Ubuntu

straight-shoota:
Installing Crystal on Windows Subsystem for Linux typically doesn't differ from installing on the respective native Linux platform. The only notable exception to that is only recently with Ubuntu 18.04, but not a big deal (see #259).
And WSL easily supports different Linux distributions, not just Ubuntu.

That's why I figure it doesn't make much sense to have detailed instructions for WSL but instead point to the respective Linux distribution's installation instructions.

Renamed `Bash on Ubuntu on Windows` to `Windows Subsystem for Linux` which is the proper and more general name. I left the path at `installation/on_bash_on_ubuntu_on_windows.md` to not break links. It could be moved and redirect, but I don't think it hurts much.

The assessment of *highly experimental*, is certainly not adequate anymore. Crystal on WSL works as good and stable as on any native platform.

I also extended the installation instructions for Debian/Ubuntu, adding a few things from the WSL page and recommended packages. It also incorporates #259.

Closes #259

### DIFF
--- a/installation/on_debian_and_ubuntu.md
+++ b/installation/on_debian_and_ubuntu.md
@@ -13,13 +13,10 @@ curl -sSL https://dist.crystal-lang.org/apt/setup.sh | sudo bash
 That will add the signing key and the repository configuration. If you prefer to do it manually, execute the following commands:
 
 ```bash
-sudo sh -c 'apt-key adv --keyserver keys.gnupg.net --recv-keys 09617FD37CC06B54 && \
-            echo "deb https://dist.crystal-lang.org/apt crystal main" > /etc/apt/sources.list.d/crystal.list && \
-            apt-get update'
+curl -sL "https://keybase.io/crystal/pgp_keys.asc" | sudo apt-key add
+echo "deb https://dist.crystal-lang.org/apt crystal main" | sudo tee /etc/apt/sources.list.d/crystal.list
+sudo apt-get update
 ```
-
-**NOTE:** There is a bug in Ubuntu 18.04 running on Windows Subystem for Linux, making `apt-key adv` fail.
-As a workaround, the first command can be replaced by `curl -sL "https://keybase.io/crystal/pgp_keys.asc" | sudo apt-key add`.
 
 ## Install
 Once the repository is configured you're ready to install Crystal:
@@ -28,13 +25,7 @@ Once the repository is configured you're ready to install Crystal:
 sudo apt install crystal
 ```
 
-Sometimes the package `build-essential` needs to be installed in order to run/build Crystal programs (see [issue #4342](https://github.com/crystal-lang/crystal/issues/4342)):
-
-```bash
-sudo apt install build-essential
-```
-
-The following packages are not required, but recommended for using the respective stlib features:
+The following packages are not required, but recommended for using the respective features in the standard library:
 
 ```bash
 sudo apt install libssl-dev      # for using OpenSSL
--- a/installation/on_debian_and_ubuntu.md
+++ b/installation/on_debian_and_ubuntu.md
@@ -6,37 +6,51 @@ In Debian derived distributions, you can use the official Crystal repository.
 
 First you have to add the repository to your APT configuration. For easy setup just run in your command line:
 
-```
-curl https://dist.crystal-lang.org/apt/setup.sh | sudo bash
+```bash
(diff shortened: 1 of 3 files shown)


installation/on_debian_and_ubuntu.md
```diff
@@ -6,37 +6,51 @@ In Debian derived distributions, you can use the official Crystal repository.
...
 
-That will add the signing key and the repository configuration. If you prefer to do it manually, execute the following commands as *root*:
+That will add the signing key and the repository configuration. If you prefer to do it manually, execute the following commands:
 
-```
-apt-key adv --keyserver keys.gnupg.net --recv-keys 09617FD37CC06B54
-echo "deb https://dist.crystal-lang.org/apt crystal main" > /etc/apt/sources.list.d/crystal.list
-apt-get update
+```bash
+sudo sh -c 'apt-key adv --keyserver keys.gnupg.net --recv-keys 09617FD37CC06B54 && \
```
RX14: please make these seperate sudo invocations. use the old trick `echo "foo" | sudo tee /etc/apt/sources.list.d/crystal.list` to replace `>`

installation/on_debian_and_ubuntu.md
```diff
@@ -6,37 +6,51 @@ In Debian derived distributions, you can use the official Crystal repository.
...
+            echo "deb https://dist.crystal-lang.org/apt crystal main" > /etc/apt/sources.list.d/crystal.list && \
+            apt-get update'
 ```
 
+**NOTE:** There is a bug in Ubuntu 18.04 running on Windows Subystem for Linux, making `apt-key adv` fail.
+As a workaround, the first command can be replaced by `curl -sL "https://keybase.io/crystal/pgp_keys.asc" | sudo apt-key add`.
```
RX14: Please replace this line in the instructions above and dont mention `adv` at all.

installation/on_debian_and_ubuntu.md
```diff
@@ -6,37 +6,51 @@ In Debian derived distributions, you can use the official Crystal repository.
...
+
 ## Install
 Once the repository is configured you're ready to install Crystal:
 
-```
-sudo apt-get install crystal
+```bash
+sudo apt install crystal
```
RX14: Please revert to support older releases.

installation/on_debian_and_ubuntu.md
```diff
@@ -6,37 +6,51 @@ In Debian derived distributions, you can use the official Crystal repository.
...
 ```
 
-Sometimes [you will need](https://github.com/crystal-lang/crystal/issues/4342) to install the package `build-essential` in order to run/build Crystal programs. You can install it with the command:
+Sometimes the package `build-essential` needs to be installed in order to run/build Crystal programs (see [issue #4342](https://github.com/crystal-lang/crystal/issues/4342)):
```
RX14: Should be fixed, the package depends on `gcc` now.

installation/on_debian_and_ubuntu.md
```diff
@@ -6,37 +6,51 @@ In Debian derived distributions, you can use the official Crystal repository.
...
 ## Upgrade
 
 When a new Crystal version is released you can upgrade your system using:
 
-```
-sudo apt-get update
-sudo apt-get install crystal
+```bash
+sudo apt update
+sudo apt install crystal
```
RX14: ditto

installation/on_debian_and_ubuntu.md
straight-shoota: `apt` is available in both Debian 8 and Ubuntu 14.04 LTS, the oldest still supported versions of both distributions. I don't think we should need support for anything else.

installation/on_debian_and_ubuntu.md
straight-shoota: Then `https://dist.crystal-lang.org/apt/setup.sh` should be changed accordingly.

installation/on_debian_and_ubuntu.md
straight-shoota: Then `https://dist.crystal-lang.org/apt/setup.sh` should be changed accordingly.

RX14: yes it should

PR #263, opened 2018-07-16
The key is missing from keys.gnupg.net

aognio:
The key is available from keybase.io instead.


straight-shoota:
Why is it not listed on keys.gnupg.net?

aognio:
No idea! I had to google in order to find this issue and the last comment mentioned the availability on keybase.io:

https://github.com/crystal-lang/crystal/issues/4417

straight-shoota:
This has already been fixed in #262.

PR #264, opened 2018-07-17
Improve `Using the Compiler`

straight-shoota:
Huge overhaul adding detailed descriptions of every compiler command.

PR #269, opened 2018-07-24
Remove out-of-date mention of "deps"

earthrid:
Removed mention of crystal's "deps" command since this has been replaced by "shards".


straight-shoota:
This is also included in #264 btw.

PR #273, opened 2018-08-07
Added instruction  how to generate ebook files 

kolyaio:
Hello, 

I added instructions how to generate ebook files by using the build in abilities of gitbook. 
Personally I prefer to read the docs as a book so I published in the gitter chat room to add those instructions.  I would appreciate any fixes or suggestions to what I did. English isn't my native language 😒


mverzilli:
Hi @kolyaio, thanks for this! I feel this is a bit ad-hoc to be included in the book. One nice way to ensure people find this kind of detailed information in the future without breaking the book's flow is to open a StackOverflow question and answer it yourself. That makes it more "googleable", and you're likely to get contributions that complete the topic (which in this case would be "How do I export a gitbook to X?").

PR #275, opened 2018-08-09
Update on_debian_and_ubuntu.md

terapackets:
`sudo apt-key add` is changed to `sudo apt-key add -` so it won't generate an error on my Ubuntu 14.04


mverzilli:
Thank you!

PR #276, opened 2018-08-09
Update inheritance.md - covariance and contravariance

KCErb:
Not sure about linking to an issue in the gitbook ... but otherwise, how's this for a crack at https://github.com/crystal-lang/crystal/issues/6507#issuecomment-411714904?

### DIFF
--- a/syntax_and_semantics/inheritance.md
+++ b/syntax_and_semantics/inheritance.md
@@ -102,3 +102,57 @@ end
 ```
 
 Without arguments or parentheses, `super` receives the same arguments as the method's arguments. Otherwise, it receives the arguments you pass to it.
+
+## Covariance and Contravariance
+
+One place this can get a little tricky is with arrays. We have to be careful when declaring an array of objects where inheritance is used. For example, consider the following
+
+```crystal
+class Foo
+end
+
+class Bar < Foo
+end
+
+foo_arr = [Bar.new] of Foo # => [#<Bar:0x10215bfe0>] : Array(Foo)
+bar_arr = [Bar.new] # => [#<Bar:0x10215bfd0>] : Array(Bar)
+bar_arr2 = [Foo.new] of Bar # compiler error
+```
+
+A Foo array can hold both Foo's and Bar's, but an array of Bar can only hold Bar and it's subclasses.
+
+One place this might trip you up is when automatic casting comes into play. For example, the following won't work:
+
+```crystal
+class Foo
+end
+
+class Bar < Foo
+end
+
+class Test
+  @arr : Array(Foo)
+  def initialize
+    @arr = [Bar.new]
+  end
+end
+```
+
+because in the initialize the default type for @arr is `Array(Bar)` but the required type is `Array(Foo)`. You can solve this by specifying the type explicitly:
+
+```crystal
+class Foo
+end
+
+class Bar < Foo
+end
+
+class Test
+  @arr : Array(Foo)
+  def initialize
+    @arr = [Bar.new] of Foo
+  end
+end
+```
+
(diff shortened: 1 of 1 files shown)


syntax_and_semantics/inheritance.md
```diff
@@ -102,3 +102,57 @@ end
 ```
 
 Without arguments or parentheses, `super` receives the same arguments as the method's arguments. Otherwise, it receives the arguments you pass to it.
+
+## Covariance and Contravariance
+
+One place this can get a little tricky is with arrays. We have to be careful when declaring an array of objects where inheritance is used. For example, consider the following
```
RX14: please be specific about what "this" is.

KCErb:
Thanks @RX14 for the review. I updated `this` to  `inheritance`.

mverzilli:
Thank you @KCErb!

syntax_and_semantics/inheritance.md
```diff
@@ -102,3 +102,57 @@ end
...
+
+class Test
+  @arr : Array(Foo)
+  def initialize
+    @arr = [Bar.new]
+  end
+end
+```
+
+because in the initialize the default type for @arr is `Array(Bar)` but the required type is `Array(Foo)`. You can solve this by specifying the type explicitly:
```
bcardiff: > default type for @arr

This is not concept of the type system. I would suggest explaining that the type of the `[Bar.new]` expression is `Array(Bar)` and that is not assignable to an `Array(Foo)` instance var.

Then it follows from the previous example why the expression `[Bar.new] of Foo` works.

syntax_and_semantics/inheritance.md
```diff
@@ -102,3 +102,57 @@ end
...
+
+class Test
+  @arr : Array(Foo)
+  def initialize
+    @arr = [Bar.new] of Foo
+  end
+end
+```
+
+The way Crystal handles the bigger topic of [covariance and contravariance](https://en.wikipedia.org/wiki/Covariance_and_contravariance_(computer_science)) in general, has more tricks and pitfalls to it, so you may be interested in [this issue / discussion](https://github.com/crystal-lang/crystal/issues/3803) for more reading.
```
bcardiff: I would not link issues from the repo. It does not help to read the guide as a book.

Also,
> has more tricks and pitfalls to it
adds no explanation or description current limitations.

I know the Covariance and Contravariance is not fully supported since the user can't express fully it's expectation in generic classes, nor the compiler infers what happens. Given that, maybe examples should cover: tuples (or named tuples), arrays, custom generic classes.

Another aspect mentioned is that method matching does not uses the notion of substitutability to determine if a definition should be used or not. In this PR only assignments were used, hence substitutability principle is used.

KCErb: I'd be happy to remove this link, seems a little strange to me. :)

As for some of the examples you mentioned, is this the best place or is it potentially growing the article too much? If we really want to address covariance and contravariance as a topic in the gitbook, I might not be the best person, it's a little over my head at the moment (though I could catch up with a bit of reading it seems).

I can think of a few roads:

1. Small: just remove the link and the stuff about "pitfalls" to keep it to one simple example of an array and link to the wikipedia article.
2. Medium: add in an example using a Tuple here or put one on Generics over in the [generics inheritance](https://crystal-lang.org/docs/syntax_and_semantics/generics.html#generic-types-inheritance) section and link to it.
3. Large: consider doing a separate page on the topic of covariance and contravariance to fully document the expected behavior in a wide range of cases, link to that in places that mention inheritance such as this article or the one I just mentioned on generics.

It's only that third option that I wouldn't feel totally qualified to write up properly. Which direction do you (folks) lean?

bcardiff:
Great! Thanks. Opening a new PR is fine. This PR is an improvement by itself :-)

KCErb:
OK cool, will do next chance

PR #278, opened 2018-08-27
Fix "crystal-lang" homebrew package name

tompark:
The `brew install crystal` command will install something called "autocode".


tompark:
Sorry about that, my bad. I didn't update brew before doing the install.

PR #284, opened 2018-09-27
Update inheritance.md based on comments in #276

KCErb:
OK, after a bit longer than I expected, here's some of the changes you recommended @bcardiff. I decided to keep the discussion short and to the point. If folks feel this is an important topic, maybe we can open an issue outlining other cases to document as @bcardiff mentioned in #276:

> maybe examples should cover: tuples (or named tuples), arrays, custom generic classes


KCErb:
Should I close this?

bcardiff:
Thanks! (and sorry 🐌!)

PR #287, opened 2018-10-07
Fixed a variable name in the if-var concurrent clojure example

icanhazbroccoli:
The example demonstrates a concurrent variable update case, which causes a non-deterministic runtime behavior. This commit changes the variable name defined in the upper scope and alternated in this clojure.


icanhazbroccoli:
Thanks for a quick merge guys! ❤️

PR #289, opened 2018-10-12
Installing crystal on macOS can take some time

zigomir:
It took me 2h 35m 56s on a 2016 MacBook Pro, Core i7 😲

I might be doing something wrong, but `brew install crystal` took that much.

I'm not 100% serious on the change I'm proposing, but it would be nice to give some heads up, if that much time is really expected.

### DIFF
--- a/installation/on_mac_osx_using_homebrew.md
+++ b/installation/on_mac_osx_using_homebrew.md
@@ -7,6 +7,8 @@ brew update
 brew install crystal
 ```
 
+(this might take quite some time ☕️)
+
 ## Troubleshooting on OSX 10.11 (El Capitan)
 
 If you get an error like:


bew:
I installed Crystal on a brand new macbook pro some weeks ago, and it took 5min or so (definitely not 2h...).

(i won't be able to give you more information, as I don't have it anymore, macos is not for me :laugh:)

proyb6:
Probably ping what the latency?

It only took me a few minutes to install on macOS Mojave and could be tournament network issue or ISP

zigomir:
Interesting. No, I'm pretty sure it wasn't network issue, it was running `make` and I had CPU at 99-100% for whole two hours and half 😲

RX14:
I can imagine compiling LLVM taking 2 hours. You shouldn't have to have compiled LLVM *or* crystal. Homebrew provides binaries which should have been downloaded.

zigomir:
I see. I don't know what happened or went wrong. It's now installed and it works. Because this isn't expected behavior I think it's safe to close this one. Thanks!

PR #294, opened 2018-11-10
Add more FreeBSD info (quarterly pkgs, port building without llvm rebuilding)

valpackett:


straight-shoota:
Awesome! This is very useful for people not familiar with Alpine Linux.

PR #298, opened 2018-11-14
fix typo in crystal tool format

marynowac:


j8r:
"`crystal fool format`" – that was comic 😄

PR #300, opened 2018-12-02
Link to crystalshards.org instead of .xyz, which is down

oprypin:


j8r:
Why closing? Note the site isn't down, only unmaintained.
I suggest to mention both.

straight-shoota:
cf hcrystal-lang/crystal-website#73

PR #165, opened 2017-12-19
Overhaul Regex literal docs

straight-shoota:
Extends the documentation of regex literals. It contains some duplication from the recently overhauled string literal docs, but I think it is better to have the most important things here directly and more in-depth details can be handled in the string literal docs.

The part about unicode escapes is currently not accurate, see crystal-lang/crystal#5412 but this is how it *should* be.

PR #301, opened 2018-12-02
Fix regex literal docs incorrectly listing \uNNNN syntax

meew0:
For some reason, this syntax is listed in the docs despite leading to an error like
```
Syntax error in regex-test.cr:12: invalid regex: PCRE does not support \L, \l, \N{name}, \U, or \u at 1

/\u1234/ # hexadecimal unicode character
^

```

The `\x{NNNN}` syntax seems to work fine however, so I've replaced the `\u` syntax with that.


straight-shoota:
Ah yes, that's how it should be, but crystal-lang/crystal#5412 is still open.
#165 was merged before fixing that.

PR #303, opened 2018-12-06
Add a link to API docs on frontpage

straight-shoota:


straight-shoota:
It's not semantically correct, but distinguishes this note from the main body. I can remove it, if not appreciated. Markdown has no syntax for admonitions.

RX14:
It actually makes it less visible by greying out the text

PR #308, opened 2018-12-10
Add html-proofer to .travis.yml

straight-shoota:
[html-proofer](https://github.com/gjtorikian/html-proofer) automatically validates image, link and script references and helps detecting broken links.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,11 +2,15 @@ sudo: false
 cache:
   directories:
     - $HOME/.npm
+    - $TRAVIS_BUILD_DIR/tmp/.htmlproofer
 before_install: |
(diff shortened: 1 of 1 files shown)


straight-shoota:
See #307 for issues found with this tool.

Please note that the configuration used for CI doesn't validate external link fragments. That seems to be error-prone because many links to method definitions on https://crystal-lang.org/api are flagged as false positive.

It can be enabled by adding `--check-external-hash`, accompanied by `--url-ignore "/crystal-lang.org\/api/"` to suppress these errors. However, this would also suppress errors where the entire link is broken, not just the fragment.

RX14:
please rebase

straight-shoota:
Done, all green

PR #310, opened 2018-12-18
Document the chained assignment

r00ster91:
This documents the chained assignment and explains that dynamic assignments like the chained assignment is not supported for constants.


straight-shoota:
I'd put the part about chained assignment into the page about variables. If at all. It's hardly worth a dedicated topic. 

And it not only works for ivars and class variables, but also for local variables. And even assignment methods.

j8r:
In the `Tuple` page a section could be added to explain multiple assignements for methods returning a `Tuple`:

```crystal
def first
  {1, "one"}
end
num, str = first
p num # 1
p str #=> "one"
```
That's were multiple assignments become interesting.

There are also the multiple assignments with `Array#split`.

bew:
@j8r it's not the same kind of assignment that is being documented here.. And IIRC yours is already documented in the book

j8r:
@bew where? I didn't see them in `Tuple` nor in `Array`

bew:
Here: https://crystal-lang.org/docs/syntax_and_semantics/multiple_assignment.html (3rd code block

j8r:
Remains the `Tuple` multiple assignments :wink: . Sometimes it's really useful, for example in `initialize` - it's worth adding to the doc.

r00ster91:
This is actually already documented on https://crystal-lang.org/api/0.27.0/Tuple.html (second code example).

@straight-shoota
> I'd put the part about chained assignment into the page about variables. If at all. It's hardly worth a dedicated topic.

I think it would fit better in the [Assignment](https://crystal-lang.org/docs/syntax_and_semantics/assignment.html) section since it's a kind of assignment similar to the [Multiple assignment](https://crystal-lang.org/docs/syntax_and_semantics/multiple_assignment.html). Without it, the section would be a bit lonely with only one member. Also, the section is complete now with the Chained assignment.

> And it not only works for ivars and class variables, but also for local variables. And even assignment methods.

I used local variables in the example so it should be obvious that it's supported but I will probably make that a bit more explict and add assignment methods.

syntax_and_semantics/chained_assignment.md
```diff
@@ -0,0 +1,34 @@
...
+
+```crystal
+person1.name = person2.name = "John"
+
+# The above is the same as:
+person1.name = "John"
+person2.name = "John"
+```
+
+And it is also available to [indexers](operators.md#indexing) (`[]=`):
```
straight-shoota: This sentence and example is unnecessary. It's already covered by the previous one ("end with `=`").

syntax_and_semantics/constants.md
```diff
@@ -64,3 +64,11 @@ end
 # File path is: /crystal_code/pseudo_constants.cr
 # Directory file is in: /crystal_code
 ```
+
+# Dynamic assignment
+
+Dynamically assigning values to constants using the [chained assignment](chained_assignment.md) or the [multiple assignment](multiple_assignment.md) is not supported for constants and results in a syntax error.
+
+```crystal
+ONE, TWO, THREE = 1, 2, 3 # => Syntax error: Multiple assignment is not allowed for constants
```
straight-shoota:
```suggestion
ONE, TWO, THREE = 1, 2, 3 # Syntax error: Multiple assignment is not allowed for constants
```

syntax_and_semantics/constants.md
```diff
@@ -64,3 +64,11 @@ end
 # File path is: /crystal_code/pseudo_constants.cr
 # Directory file is in: /crystal_code
 ```
+
+# Dynamic assignment
+
+Dynamically assigning values to constants using the [chained assignment](chained_assignment.md) or the [multiple assignment](multiple_assignment.md) is not supported for constants and results in a syntax error.
```
r00ster91:
```suggestion
Dynamically assigning values to constants using the [chained assignment](chained_assignment.md) or the [multiple assignment](multiple_assignment.md) is not supported and results in a syntax error.
```

syntax_and_semantics/assignment.md
```diff
@@ -62,3 +62,88 @@ objects[1] += 2        # same as: objects[1] = objects[1] + 2
 objects[1] ||= 2       # same as: objects[1]? || (objects[1] = 2)
 objects[1] &&= 2       # same as: objects[1]? && (objects[1] = 2)
 ```
+
+# Chained assignment
+
+You can assign multiple variables to one value at the same time using the chained assignment:
```
straight-shoota:
```suggestion
You can assign the same value to multiple variables using chained assignment:
```

SUMMARY.md
```diff
@@ -34,6 +34,7 @@
        * [Proc](syntax_and_semantics/literals/proc.md)
    * [Assignment](syntax_and_semantics/assignment.md)
        * [Multiple assignment](syntax_and_semantics/multiple_assignment.md)
+       * [Chained assignment](syntax_and_semantics/chained_assignment.md)
```
straight-shoota: This line and the one above need to go when they're not separate pages.

syntax_and_semantics/assignment.md
```diff
@@ -62,3 +62,88 @@ objects[1] += 2        # same as: objects[1] = objects[1] + 2
...
+
+```crystal
+var1 = var2 = var3 = 123
```
straight-shoota:
```suggestion
a = b = c = 123
```
Simple var names are easier to read than indexed ones.

syntax_and_semantics/assignment.md
```diff
@@ -62,3 +62,88 @@ objects[1] += 2        # same as: objects[1] = objects[1] + 2
...
+
+# The above is the same as this:
+var1 = 123
+var2 = 123
+var3 = 123
+```
+
+The chained assignment is not only available to [local variables](local_variables.md) but also to [instance variables](methods_and_instance_variables.md) and [class variables](class_variables.md).
```
straight-shoota: Maybe incorporate the following sentence? It's the same idea.
```suggestion
The chained assignment is not only available to [local variables](local_variables.md) but also to [instance variables](methods_and_instance_variables.md), [class variables](class_variables.md) and setter methods (methods ending with the equals sign.`=`).
```

syntax_and_semantics/assignment.md
```diff
@@ -62,3 +62,88 @@ objects[1] += 2        # same as: objects[1] = objects[1] + 2
...
 ```
+
+# Chained assignment
+
+You can assign multiple variables to one value at the same time using the chained assignment:
+
+```crystal
+var1 = var2 = var3 = 123
+
+# The above is the same as this:
```
straight-shoota: It's actually not the same and it's only equivalent in some cases.
The value is duplicated and not evaluated multiple times. For example with `a = b = c = rand` all variables have the same value.
I would only add the resulting values to the example:
```cr
a # => 123
b # => 123
c # => 123
```

And maybe there could be a note that this works because the assignment operator returns the assigned value and a chained assignment expression is essentially `a = (b = (c = 123))`.

syntax_and_semantics/assignment.md
```diff
@@ -62,3 +62,78 @@ objects[1] += 2        # same as: objects[1] = objects[1] + 2
...
+
+```crystal
+a = 1
+b = 2
+a, b = b, a
+a #=> 2
+b #=> 1
+```
+
+If the right-hand side contains just one expression, it is considered an indexed type and the following syntax sugar applies:
```
RX14: I'd rather say

>If ..., the type is indexed for each variable on the left-hand side like so:

PR #314, opened 2019-01-02
Clarify visibility of protected methods

straight-shoota:
Superseeds #236

/cc @philxyz

### DIFF
--- a/syntax_and_semantics/visibility.md
+++ b/syntax_and_semantics/visibility.md
@@ -118,16 +118,45 @@ end
 Namespace::Bar.new.bar
 ```
 
-A `protected` class method can be invoked from an instance method and the other way around:
+A `protected` method can only be invoked from the scope of its class or its descendants. That includes the class scope and bodies of class methods and instance methods of the same type the protected method is defined on, as well as all types including or inherinting that type and all types in that namespace.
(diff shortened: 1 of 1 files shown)


j8r:
A `module` can be added in examples.

straight-shoota:
I fear the example is already too big. Besides module there could also be struct. I think it's fine as it is.

syntax_and_semantics/visibility.md
```diff
@@ -118,16 +118,45 @@ end
...
+  def instance_method
+    Parent.protected_method # OK
+  end
+  
+  def self.class_method
+    Parent.protected_method # OK
+  end
+end
+
+class Parent::Sub
```
j8r: I mean replacing one of this classes by a module. There are already 3 classes

j8r: I mean replacing one of this classes by a module. There are already 3 classes

j8r: You know it, but new comers won't know systematically.

straight-shoota: They won't learn it either when this line is changed to `module`. On the contrary, it might create confusion.

PR #321, opened 2019-01-31
Document the @def macro instance variable

r00ster91:
This documents the `@def` macro instance variable which can be used to retrieve information about the method, the macro is in. Now all macro instance variables are documented.

### DIFF
--- a/syntax_and_semantics/macros.md
+++ b/syntax_and_semantics/macros.md
@@ -290,6 +290,24 @@ Foo.new.describe #=> "Class is Foo"
 Foo.describe #=> "Class is Foo"
 ```
 
+## Method information
+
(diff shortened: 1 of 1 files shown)


syntax_and_semantics/macros.md
```diff
@@ -290,6 +290,24 @@ Foo.new.describe #=> "Class is Foo"
 Foo.describe #=> "Class is Foo"
 ```
 
+## Method information
+
+When a macro is invoked you can access the method, the macro is in with a special instance variable: `@def`. The type of this variable is `Def` unless the macro is outside of a method, in this case it's `NilLiteral`.
```
straight-shoota: Could you add links to these types in the API docs?

r00ster91: Did that and also linked to all the other types.

PR #322, opened 2019-02-10
Extend next's documentation

### LINKED ISSUE #97: Minor documentation request for 'next'
See original request: https://github.com/crystal-lang/crystal/issues/4365

r00ster91:
Fixes #97 and documents that `next` can be used to exit from blocks.

### DIFF
--- a/syntax_and_semantics/next.md
+++ b/syntax_and_semantics/next.md
@@ -15,7 +15,7 @@ end
 # The above prints the numbers 2, 4 and 5
 ```
 
-But `next` can also be used to exit from blocks, for example:
+`next` can also be used to exit from a block, for example:
 
 ```crystal
 def block
@@ -31,7 +31,7 @@ end
 # The above prints the numbers 1 and 2
 ```
 
-Similar to [`break`](break.md), `next` can also take a parameter which will then be the value that gets returned:
+Similar to [`break`](break.md), `next` can also take a parameter which will then be returned by `yield`.
 
 ```crystal
 def block
--- a/syntax_and_semantics/next.md
+++ b/syntax_and_semantics/next.md
@@ -39,9 +39,8 @@ def block
 end
 
 block do
-  puts 1, 2
-  next 3
+  next "hello"
 end
 
-# The above prints the numbers 1, 2 and 3
+# The above prints "hello"
 ```
--- a/syntax_and_semantics/next.md
+++ b/syntax_and_semantics/next.md
@@ -23,12 +23,12 @@ def block
 end
 
 block do
-  puts 1, 2
+  puts "hello"
   next
-  puts 3
+  puts "world"
 end
 
-# The above prints the numbers 1 and 2
+# The above prints "hello"
 ```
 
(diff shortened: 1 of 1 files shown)


syntax_and_semantics/next.md
```diff
@@ -11,5 +11,39 @@ while a < 5
   end
   puts a
 end
+
 # The above prints the numbers 2, 4 and 5
 ```
+
+But `next` can also be used to exit from blocks, for example:
```
straight-shoota:
```suggestion
`next` can also be used to exit from a block, for example:
```

I'm not sure if `exit` is a great term here, but I don't have a better idea.

syntax_and_semantics/next.md
```diff
@@ -11,5 +11,39 @@ while a < 5
...
+
+Similar to [`break`](break.md), `next` can also take a parameter which will then be the value that gets returned:
+
+```crystal
+def block
+  yield
+end
+
+three = block do
+  puts 1, 2
```
straight-shoota: This example cane be simplified. There is no need for a variable, just add `puts yield` directly in `block`. `puts 1, 2` is also unnecessary.

syntax_and_semantics/next.md
```diff
@@ -11,5 +11,39 @@ while a < 5
...
+block do
+  puts 1, 2
+  next
+  puts 3
+end
+
+# The above prints the numbers 1 and 2
+```
+
+Similar to [`break`](break.md), `next` can also take a parameter which will then be the value that gets returned:
```
straight-shoota:
```suggestion
Similar to [`break`](break.md), `next` can also take a parameter which will then be returned by `yield`.
```

syntax_and_semantics/next.md
r00ster91: The `puts 1, 2` is part of the example. When I would remove it then just printing `3` doesn't make much sense. I would probably make it `"hello"` then and then it would look pretty redundant and useless:
```crystal
def block
  puts yield
end

block do
  next "hello"
end
```
*why isn't this just `puts "hello"`?*

But when also having `puts 1, 2` and then printing the last number using `next`, this thought is resolved I think.

straight-shoota: > why isn't this just puts "hello"?

Because the intention is to demonstrate the return value of `yield`. For this purpose you don't need to print `1, 2`. It's just unrelated noise.
Printing two numbers and then returning the next one using `next` doesn't have any useful meaning either.

Obviously, changing the value from `3` to a nice `"hello"` is better.

syntax_and_semantics/next.md
```diff
@@ -11,5 +11,36 @@ while a < 5
...
+
+`next` can also be used to exit from a block, for example:
+
+```crystal
+def block
+  yield
+end
+
+block do
+  puts 1, 2
```
RX14: Please make this `puts` print only one value, it feels weird to complicate the example by making readers recall that puts can take multiple arguments and prints them on separate lines.

PR #323, opened 2019-02-20
Clarify the non-existence of `public`

anicholson:
As raised by my confusion in [this issue][1].

[1]: https://github.com/crystal-lang/crystal/issues/7459

### DIFF
--- a/syntax_and_semantics/visibility.md
+++ b/syntax_and_semantics/visibility.md
@@ -1,8 +1,8 @@
 # Visibility
 
-Methods are public by default: the compiler will always let you invoke them, even if there is no `public` keyword.
+Methods are public by default: the compiler will always let you invoke them. In fact, there isn't even a `public` keyword in the language!
 
-Methods can be marked as `private` or `protected`.
+Methods can, however, be marked as `private` or `protected`.
 
 ## Private methods
 


syntax_and_semantics/visibility.md
```diff
@@ -1,8 +1,8 @@
 # Visibility
 
-Methods are public by default: the compiler will always let you invoke them, even if there is no `public` keyword.
+Methods are public by default: the compiler will always let you invoke them. In fact, there isn't even a `public` keyword in the language!
```
j8r: I prefered the previous sentence, which has a more neutral tone.
I suggest a simple and neutral: ``Methods are public by default: the compiler will always let you invoke them. There is no `public` keyword.``

syntax_and_semantics/visibility.md
```diff
@@ -1,8 +1,8 @@
...
 
-Methods can be marked as `private` or `protected`.
+Methods can, however, be marked as `private` or `protected`.
```
j8r: `however` suggests that having no `public` keyword is therefore a disadvantage. Not really.

syntax_and_semantics/visibility.md
anicholson: @j8r: Thanks for the feedback :)

You're correct about the surprised tone: I added it because it _is_ surprising that a language has `private` & `protected` _without_ also having `public` too.

Ruby, Python, Java, C, Objective-C, C++, C# & Swift all have `public` in their list of modifier syntax - and for many of those, `public` is the default. The only language I could trivially find that had `private` & `protected` without having `public` too was Scala.

I'm not saying it's better; I'm just saying it's _expected_ (which is why this issue arose for me in the first place!).

So I respectfully ask to leave that in.

syntax_and_semantics/visibility.md
anicholson: That's fair :) :+1:

anicholson:
@j8r I've addressed your feedback.

I have one further question: perhaps there should be an additional example about the implications of this in [the inheritance page][1]?

Specifically, if you want to override a private or protected method in a derived class _while maintaining its visibility_, you need to add the visibility modifier there too.

Something to the effect of:

```crystal
class Base

  def api_method
    internal_method
  end

  protected def internal_method
    :protected
  end
end

class LeakyDerived < Base
  def internal_method
    :public
  end
end

class CorrectlyDerived < Base
  protected def internal_method
    :still_protected
  end  
end

b = Base.new

b.internal_method         # Compiler error
b.api_method              # :protected

d = LeakyDerived.new

d.internal_method         # :public; Oops!
d.api_method              # :public

c = CorrectlyDerived.new

c.internal_method         # Compiler error as expected
c.api_method              # :still_protected
```

[1]: https://crystal-lang.org/reference/syntax_and_semantics/inheritance.html

PR #330, opened 2019-04-02
Add documentation for offsetof

straight-shoota:
Implemented in https://github.com/crystal-lang/crystal/pull/7589

/cc @malte-v @asterite

### DIFF
--- a/syntax_and_semantics/offsetof.md
+++ b/syntax_and_semantics/offsetof.md
@@ -1,0 +1,22 @@
+# offsetof
+
+The `offsetof` instruction returns the byte offset of an instance variable in a struct or class type. It accepts the type as first argument and the instance variable name prefixed by an `@` as second argument:
+
+```cr
+offsetof(Type, @ivar)
+```
+
+This is a low-level primitive and only useful if a C API needs to directly interface with the data layout of a Crystal type.
+It returns the value of [`LLVMOffsetOfElement`](http://llvm.org/doxygen/group__LLVMCTarget.html#ga9971347f4072d348862519bbacbd71a7).
+
+Example:
+```cr
+struct Foo
+  @x = 0_i64
+  @y = 34_i32
+end
+
+foo = Foo.new
+
+offsetof(typeof(foo), @y) # => 8
+```
--- a/syntax_and_semantics/offsetof.md
+++ b/syntax_and_semantics/offsetof.md
@@ -16,7 +16,5 @@ struct Foo
   @y = 34_i32
 end
 
-foo = Foo.new
-
-offsetof(typeof(foo), @y) # => 8
+offsetof(Foo, @y) # => 8
 ```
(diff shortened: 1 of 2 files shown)


malte-v:
`typeof` isn't allowed inside `offsetof`, so I would just use `offsetof(Foo, @y)` in the example.

syntax_and_semantics/offsetof.md
```diff
@@ -0,0 +1,20 @@
+# offsetof
+
+The `offsetof` instruction returns the byte offset of an instance variable in a struct or class type. It accepts the type as first argument and the instance variable name prefixed by an `@` as second argument:
```
r00ster91: Let's stay consistent and refer to them as "expressions", just like in the other 3 expression docs.
```suggestion
The `offsetof` expression returns the byte offset of an instance variable in a struct or class type. It accepts the type as the first argument and the instance variable name prefixed by an `@` as the second argument:
```

straight-shoota: I don't think we should describe them as "expressions". `offsetof` itself is not an expression in Crystal. A call like `offsetof(Type, @foo)` is an expression.
I'm open to alternative suggestions.

r00ster91: Keyword?

malte-v: I would go with "expression". `offsetof` nodes are transformed into number literals at compile time, which certainly are expressions. "Instruction" would be a bit misleading here since there is no actual instruction generated.

bew: Maybe ``An `offsetof` expression``, replacing `The` by `An` to show that it's not only `offsetof` but kinda an 'instance' of offsetof, like `offsetof(Foo, @bar)`

straight-shoota: @malte-v Yes, `offsetof` nodes are expressions, there's no debate. But this doesn't relate to a concrete expression as a call to `offsetof`, but the `offsetof` construct itself. It's similar to a method (or rather a macro method) but I don't think we have a specific name for this.

r00ster91: I would go with "Keyword" or "Primitive" (this term is used on https://crystal-lang.org/reference/syntax_and_semantics/low_level_primitives.html).

syntax_and_semantics/offsetof.md
```diff
@@ -0,0 +1,20 @@
...
+
+```cr
+offsetof(Type, @ivar)
+```
+
+This is a low-level primitive and only useful if a C API needs to directly interface with the data layout of a Crystal type.
+It returns the value of [`LLVMOffsetOfElement`](http://llvm.org/doxygen/group__LLVMCTarget.html#ga9971347f4072d348862519bbacbd71a7).
```
RX14: This detail isn't neccesary.

syntax_and_semantics/offsetof.md
bcardiff: I think that 

> An `offset` expression ... 

is accurate enough. We are not saying that the sole keyword is the expression.

Other than that this is GTG.

syntax_and_semantics/offsetof.md
RX14: ping?

PR #331, opened 2019-04-02
Add documentation for CRYSTAL_LIBRARY_PATH

straight-shoota:
Introduced by crystal-lang/crystal#7562

/cc @bcardiff

### DIFF
--- a/using_the_compiler/README.md
+++ b/using_the_compiler/README.md
@@ -49,9 +49,11 @@ The `--static` flag can be used to build a statically-linked executable:
 $ crystal build hello_world.cr --release --static
 ```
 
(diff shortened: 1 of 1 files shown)


bcardiff:
I'm ok with adding the env var description there, but I feel that a proper description of the lookup path and name conventions should probably be at ./syntax_and_semantics/c_bindings/lib.md . A link from one to the other might help. But without the context of a `lib` and `@[Link(..., static: true)]` is hard to fully know how to use it.

straight-shoota:
Yes, usage needs to be documented properly. I think this could be a separate PR from someone else. I'm not to confident in this area.

PR #332, opened 2019-04-02
Improve description of range literal

straight-shoota:
Also includes documentation of begin-less and end-less ranges introduced by https://github.com/crystal-lang/crystal/pull/7179

/cc @asterite

### DIFF
--- a/syntax_and_semantics/literals/range.md
+++ b/syntax_and_semantics/literals/range.md
@@ -10,7 +10,7 @@ A [Range](http://crystal-lang.org/api/Range.html) represents an interval between
 (0...5).to_a # => [0, 1, 2, 3, 4]
 ```
 
-NOTE: Range literals are often wrapped in parentheses, for example if it is meant to be used as the receiver of a call. `0..5.to_a` without parentheses would be semantically equivalent to `0..(5.to_a)` because method calls and other operators have higher precedence than the range literal.
+**NOTE:** Range literals are often wrapped in parentheses, for example if it is meant to be used as the receiver of a call. `0..5.to_a` without parentheses would be semantically equivalent to `0..(5.to_a)` because method calls and other operators have higher precedence than the range literal.
 
 An easy way to remember which one is inclusive and which one is exclusive it to think of the extra dot as if it pushes *y* further away, thus leaving it outside of the range.
 
--- a/syntax_and_semantics/literals/range.md
+++ b/syntax_and_semantics/literals/range.md
@@ -1,6 +1,6 @@
 # Range
 
-A [Range](http://crystal-lang.org/api/Range.html) represents an interval between two values. It is typically constructed with a range literal, consisting of two `..` or three dots:
+A [Range](http://crystal-lang.org/api/Range.html) represents an interval between two values. It is typically constructed with a range literal, consisting of two or three dots:
 
 * `x..y`: Two dots denote an inclusive range, including `x` and `y` and all values in between (in mathematics: `[x, y]`) .
 * `x...y`: Three dots denote an exclusive range, including `x` and all values up to but not including `y` (in mathematics: `[x, y)`).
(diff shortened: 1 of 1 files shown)


syntax_and_semantics/literals/range.md
```diff
@@ -1,14 +1,32 @@
 # Range
 
-A [Range](http://crystal-lang.org/api/Range.html) is typically constructed with a range literal:
+A [Range](http://crystal-lang.org/api/Range.html) represents an interval between two values. It is typically constructed with a range literal, consisting of two `..` or three dots:
 
-```crystal
-x..y  # an inclusive range, in mathematics: [x, y]
-x...y # an exclusive range, in mathematics: [x, y)
+* `x..y`: Two dots denote an inclusive range, including `x` and `y` and all values in between (in mathematics: `[x, y]`) .
```
r00ster91: "two dots" and "three dots" are already mentioned above and you see them right before the description. I think this can be kept shorter here.
```suggestion
* `x..y`: Denotes an inclusive range, including `x` and `y` and all values in between (in mathematics: `[x, y]`) .
```

syntax_and_semantics/literals/range.md
```diff
@@ -1,14 +1,32 @@
...
+* `x...y`: Three dots denote an exclusive range, including `x` and all values up to but not including `y` (in mathematics: `[x, y)`).
```
r00ster91:
```suggestion
* `x...y`: Denotes an exclusive range, including `x` and all values up to but not including `y` (in mathematics: `[x, y)`).
```

syntax_and_semantics/literals/range.md
```diff
@@ -1,14 +1,32 @@
...
 
-# Example:
+```cr
 (0..5).to_a # => [0, 1, 2, 3, 4, 5]
 (0...5).to_a # => [0, 1, 2, 3, 4]
 ```
 
+NOTE: Range literals are often wrapped in parenthesis, for example if it is meant to be used as the receiver of a call. `0..5.to_a` without parenthesis would be semantically equivalent to `0..(5.to_a)` because method calls and other operators have higher precedence than the range literal.
```
r00ster91: I don't think `NOTE:` should be used here too because it doesn't do anything and has no markdown.
```suggestion
Note that range literals are often wrapped in parentheses, for example if it is meant to be used as the receiver of a call. `0..5.to_a` without parentheses would be semantically equivalent to `0..(5.to_a)` because method calls and other operators have higher precedence than the range literal.
```

syntax_and_semantics/literals/range.md
```diff
@@ -1,14 +1,32 @@
 # Range
 
-A [Range](http://crystal-lang.org/api/Range.html) is typically constructed with a range literal:
+A [Range](http://crystal-lang.org/api/Range.html) represents an interval between two values. It is typically constructed with a range literal, consisting of two `..` or three dots:
```
r00ster91: When there's `` `..` `` after "two", why isn't there `` `...` `` after "three"? I think these can be omitted. You already see right below what two and three dots look like.
```suggestion
A [Range](http://crystal-lang.org/api/Range.html) represents an interval between two values. It is typically constructed with a range literal, consisting of two or three dots:
```

syntax_and_semantics/literals/range.md
straight-shoota: I disagree. I'd rather use full sentences here.

syntax_and_semantics/literals/range.md
straight-shoota: Yes, `NOTE` doesn't have a special syntactic meaning, but it's nevertheless a useful admonition. We're using this in other locations as well.

syntax_and_semantics/literals/range.md
```diff
@@ -1,14 +1,32 @@
...
+* `x..y`: Two dots denote an inclusive range, including `x` and `y` and all values in between (in mathematics: `[x, y]`) .
+* `x...y`: Three dots denote an exclusive range, including `x` and all values up to but not including `y` (in mathematics: `[x, y)`).
 
-# Example:
+```cr
 (0..5).to_a # => [0, 1, 2, 3, 4, 5]
 (0...5).to_a # => [0, 1, 2, 3, 4]
 ```
 
+NOTE: Range literals are often wrapped in parentheses, for example if it is meant to be used as the receiver of a call. `0..5.to_a` without parentheses would be semantically equivalent to `0..(5.to_a)` because method calls and other operators have higher precedence than the range literal.
```
r00ster91: Can we at least manually add some markdown then? To me it looks so noisy and useless when a markdown-like expression such as "NOTE:" is used but doesn't actually do anything.
```suggestion
**NOTE:** Range literals are often wrapped in parentheses, for example if it is meant to be used as the receiver of a call. `0..5.to_a` without parentheses would be semantically equivalent to `0..(5.to_a)` because method calls and other operators have higher precedence than the range literal.
```

PR #336, opened 2019-04-23
Document the useage of &. with an argument

### LINKED ISSUE #328: Document &. with argument notation
In the `Short one-argument syntax` section.

Ref: https://github.com/crystal-lang/crystal/issues/7549

Blacksmoke16:
Resolves #328

### DIFF
--- a/syntax_and_semantics/blocks_and_procs.md
+++ b/syntax_and_semantics/blocks_and_procs.md
@@ -179,7 +179,7 @@ and
 method { |argument| argument.some_method }
 ```
 
-Can both be written as:
+can both be written as:
 
 ```crystal
 method &.some_method
--- a/syntax_and_semantics/blocks_and_procs.md
+++ b/syntax_and_semantics/blocks_and_procs.md
@@ -163,7 +163,9 @@ The block variable `second` also includes the `Nil` type because the last `yield
 
 ## Short one-argument syntax
 
-A short syntax exists for specifying a block that receives a single argument and invokes a method on it. This:
+If a block has a single argument and invokes a method on it, the block can be replaced with the short syntax argument.
+
+This:
 
 ```crystal
 method do |argument|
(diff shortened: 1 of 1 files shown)


syntax_and_semantics/blocks_and_procs.md
```diff
@@ -198,6 +198,18 @@ method &.+(2)
 method &.[index]
 ```
 
+If the block is not the only parameter being passed to the method, the short syntax argument must be supplied within the argument list.
```
straight-shoota:
```suggestion
If the block is not the only parameter being passed to the method, the short syntax argument can be supplied within the argument list.
```

syntax_and_semantics/blocks_and_procs.md
```diff
@@ -198,6 +198,18 @@ method &.+(2)
...
+
+```crystal
+["a", "b"].join(",") { |s| s.upcase }
```
straight-shoota: I don't think we really need this example, but if we keep it, it should go after the other one which illustrates the actual feature.

syntax_and_semantics/blocks_and_procs.md
Blacksmoke16: Why not `must` here?  Since it was _wasn't_ supplied in the argument list it wouldn't error, like the issue https://github.com/crystal-lang/crystal/issues/7549

straight-shoota: I reconsidered and think this sentence and line 166 need to be changed entirely. The short syntax argument is *always* supplied in the argument list. If there are no other arguments, it's the *only* element in that list, but it's always a normal argument. This should be clarified.
Given `foo &.bar` and `foo { |f| f.bar }`, it looks like `&.bar` is syntactically equivalent to `{ |f| f.bar }`, but in fact it's not. `&.bar` is a regular argument while `{ |f| f.bar }` is a block. This is obvious when adding the optional parenthesis: `foo(&.bar)` vs. `foo() { |f| f.bar }`

Blacksmoke16:
@straight-shoota I reworked the wording/examples in the section to make it more clear that it is simply an argument that is equivalent to the block.

If you would review again.

syntax_and_semantics/blocks_and_procs.md
```diff
@@ -163,39 +163,58 @@ The block variable `second` also includes the `Nil` type because the last `yield
...
 ```
 
-Can be written as this:
+and
+
+```crystal
+method { |argument| argument.some_method }
+```
+
+Can both be written as:
```
r00ster91: "This", the code block, the "and", the other code block and these words are still one sentence so this shouldn't be capitalized.
```suggestion
can both be written as:
```

PR #312, opened 2018-12-30
added a troubleshooting section for macOS mojave

dustinlahr:
I ran into this issue recently trying out crystal and hoped to contribute!

[crystal-7209](https://github.com/crystal-lang/crystal/issues/7209)

PR #337, opened 2019-05-01
Update instructions for homebrew on Macs with Mojave

bohtho:
Updated installation instructions for homebrew on Mac to include workaround for Mojave, which has been out since September 24, 2018 for macs from mid 2012 and newer, as per https://forum.crystal-lang.org/t/noob-error-in-brew-installation-on-mac/714/6?u=thbo and https://github.com/crystal-lang/crystal/pull/7494.


straight-shoota:
There is already #312 for this, it just needs another approval.

bohtho:
Yeah, thanks for the heads up, but I´ll keep it up a little while longer, as #337 seems to be seen as more controversial, whilst this is shorter, to the point and entirely consistent with the existing troubleshooting text for El Capitan.

PR #340, opened 2019-05-21
Fix the module instantiation example

smadeja:
The supplied compiler output is wrong and confusing because Class actually has a 'new' method.


straight-shoota:
For the record: `Class` has a `.new` method, but `Moo:Class` (essentially `Moo.class`) does not.

The change is correct anyway to fix the compiler output.

smadeja:
Thanks for the clarification! :-)

PR #350, opened 2019-06-25
Clarify limitation of type restrictions to local variables

bcardiff:
Closes https://github.com/crystal-lang/crystal/issues/7847

### DIFF
--- a/syntax_and_semantics/if_var.md
+++ b/syntax_and_semantics/if_var.md
@@ -35,7 +35,7 @@ Of course, reassigning a variable inside the `then` branch makes that variable h
 
 ## Limitations
 
-The above logic **doesn’t** work with instance variables, class variables and variables bound in a closure. The value of these kinds of variables could potentially be affected by another fiber after the condition was checked, rendering it `nil`.
+The above logic works **only for local variables**. It doesn’t work with instance variables, class variables, or variables bound in a closure. The value of these kinds of variables could potentially be affected by another fiber after the condition was checked, rendering it `nil`. It also does not work with constants.
 
 ```crystal
 if @a


syntax_and_semantics/if_var.md
```diff
@@ -35,7 +35,7 @@ Of course, reassigning a variable inside the `then` branch makes that variable h
 
 ## Limitations
 
-The above logic **doesn’t** work with instance variables, class variables and variables bound in a closure. The value of these kinds of variables could potentially be affected by another fiber after the condition was checked, rendering it `nil`.
+The above logic works **only for local variables**. It doesn’t work with instance variables, class variables, variables bound in a closure and constants. The value of these kinds of variables could potentially be affected by another fiber after the condition was checked, rendering it `nil`.
```
bew: > The value of these kinds of variables could potentially be affected by another fiber after the condition was checked

This is wrong for constants though.. since they can't be changed

bcardiff: > The value of these kinds of **variables**

bew: yeeah right ok ^^

bcardiff:
build failed because llvm.org is having some issues 🤷‍♂

obskyr:
Hmm. I also think that the phrasing is a bit confusing with the "the value of these kinds of variables […]" placed as it is. How about this instead?

> The above logic works **only for local variables**. It doesn’t work with instance variables, class variables, or variables bound in a closure. The value of these kinds of variables could potentially be affected by another fiber after the condition was checked, rendering it `nil`. It also does not work with constants.

syntax_and_semantics/if_var.md
```diff
@@ -35,7 +35,7 @@ Of course, reassigning a variable inside the `then` branch makes that variable h
 
 ## Limitations
 
-The above logic **doesn’t** work with instance variables, class variables and variables bound in a closure. The value of these kinds of variables could potentially be affected by another fiber after the condition was checked, rendering it `nil`.
+The above logic works **only for local variables**. It doesn’t work with instance variables, class variables, or variables bound in a closure. The value of these kinds of variables could potentially be affected by another fiber after the condition was checked, rendering it `nil`. It also does not work with constants.
```
RX14: Why not just add constants to the existing list?

obskyr: As briefly discussed above, the reason not to do that is that the next sentence goes on to say that those kinds of variables can be affected by another fiber, a rationale which isn't valid for constants.

RX14: my bad

PR #352, opened 2019-07-30
Gender neutral pronoun

dancojocaru2000:
Ye, I know, not a big deal, but in today's society, who knows?


RX14:
Singular "they" is good!

PR #354, opened 2019-08-01
Document the new Annotation methods

Blacksmoke16:


syntax_and_semantics/annotations.md
```diff
@@ -78,6 +78,39 @@ end
...
 
+The `named_args` method can be used to read all key/value pairs on an annotation as a `NamedTupleLiteral`.
+
+```crystal
+annotation MyAnnotation
+end
+
+@[MyAnnotation(value: 2, name: "Jim")]
+def annotation_named_args
+  {{@def.annotation(MyAnnotation).named_args}}
```
straight-shoota: I think the formatter would go like this:
```suggestion
  {{ @def.annotation(MyAnnotation).named_args }}
```

syntax_and_semantics/annotations.md
```diff
@@ -78,6 +78,39 @@ end
 annotation_value # => 2
 ```
 
+The `named_args` method can be used to read all key/value pairs on an annotation as a `NamedTupleLiteral`.
```
straight-shoota: Maybe add that these don't need to be declared in the annotation type and that different instantiations (is that the correct term?) of an annotation can have different arguments.
Ditto for `args`.

syntax_and_semantics/annotations.md
```diff
@@ -78,6 +78,39 @@ end
...
+```crystal
+annotation MyAnnotation
+end
+
+@[MyAnnotation(value: 2, name: "Jim")]
+def annotation_named_args
+  {{@def.annotation(MyAnnotation).named_args}}
+end
+
+puts annotation_named_args # => {value: 2, name: "Jim"}
```
straight-shoota:
```suggestion
annotation_named_args # => {value: 2, name: "Jim"}
```

syntax_and_semantics/annotations.md
```diff
@@ -78,6 +78,39 @@ end
...
+```
+
+This can also be used within macros.
```
straight-shoota: Isn't that obvious? I mean, does this second example provide a meaningful benefit? I see it more as unnecessary clutter.
Alternatively we could scrap the first one and only show this with more stuff, but I doubt that's really necessary.

syntax_and_semantics/annotations.md
Blacksmoke16: Formatter doesn't currently do anything to it.

Blacksmoke16:
@straight-shoota How's this?

syntax_and_semantics/annotations.md
straight-shoota: It probably should... I'd prefer it that way and equal to all `{%` macro tags.

syntax_and_semantics/annotations.md
```diff
@@ -89,26 +89,27 @@ def annotation_named_args
   {{@def.annotation(MyAnnotation).named_args}}
 end
 
-puts annotation_named_args # => {value: 2, name: "Jim"}
+annotation_named_args # => {value: 2, name: "Jim"}
 ```
 
-This can also be used within macros.
+Since this method returns a `NamedTupleLiteral`, all of the [methods](https://crystal-lang.org/api/Crystal/Macros/NamedTupleLiteral.html) on that type are available for use.  Especially `#double_splat` which makes it easy to pass annotation arguments to classes.
```
straight-shoota: Do you mean "to methods"?

syntax_and_semantics/annotations.md
Blacksmoke16: Indeed.  Would be nice.

PR #360, opened 2019-09-28
Update http_server.md to use Time.local

TillSchroder:
When running this example I get:
>Warning: Deprecated Time.now. Use `Time.local` or `Time.utc` instead.

It works without warning when using Time.local instead.


bcardiff:
Thanks @TillSchroder!

PR #361, opened 2019-09-28
Update integers.md to mention Int128 and UInt128

roetlich:
I noticed that Crystal supports 128 bit integers (nice!), but they aren't in the documentation.

Are 128 bit ints supported on all platforms?

### DIFF
--- a/syntax_and_semantics/literals/integers.md
+++ b/syntax_and_semantics/literals/integers.md
@@ -1,6 +1,6 @@
 # Integers
 
-There are four signed integer types, and four unsigned integer types:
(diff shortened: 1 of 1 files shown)


bcardiff:
Its support is not yet complete. 128bits Literals are not supported still, and when using a 32-bits environment the compiler does not emit all the needed compiler-rt methods.

Once that is changes we can merge this.

straight-shoota:
Alternative PR: #361

beta-ziliani:
Closing since the related PR (with modifications) was merged.

PR #362, opened 2019-09-30
Update http_server.md

Yashko:
In server.cr:5:59
 5 | context.response.print "Hello world! The time is #{Time.now}"
                                                             ^--
Warning: Deprecated Time.now. Use `Time.local` or `Time.utc` instead.
A total of 1 warnings were found.
Listening on http://127.0.0.1:8080


bcardiff:
Thanks @Yashko, the same was noticed in #360 with a couple of more cases.

PR #364, opened 2019-10-01
Create a small guide for Collections

rebuilding-patrick:
Currently many common functions for working with Collections are hidden in the standard library without any reference. This is a short guide that aims to show how to use those functions.


straight-shoota:
This is a good starting point. But I'd suggest to focus primarily on `Indexable`, `Enumerable` and `Iterable` and actually explain their use cases and differences.

PR #365, opened 2019-10-03
you can test without colors

ilanpillemer:
Took me hours to find this as I am using acme editor which meant I was getting all the all the color commands as junk and you cannot turn off through `export TERM=dumb`. Adding it to the gitbook so others dont struggle.

### DIFF
--- a/guides/testing.md
+++ b/guides/testing.md
@@ -116,9 +116,14 @@ By convention, specs live in the `spec/` directory of a project. Spec files must
 
 You can compile and run specs from folder trees, individual files or specific lines in a file.
 
+You can turn off colors with the switch `--no-color`.
+
 ```bash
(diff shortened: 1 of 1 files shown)


RX14:
Should we be detecting certain `TERM` values as not supporting colour?

ilanpillemer:
- yes, I think so.. see https://github.com/crystal-lang/crystal/issues/8261

guides/testing.md
```diff
@@ -116,10 +116,15 @@ By convention, specs live in the `spec/` directory of a project. Spec files must
 
 You can compile and run specs from folder trees, individual files or specific lines in a file.
 
+You can turn off colors with the switch `--no-color`
```
straight-shoota:
```suggestion
You can turn off colors with the switch `--no-color`.
```

Issue #366, opened 2019-10-05
Add wrapping operators


Blacksmoke16:
With overflow exceptions turned on by default and finally released, we should add some docs for the wrapping operators. I.e. `&+`, `&**`, etc.

Issue #261, opened 2018-07-16
Move away from Gitbook


straight-shoota:
If you want to build `crystal-book` locally, just setting up a working environment is not very easy.
Just installing the dependencies - about 140+ npm plugins - takes quite a long time. Then you need to run `npm install` **and** `gitbook install` to install all gitbook plugins. Running `npm install` again after `gitbook install` removes some dependencies....

It seems like the open source version of gitbook (and the CLI) is pretty much abandoned (see https://github.com/GitbookIO/gitbook/issues/1808) and hasn't seen any substantial updates for two years.

There are good alternative documentation generators out there, and we can certainly find something that fits.

Obviously, such a change does not necessarily provide direct benefits for readers of the documentation. But it should make it easier for contributers and help stay on path for the long run.

j8r:
Even more if there is one in Crystal, :100:% for it!
We could have one derived from `crystal docs`.

straight-shoota:
@madranet Could you please open a new issue for this? This is a different problem and not entirely dependent on switching away from Gitbook as it can at least partially be fixed with the current tools available.

madranet:
@straight-shoota 

Done: https://github.com/crystal-lang/crystal-book/issues/282

straight-shoota:
I've been experimenting with switching to Jekyll as site generator to resolve this and other issues. So far, it works really good :+1:
Jekyll feels like a natural option because it's very popular, there are lots of resources and we're already using it in https://github.com(crystal-lang/crystal-website
While Jekyll's initial focus is on blogging, it is very flexible and has been proven for building other things such as documentation sites. There are lot's of examples for that. One that I find appealing and has a similar feature set as we need is https://docs.sentry.io/ They use Jekyll with a [minimal-mistakes](https://mmistakes.github.io/minimal-mistakes) theme plus a few customizations. It looks really nice. 

So I've just initialized a new Jekyll project, moved the existing content to a collection folder and that's how it looks with minimal-mistakes theme:
![image](https://user-images.githubusercontent.com/466378/73869453-91da3d00-484a-11ea-9de0-0570c105dbae.png)

It pretty much works out of the box with minimal changes to the existing content. I only had to replace the h1 title in the first line of each markdown file with a Jekyll frontmatter.

* lunr.js-based search works out of the box. It still experiences issues similar to #242 but the search engine is plugable. Though I doubt there's a feasible alternative to lunr.js
* In-page table of contents works out of the box
[...]

PR #370, opened 2019-10-16
Rework operators

straight-shoota:


syntax_and_semantics/operators.md
```diff
@@ -1,18 +1,35 @@
...
-is the same as:
+There are however two special rules regarding operators syntax:
 
-```crystal
-a.+(b)
-```
+* The `.` between receiver and method name (i.e. the *operator*) can be omitted.
+* Chained operators are modified to enforce operator precedence.
+
+Enforcing operator preference makes sure that an expression such as `1 * 2 + 3 * 4`
```
bcardiff:
```suggestion
Enforcing operator precedence makes sure that an expression such as `1 * 2 + 3 * 4`
```

syntax_and_semantics/operators.md
```diff
@@ -21,71 +38,201 @@ struct Vector2
...
+| `&*` | multiplication (overflowing) | `1 &* 2` | yes |
+| `/` | division | `1 / 2` | yes |
+| `//` | floor division | `1 // 2` | yes |
+| `%` | modulus | `1 % 2` | yes |
+
+#### Additive
+| Operator | Description | Example | Overloadable |
+|---|---|---|---|
+| `+` | addition | `1 + 2` | yes |
+| `&+` | addition (overflowing) | `1 &+ 2` | yes |
```
bcardiff: I think is either `Overflow addition` (swift uses that term) or `wrapping addition`.  It's weird to me that the overflowing addition does not raise overflow (but it might be my Spanish getting in the middle).

syntax_and_semantics/operators.md
```diff
@@ -21,71 +38,201 @@ struct Vector2
...
+| `-` | subtraction | `1 - 2` | yes |
+| `&-` | subtraction (overflowing) | `1 &- 2` | yes |
+
+### Other unary operators
+| Operator | Description | Example | Overloadable |
+|---|---|---|---|
+| `!` | inversion | `!true` | no |
+| `~` | binary complement | `~1` | yes |
+
+### Shifts
```
bcardiff:
```suggestion
### Shifts / Streams
```

maybe...

syntax_and_semantics/operators.md
```diff
@@ -21,71 +38,201 @@ struct Vector2
...
+| `..` | range | `1..10` | no |
+| `...` | exclusive range | `1...10` | no |
+
+### Splats
+| Operator | Description | Example | Overloadable |
+|---|---|---|---|
+| `*` | splat | `*foo` | no |
+| `**` | double splat | `**foo` | no |
+
+### Block
```
bcardiff: I'm not sure the Blocks section belongs here. Because they are not operators but syntax that use symbols.

syntax_and_semantics/operators.md
```diff
@@ -21,71 +38,201 @@ struct Vector2
...
+| `=` | assignment | `` | no |
+| `[]=` | index assignment | `` | yes |
+| `+=` | addition *and* assignment | `` | yes |
+| `&+=` | addition (overflowing) *and* assignment | `` | yes |
+| `-=` | subtraction *and* assignment | `` | yes |
+| `&-=` | subtraction (overflowing) *and* assignment | `` | yes |
+| `*=` | multiplication *and* assignment | `` | yes |
+| `&*=` | multiplication (overflowing) *and* assignment | `` | yes |
+| `/=` | division *and* assignment | `` | yes |
+| `//=` | floor *and* division assignment | `` | yes |
```
bcardiff:
```suggestion
| `//=` | floor division *and* assignment | `` | yes |
```

bcardiff:
I'm not sure what are you imagining for the rest of the section, but maybe mentioning something about the typing for binary operators will make sense: 

* `typeof(a <op> b) == typeof(a)` so that `a <op>= b` avoids unions.
* yet there could be exceptions, like `//`.

Or maybe this and the overflow handling should be in an arithmetics section on their own.

syntax_and_semantics/operators.md
straight-shoota: Theoretically they're are operators and so are `.`, `::` and others. I'm not yet sure how to handle all this.

Blacksmoke16:
This essentially covers #366, just a heads up.

straight-shoota:
Unfortunately the tables are broken. It seems the gitbook markdown parser doesn't understand escaping pipe characters inside tables (`\|`).
So unless we find some kind of workaround, I need to transform all tables including operators with `|` to HTML. 😢 

This makes me want #261 even more.

syntax_and_semantics/operators.md
```diff
@@ -1,18 +1,42 @@
...
+similar semantics to the generic operator meaning to avoid cryptic code that is
+confusing and behaves unexpectedly.
 
-You can define an operator for a type like this:
+### Unary operators
+
+Unary operators have only a single operand. Thus, a method implementation has no
+arguments and only operates on the first operand, the receiver `self`.
+
+The following example demonstrates the `Vector2` type as a two-dimensional vector
```
kinxer: Why is there now an example only for defining a unary operator but not for defining a binary operator?

syntax_and_semantics/operators.md
```diff
@@ -21,101 +45,271 @@ struct Vector2
...
-!   # not
-~   # bitwise complement
-```
+Binary operators have two operands. Thus, a method implementation receives
+exactly one argument for the second operator. The first operand is the receiver
+`self`.
 
-These are defined without arguments. For example
+The following example demonstrates the `Vector2` type as a two-dimensional vector
+with with a binary operator method `+` for vector addition.
```
kinxer: You have a double "with" here.

syntax_and_semantics/operators.md
```diff
@@ -21,101 +45,271 @@ struct Vector2
...
 
-For example:
+## List of Operators
 
-```crystal
-class MyArray
-  def [](index)
-    # ...
-  end
+This list is sorted by precedence, upper entries bind stronger than lower ones.
```
kinxer: This would read more smoothly with a conjunction ("so", perhaps) after the comma.

asterite:
I just wanted to chime in and say "THANK YOU" for improving the language reference. Many times me or others send PRs but we forgot to mention things here, and the docs are super important for everyone to get on board with the language.

syntax_and_semantics/operators.md
```diff
@@ -21,101 +45,272 @@ struct Vector2
...
+## Binary operators
 
-```crystal
-+   # positive
--   # negative
-!   # not
-~   # bitwise complement
-```
+Binary operators have two operands. Thus, a method implementation receives
+exactly one argument for the second operator. The first operand is the receiver
```
vlazar:
```suggestion
exactly one argument for the second operand. The first operand is the receiver
```

syntax_and_semantics/operators.md
```diff
@@ -21,101 +45,272 @@ struct Vector2
...
+| `\|` | binary OR | `1 \| 2` | yes |
+| `^` | binary XOR | `1 ^ 2` | yes |
+
+### Equality
+
+There are three base operators for testing equality:
+
+* `==`: Checks whether the values of the operands are equal
+* `=~`: Checks whether the value of the first operand matches the value of the
+second operator with pattern matching.
```
vlazar:
```suggestion
second operand with pattern matching.
```

syntax_and_semantics/operators.md
```diff
@@ -1,18 +1,42 @@
...
+
+One can assign any meaning to the operators, but it is advisable to stay within
+similar semantics to the generic operator meaning to avoid cryptic code that is
+confusing and behaves unexpectedly.
 
-You can define an operator for a type like this:
+### Unary operators
+
+Unary operators have only a single operand. Thus, a method implementation has no
+arguments and only operates on the first operand, the receiver `self`.
```
RX14: just say it only operates on `self`.

syntax_and_semantics/operators.md
```diff
@@ -21,101 +45,272 @@ struct Vector2
...
+There are no real ternary operators in Crystal.
 
-```crystal
-[]  # array index (raises on out of bounds)
-[]? # array index (nil on out of bounds)
-[]= # array index assignment
-```
+The conditional operator (`? :`) however, is technically a ternary operator.
+But it is not implemented as such, the compiler simply transforms it to an `if`
+condition.
```
RX14: This doesn't mean anything in that it doesn't have any practical effect...

straight-shoota: I agree. Still want to mention it because you might look for ternary if under operators. But it could be a shorter reference. Suggestions welcome.

syntax_and_semantics/operators.md
```diff
@@ -21,101 +45,271 @@ struct Vector2
...
+
+| Operator | Description | Example | Overloadable |
+|---|---|---|---|
+| `&` | binary AND | `1 & 2` | yes |
+| `\|` | binary OR | `1 \| 2` | yes |
+| `^` | binary XOR | `1 ^ 2` | yes |
+
+### Equality
+
+There are three base operators for testing equality:
```
j8r: Simpler:
```suggestion
Three base operators test equality:
```

syntax_and_semantics/operators.md
```diff
@@ -21,101 +45,271 @@ struct Vector2
...
+| `>>=` | right shift *and* assignment | `` | no |
+| `\|\|=` | logical or *and*  assignment | `` | no |
+| `&&=` | logical and *and* assignment | `` | no |
 
-## Meaning
+### Index Accessors
 
-One can assign any meaning to the operators, but the convention is to follow the above ones to avoid cryptic code, or code that behaves in an unexpected way.
+Index accessors are used to query a value by index or key, for example an array
+item or map entry. The nilable variant `[]?` is supposed to return `nil` when
```
j8r: Supposed? There are cases it won't be true?

straight-shoota: Depends entirely on the implementation.

syntax_and_semantics/operators.md
RX14: I still don't see the problem with calling it what it is: a ternary operator.

straight-shoota: Not sure I understand what you want me to do.

RX14: Remove the false claim that "There are no ternary operators in Crystal."

syntax_and_semantics/operators.md
```diff
@@ -21,101 +45,271 @@ struct Vector2
...
+This list is sorted by precedence, so upper entries bind stronger than lower
+ones.
 
-  def [](index1, index2, index3)
-    # ...
-  end
+| Category        | Operators                                            |
+|-----------------|------------------------------------------------------|
+| Index accessors | `[]`, `[]?`                                          |
+| Unary           | `+`, `&+`, `-`, `&-`, `!`, `~`, `*`, `**`,           |
```
bcardiff:
```suggestion
| Unary           | `+`, `&+`, `-`, `&-`, `!`, `~`                       |
| Splats          | `*`, `**`                                            |
```

I found it confusing that spalts, and double splats are included in this table as unary operators but they are not present in the section bellow as unary operators.

syntax_and_semantics/operators.md
```diff
@@ -21,101 +45,271 @@ struct Vector2
...
+| `^` | binary XOR | `1 ^ 2` | yes |
+
+### Equality
+
+Three base operators test equality:
+
+* `==`: Checks whether the values of the operands are equal
+* `=~`: Checks whether the value of the first operand matches the value of the
+second operand with pattern matching.
+* `===`: Checks whether the values of the operands are equal in [case equality](case.html).
```
bcardiff: Maybe it's worth noticing that `===` is not symmetric. `t === u` calls `T#===(u : U)` and means `u` belongs or matches `t`.

straight-shoota: No operator in Crystal is intrinsically commutative. Even `A#==(B)` can be different from `B#==(A)`, depending on the implementation. That should be avoided because it's confusing, but it is possible.
But yeah, I'll try to improve the wording for `===`. It's excplicitly *not* desgined as commutative.

syntax_and_semantics/operators.md
straight-shoota: There is actually a section about splats.

syntax_and_semantics/operators.md
RX14: This still hasn't been done...

syntax_and_semantics/operators.md
```diff
@@ -21,101 +52,273 @@ struct Vector2
...
+| `&&=` | logical and *and* assignment | `` | no |
 
-## Meaning
+### Index Accessors
 
-One can assign any meaning to the operators, but the convention is to follow the above ones to avoid cryptic code, or code that behaves in an unexpected way.
+Index accessors are used to query a value by index or key, for example an array
+item or map entry. The nilable variant `[]?` is supposed to return `nil` when
+the index is not found, while the non-nilable variant raises in that case.
+Implementations in the standard-library usually raise [`KeyError`](https://crystal-lang.org/api/KeyError.html)
```
bcardiff:
```suggestion
Implementations in the standard-library usually raise [`KeyError`](https://crystal-lang.org/api/latest/KeyError.html)
```

syntax_and_semantics/operators.md
```diff
@@ -21,101 +52,273 @@ struct Vector2
...
+or [`IndexError`](https://crystal-lang.org/api/IndexError.html).
```
bcardiff:
```suggestion
or [`IndexError`](https://crystal-lang.org/api/latest/IndexError.html).
```

RX14:
@straight-shoota i commented above too

syntax_and_semantics/operators.md
straight-shoota: Done.

syntax_and_semantics/operators.md
```diff
@@ -21,101 +53,467 @@ struct Vector2
...
-* `<=` – less or equal
-* `>` – greater
-* `>=` – greater or equal
-* `<=>` – comparison
-* `===` – [case equality](case.html)
-
-## Indexing
+The [conditional operator (`? :`)](./ternary_if.md) is the only ternary
+operator. It is transformed to an `if` expression by the compiler, so there are
+actually no ternary operator methods.
```
RX14: Perhaps something like

> It is not parsed as a method, and it's meaning cannot be changed

syntax_and_semantics/operators.md
```diff
@@ -21,101 +53,467 @@ struct Vector2
...
+<th>Example</th>
+<th>Overloadable</th>
+</tr>
+</thead>
+<tbody>
+<tr>
+<td><code>&amp;&amp;</code></td>
+<td><a href="and.html">logical AND</a></td>
+<td><code>true &amp;&amp; false</code></td>
+<td>yes</td>
```
bcardiff: I just noticed that it wrongly says that logical operators are overloadable. The compiler defines the typing and semantics of these operators.

PR #373, opened 2019-11-01
Add section on macro evaluation within comments.

KimBurgess:
Add further details on macro processing, specifically within comments. Provides an example of where this may be used, and areas where it may cause confusions for new comers (e.g. me).

Based on discussion in crystal-lang/crystal#8408.

### DIFF
--- a/syntax_and_semantics/macros.md
+++ b/syntax_and_semantics/macros.md
@@ -414,6 +414,33 @@ greeting_for_bob "hola"     # => "¡hola bob!"
 
 Notice the variables in the inner macro are not available within the `verbatim` block. The contents of the block are transferred "as is", essentially as a string, until re-examined by the compiler.
 
+## Comments
+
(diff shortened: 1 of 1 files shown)


KimBurgess:
Sorry about the second update after approval - just fixed up a typo.

syntax_and_semantics/macros.md
```diff
@@ -414,6 +414,33 @@ greeting_for_bob "hola"     # => "¡hola bob!"
...
+```crystal
+{% for name, index in ["foo", "bar", "baz"] %}
+  # Provides a placeholder {{name.id}} method. Always returns {{index}}.
+  def {{name.id}}
+    {{index}}
+  end
+{% end %}
+```
+
+This evaluation applies to both interpolation and directives. As a result of this, macro's cannot be commented out.
```
bew: `macro's` -> `macros`

PR #374, opened 2019-11-17
Update macros.md

rdp:
It is confusing that there are several methods not mentioned, so add a link to them.


rdp:
OK I think I found a better place, superceded by #386 thanks.

PR #377, opened 2019-12-17
New Shipping Apps section

ftarulla:
with _using docker_ and _using snapcraft_ subsections


guides/shipping.md
```diff
@@ -0,0 +1,126 @@
...
+```shell-session
+$ crystal ./src/staticserver.cr -- -p 8080 -f .
+Listening on http://0.0.0.0:8080 and serving files in path .
+```
+
+**Note:** the default behavior is to listen on `port 80` and serve the folder `/www`.
+
+## Compiling our application
+
+Let’s go over Crystal’s [introduction](https://crystal-lang.org/reference/). One of the  main goals of the language is to _Compile to efficient native code_. That means that each time that we compile our code then an executable is built, but with an important property: it has a target platform (architecture and operating system), which is where the application will run. Crystal knows the target platform because is the same as the one being used to compile.
```
straight-shoota:
```suggestion
Let’s go over Crystal’s [introduction](https://crystal-lang.org/reference/). One of the  main goals of the language is to _Compile to efficient native code_. That means that each time that we compile our code then an executable is built, but with an important property: it has a target platform (architecture and operating system), which is where the application will run. Crystal chooses the target platform by default as the same as the one being used to compile.
```

guides/shipping.md
```diff
@@ -0,0 +1,126 @@
...
+For example, if we use a Linux OS based computer for compiling, then the executable will be meant to run on a Linux OS (and in some cases we will need to use the same Linux distribution).
```
straight-shoota: `(and in some cases we will need to use the same Linux distribution)` - I'm not aware of any distribution-specific limitations. Do you have an example for that?

guides/shipping.md
```diff
@@ -0,0 +1,126 @@
...
+
+Can we set the target when calling the compiler? Oh, that’s a great idea, but for now it’s not an option (there are a lot of great buccanears [working on a solution](https://forum.crystal-lang.org/t/cross-compiling-automatically-to-osx/1330/12) and remember that Crystal is open source: so you are welcome aboard!)
```
straight-shoota: This section suggest you can't cross compile Crystal. That is very much possible, though.

Only linking the final executable is not covered for a cross-compile target.

I'd either rephrase this paragraph or maybe even remove it entirely?

guides/shipping.md
```diff
@@ -0,0 +1,126 @@
...
+
+Let’s compile our application:
+
+```shell-session
+$ shards build --production
```
straight-shoota: This is not about `shards`, so it rather only use `crystal build`.

guides/shipping.md
bcardiff: I agree to rephrase it. Cross compiling can be mention as something possible as long as you are aware of the difference between compiling, linking. But I would not aim for an explanation. At most that currently you would need a target environment to perform the linking.

guides/shipping.md
bcardiff: I like that shards create a nicer directory structure.

But either we show the shards.yml and the structure of the whole folder.
Or we use `crystal build` (and update the samples below removing the `bin`).

But since we are simulating a whole app, using shards makes sense to me. Otherwise, it feels we are not using our own tools.

guides/shipping.md
straight-shoota: Sure, but that's not restricted by the distro, but the specific environment setup. You should be able to run a Linux executable on any Linux system (i.e. not needing to use the same distro), assuming the environment is configured appropriately.

straight-shoota:
> Does this help to frame better the intention of this guide?

I think I already understood the intention before, but I still think we should not have a shipping guide that's only about two arbitrary shipping methods.

> But Docker and Snapcraft although more have than a plain binary have enough momentum to be included.

I'm not doubting that they can't be included, I'm just saying they shouldn't be the first and only options to be mentioned. They're still useful guides, that's why I suggested to publish them on the blog for now.

> There will always be a chance for things to improve and get better, but we want to offer at least one entry guide for how to complete the whole development cycle and don't send the users in the wild for the seek of answers.

That's good. But when we start a section about shipping apps, it should not consist of only two arbitrary options for a similar mechanism.

bcardiff:
The description in the top-level aims to describe the needs. Even if we add some other distribution method it will still be arbitrary N methods to me, and there needs to be a start.

> But when we start a section about shipping apps, it should not consist of only two arbitrary options for a similar mechanism.

What other content you think is a minimum requirement to add a shipping section?

straight-shoota:
The main issue I have is that these particular examples suggest that Crystal applications can only be shipped containerized. And that's far from the truth. 
Additionally, it would be great to have at least one example for shipping for macOS.

So the minimum required alternatives are basic, non-containerized shipping methods like shipping a single binary, covering both dynamically linked and statically linked examples. For many applications, including the example static file server, shipping a statically linked binary should work pretty well without much hassle. You don't need to reach for complex containerization tools for that.

bcardiff:
I am happy to include a couple of paragraphs in the main file to mitigate the confusion you suggest and explain there are lots of other possibilities.

My main experience is using the approaches included in this PR: Docker & snapcraft.

But further complete guides can be written added once these are in place.

Talking about specific additional guides:

* For osx we could add a guide to make a brew tap formula. But it seems that it can be added later.
Having other alternative to build .pkg would be great, but the only alternative I am aware of is using omnibus and is too convoluted for a guide. (Or maybe is my lack of deep understanding)

* For static building using Alpine or other Linux, I would feel comfortable with others making the first attempt. Since it is not something I regularly use. There were some ideas to have some base Dockerfile/template that will facilitate this but is something that has not happened yet, at least within the organization.

The efforts are to help the user discover the current possibilities. Not to develop something and write a short guide about it.

RX14:
Just produce a binary, talk about static vs dynamic linking and the problems you have with shipping the binary, mention musl and alpine as an option but it's not ready yet, then go ingo snap/docker.

PR #378, opened 2020-01-02
New Continuous Integration section

ftarulla:
Using TravisCI
Using CircleCI (coming soon)

### DIFF
--- a/guides/ci/travis.md
+++ b/guides/ci/travis.md
@@ -45,7 +45,7 @@ script:
   - docker run -v $PWD:/src -w /src crystallang/crystal:0.31.1 crystal spec
 ```
 
-**Note:** We may find the list with the different official [Crystal docker images](https://hub.docker.com/r/crystallang/crystal/tags) at [DockerHub](https://hub.docker.com/r/crystallang/crystal).
+**Note:** A list with the different official [Crystal docker images](https://hub.docker.com/r/crystallang/crystal/tags) is available at [DockerHub](https://hub.docker.com/r/crystallang/crystal).
 
 **Note:** This is a great trick that will let us use different Crystal versions without the need of actually installing the compiler and without the need of creating a `Dockerfile` configuration file.
 
--- a/guides/ci/travis.md
+++ b/guides/ci/travis.md
@@ -354,4 +354,4 @@ script:
   - crystal spec
 ```
 
-Let's push these changes. Travis CI will run, and it will install dependencies, but then it will cache the `lib/` and `$HOME/.cache/shards` folders. The following runs will use the cached dependencies.
+Let's push these changes. Travis CI will run, and it will install dependencies, but then it will cache the `$HOME/.cache/shards` folder. The following runs will use the cached dependencies.
--- a/guides/ci/travis.md
+++ b/guides/ci/travis.md
@@ -136,9 +136,11 @@ script:
 
 Again, these are only examples of two ways of running our application and installing (and using) dependencies.
 
+> *Note:* Dockerfile arguments can be used to use the same Dockerfile for latest, nightly or a specific version.
+
 ## Using a database (example using MySQL)
 
-As you may notice, in the last configuration file we are using `docker` as a `service`. Travis CI may start different services as needed. For example, we may need  [MySQL](https://docs.travis-ci.com/user/database-setup/#mysql)
+As you may notice, in the last configuration file we are using `docker` as a `service`. Travis CI may start different services as needed. For example, we may need [MySQL](https://docs.travis-ci.com/user/database-setup/#mysql).
 
 ```yaml
 # .travis.yml
@@ -205,11 +207,7 @@ class World
     World.new([] of Location)
   end
 
-  def initialize
-    initialize([] of Location)
-  end
-
-  def initialize(living_cells)
+  def initialize(living_cells = [] of Location)
     @living_cells = living_cells
   end
 
@@ -222,7 +220,7 @@ class World
   end
 
   def save
-    DB.connect ENV["DATABASE_URL"], do |cnn|
+    DB.connect ENV["DATABASE_URL"] do |cnn|
       cnn.exec("insert into worlds (living_cells) values (?);", @living_cells.size)
     end
   end
@@ -331,7 +329,7 @@ Fetching https://github.com/crystal-lang/crystal-db.git
 
 This takes time and on the other hand these libraries might not change as often as our application, so it looks like we may cache them and save time.
 
-Travis CI [uses caching](https://docs.travis-ci.com/user/caching/) to improve some parts of the building path. For our example. here is the new configuration file **with cache enabled**:
+Travis CI [uses caching](https://docs.travis-ci.com/user/caching/) to improve some parts of the building path. For our example. Here is the new configuration file **with cache enabled**:
 
 ```yml
 # .travis.yml
--- a/guides/ci/travis.md
+++ b/guides/ci/travis.md
@@ -255,12 +255,10 @@ language: crystal
 crystal:
   - latest
 
-cache:
-  directories:
-    - $HOME/.cache/shards
+cache: shards
 
 script:
   - crystal spec
 ```
 
-Let's push these changes. Travis CI will run, and it will install dependencies, but then it will cache the `$HOME/.cache/shards` folder. The following runs will use the cached dependencies.
+Let's push these changes. Travis CI will run, and it will install dependencies, but then it will cache the shards cache folder which, usually, is `~/.cache/shards`. The following runs will use the cached dependencies.
--- a/guides/ci/travis.md
+++ b/guides/ci/travis.md
@@ -70,16 +70,17 @@ matrix:
        - docker run -v $PWD:/src -w /src crystallang/crystal:0.31.1 crystal spec
 ```
 
-## Using binary dependencies
+## Installing shards packages
 
-Before presenting some examples, it is important to mention that there are many ways to achieve this and **it will heavily depend on our development workflow**.
+Thanks to the `language: crystal` support, Travis CI automatically runs `shards install`. And for faster ci times we may use [caching](#caching)
 
-With this in mind, let's continue!
+## Installing binary dependencies
 
-Using shards dependencies (i.e. libraries declared in `shard.yml`) will not be a problem because, thanks to the `language: crystal` support, Travis CI automatically runs `shards install`.
+> It is important to mention that there are many ways to achieve this and **it will heavily depend on our development workflow**.
 
-But, how do we install dependencies outside `shards.yml`?
-Well, here is a first example installing SQLite3 using Travis CI configuration file:
+Our application or maybe some shards may required libraries and packages. In that case we will need to install them using `apt` (or similar).
+
+Here is a first example installing SQLite3 using Travis CI configuration file:
 
 ```yaml
 # .travis.yml
@@ -88,18 +89,15 @@ crystal:
   - latest
 
 before_install:
-  - sudo apt-get -y install sqlite3
+  - sudo apt-get -y install libsqlite3-dev
 
 addons:
   apt:
     update: true
 
 script:
-  - sqlite3 --version
   - crystal spec
 ```
-
-**Note:** We are using `sqlite3` just for printing SQLite's version, as an example of external dependency.
 
 The same can be accomplished using a containerization service like Docker. Here is an example of a `Dockerfile` installing SQLite3:
 
@@ -107,7 +105,7 @@ The same can be accomplished using a containerization service like Docker. Here 
 # Dockerfile
 FROM crystallang/crystal:latest
 
-RUN apt-get update && apt-get install -y sqlite3
+RUN apt-get update && apt-get install -y libsqlite3-dev
 ```
 
 And here is the Travis CI configuration file:
@@ -126,13 +124,9 @@ before_install:
 script:
   # run specs in the container
   - docker run -v $PWD:/src -w /src testing crystal spec
-  # sqlite3
-  - docker run -v $PWD:/src -w /src testing sqlite3 --version
 ```
 
-Again, these are only examples of two ways of running our application and installing (and using) dependencies.
-
-> *Note:* Dockerfile arguments can be used to use the same Dockerfile for latest, nightly or a specific version.
+**Note:** Dockerfile arguments can be used to use the same Dockerfile for latest, nightly or a specific version.
 
 ## Using a database (example using MySQL)
 
--- a/guides/ci/travis.md
+++ b/guides/ci/travis.md
@@ -151,50 +151,23 @@ script:
   - crystal spec
 ```
 
-To continue this example, let's add a new test that uses the database.
-(**important:** the new test case is only for testing Travis CI database service, and it's not a good design or test case by any means)
-
-So, let's add the new test:
+Here is the new test file for testing against the database:
 
 ```crystal
-# game_of_life_spec.cr
+# spec/simple_db_spec.cr
+require "./spec_helper"
+require "mysql"
 
-[...]
-
-describe "an empty world" do
-
-  [...]
-
-  # this is the new test:
-  it "should be persisted" do
-    World.empty.save
-  end
-end
-```
-
-And the new implementation:
-
-```crystal
-# game_of_life.cr
-require "mysql" # let's require the MySQL driver
-
-[...]
-
-class World
-  [...]
-
-  # and here is the new implementation:
-  def save
-    DB.connect ENV["DATABASE_URL"] do |cnn|
-      cnn.exec("insert into worlds (living_cells) values (?);", @living_cells.size)
-    end
+it "connects to the database" do
+  DB.connect ENV["DATABASE_URL"] do |cnn|
+    cnn.query_one("SELECT 'foo'", as: String).should eq "foo"
   end
 end
 ```
 
 When pushing this changes Travis CI will report the following error: `Unknown database 'test' (Exception)`, showing that we need to configure the MySQL service **and also setup the database**:
 
->It's really **important** to notice that the lines we are adding to `.travis.yml` will depend exclusively on the development workflow we are using!
+> It's really **important** to notice that the lines we are adding to `.travis.yml` will depend exclusively on the development workflow we are using!
 > And remember that this is only an example using MySQL.
 
 ```yaml
@@ -221,7 +194,7 @@ We are [using a `setup.sql` script](https://andidittrich.de/2017/06/travisci-set
 ```sql
 -- setup.sql
 CREATE DATABASE IF NOT EXISTS test;
-CREATE TABLE `test`.`worlds` (`id` serial,`living_cells` int NOT NULL DEFAULT 0, PRIMARY KEY (id));
+-- CREATE TABLE ... etc ...
 ```
 
 Pushing these changes will trigger Travis CI and the build should be successful!!
--- a/guides/ci/travis.md
+++ b/guides/ci/travis.md
@@ -45,7 +45,7 @@ script:
   - docker run -v $PWD:/src -w /src crystallang/crystal:0.31.1 crystal spec
 ```
 
-**Note:** We may find the list with the different official [Crystal docker images](https://hub.docker.com/r/crystallang/crystal/tags) at [DockerHub](https://hub.docker.com/r/crystallang/crystal)
+**Note:** We may find the list with the different official [Crystal docker images](https://hub.docker.com/r/crystallang/crystal/tags) at [DockerHub](https://hub.docker.com/r/crystallang/crystal).
 
 **Note:** This is a great trick that will let us use different Crystal versions without the need of actually installing the compiler and without the need of creating a `Dockerfile` configuration file.
 
@@ -81,7 +81,9 @@ matrix:
 
 Before presenting some examples, it is important to mention that there are many ways to achieve this and **it will heavily depend on our development workflow**.
 
-With this in mind, here is a first example installing Git using Travis CI configuration file:
+With this in mind, let's continue!
+
+Using shards dependencies (i.e. libraries declared in `shard.yml`) will not be a problem, because thanks to the `language: crystal` support, TravisCI automatically runs `shards install`. But, how do we install dependencies outside `shards.yml`? Well, here is a first example installing SQLite3 using Travis CI configuration file:
 
 ```yaml
 # .travis.yml
@@ -90,28 +92,26 @@ crystal:
   - latest
 
 before_install:
-  - sudo apt-get -y install git
+  - sudo apt-get -y install sqlite3
 
 addons:
   apt:
     update: true
 
 script:
-  - git --version
+  - sqlite3 --version
   - crystal spec
 ```
 
-**Note:** We are using `git` just for printing GIT's version, as an example of external dependency.
+**Note:** We are using `sqlite3` just for printing SQLite's version, as an example of external dependency.
 
-**Note:** Git is already installed by TravisCI and available but (again) we are using it as an example of installing and using a dependency.
-
-The same can be accomplished using a containerization service like Docker. Here is an example of a `Dockerfile` installing Git:
+The same can be accomplished using a containerization service like Docker. Here is an example of a `Dockerfile` installing SQLite3:
 
 ```dockerfile
 # Dockerfile
 FROM crystallang/crystal:latest
 
-RUN apt-get update && apt-get install -y git
+RUN apt-get update && apt-get install -y sqlite3
 ```
 
 And here is the TravisCI configuration file:
@@ -126,17 +126,13 @@ services:
 before_install:
   # build image using Dockerfile:
   - docker build -t testing .
-  # Git installed by Travis CI
-  - git --version # => 2.21.0
 
 script:
   # run specs in the container
   - docker run -v $PWD:/src -w /src testing crystal spec
-  # Git installed in Docker container (see Dockerfile)
-  - docker run -v $PWD:/src -w /src testing git --version # => 2.17.1
+  # sqlite3
+  - docker run -v $PWD:/src -w /src testing sqlite3 --version
 ```
-
-We may notice that GIT's version running in the `testing` container is different from Git running by default in Travis CI.
 
 Again, these are only examples of two ways of running our application and installing (and using) dependencies.
 
--- a/guides/ci/travis.md
+++ b/guides/ci/travis.md
@@ -23,7 +23,7 @@ script:
   - crystal tool format --check
 ```
 
-Although a basic configuration file, it will let's use Travis CI. Now, let's go to Travis CI dashboard to [add the GitHub repository](https://docs.travis-ci.com/user/tutorial/#to-get-started-with-travis-ci) From this moment, Travis CI will run the `specs` against the Crystal compiler (using both `latest` and `nightly` releases)
+Although a basic configuration file, it will let's use Travis CI. Now, let's go to Travis CI dashboard to [add the GitHub repository](https://docs.travis-ci.com/user/tutorial/#to-get-started-with-travis-ci). From this moment, Travis CI will run the `specs` against the Crystal compiler (using both `latest` and `nightly` releases)
 
 ### Using a specific Crystal release
 
@@ -42,7 +42,7 @@ services:
   - docker
 
 script:
-  - docker run -v $PWD:/src -w /src crystallang/crystal:0.31.1 crystal spec spec/game_of_life_spec.cr
+  - docker run -v $PWD:/src -w /src crystallang/crystal:0.31.1 crystal spec
 ```
 
 **Note:** We may find the list with the different official [Crystal docker images](https://hub.docker.com/r/crystallang/crystal/tags) at [DockerHub](https://hub.docker.com/r/crystallang/crystal)
@@ -74,7 +74,7 @@ matrix:
      services:
        - docker
      script:
-       - docker run -v $PWD:/src -w /src crystallang/crystal:0.31.1 crystal spec spec/game_of_life_spec.cr
+       - docker run -v $PWD:/src -w /src crystallang/crystal:0.31.1 crystal spec
 ```
 
 ## Using external dependencies
@@ -98,7 +98,7 @@ addons:
 
 script:
   - git --version
-  - crystal spec spec/game_of_life_spec.cr
+  - crystal spec
 ```
 
 **Note:** We are using `git` just for printing GIT's version, as an example of external dependency.
@@ -131,7 +131,7 @@ before_install:
 
 script:
   # run specs in the container
-  - docker run -v $PWD:/src -w /src testing crystal spec spec/game_of_life_spec.cr
+  - docker run -v $PWD:/src -w /src testing crystal spec
   # Git installed in Docker container (see Dockerfile)
   - docker run -v $PWD:/src -w /src testing git --version # => 2.17.1
 ```
--- a/guides/ci/travis.md
+++ b/guides/ci/travis.md
@@ -27,12 +27,7 @@ Although a basic configuration file, it will let us use Travis CI. Now, let's go
 
 ### Using a specific Crystal release
 
-So, let's suppose we want to use version [0.31.1](https://github.com/crystal-lang/crystal/releases/tag/0.31.1)
- of the Crystal compiler. We will be tempting to add `- 0.31.1`. But, as we may read at the [Crystal configuration options](https://docs.travis-ci.com/user/languages/crystal/#configuration-options), we can only use the values `latest` and `nightly`.
-
-So, how do we test against a specific version?
-
-For this we are going to use [Docker](https://www.docker.com/). First we need to add Docker as a service in `.travis.yml`, and then use `docker` from the CLI to run the specs with a specific Crystal version, like this:
+So, let's suppose we want to use version [0.31.1](https://github.com/crystal-lang/crystal/releases/tag/0.31.1) of the Crystal compiler. For this we are going to use [Docker](https://www.docker.com/). First we need to add Docker as a service in `.travis.yml`, and then we can use `docker` commands in our build steps, like this:
 
 ```yml
 # .travis.yml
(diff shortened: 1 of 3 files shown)


guides/ci/travis.md
```diff
@@ -0,0 +1,364 @@
...
+
+If you are new to continuous integration (or you want to refresh the basic concepts) we may start reading the [core concepts guide](https://docs.travis-ci.com/user/for-beginners/) and then a [short tutorial](https://docs.travis-ci.com/user/tutorial/).
+
+Now let’s see an example of using Travis CI with Crystal.
+
+## Build and run specs
+
+### Using `latest` and `nightly`
+
+When [creating a Crystal project](https://crystal-lang.org/reference/using_the_compiler/#creating-a-crystal-project) using `crystal init`, Crystal creates a `.travis.yml` file for us:
```
bcardiff: Can we have relative links for thinks pointing to `/reference`?

guides/ci/travis.md
```diff
@@ -0,0 +1,364 @@
...
+crystal:
+  - latest
+  - nightly
+
+script:
+  - crystal spec
+  - crystal tool format --check
+```
+
+Although a basic configuration file, it will let’s use Travis CI. Now, let’s go to Travis CI dashboard to [add the GitHub repository](https://docs.travis-ci.com/user/tutorial/#to-get-started-with-travis-ci) From this moment, Travis CI will run the `specs` against the Crystal compiler (using both `latest` and `nightly` releases)
```
bcardiff:
```suggestion
Although a basic configuration file, it will let’s use Travis CI. Now, let’s go to Travis CI dashboard to [add the GitHub repository](https://docs.travis-ci.com/user/tutorial/#to-get-started-with-travis-ci). From this moment, Travis CI will run the `specs` against the Crystal compiler (using both `latest` and `nightly` releases)
```

guides/ci/travis.md
```diff
@@ -0,0 +1,364 @@
...
+     crystal:
+       - nightly
+     script:
+       - crystal spec
+
+   - language: bash
+     services:
+       - docker
+     script:
+       - docker run -v $PWD:/src -w /src crystallang/crystal:0.31.1 crystal spec spec/game_of_life_spec.cr
```
bcardiff:
```suggestion
       - docker run -v $PWD:/src -w /src crystallang/crystal:0.31.1 crystal spec
```

there is no need to use a different command `crystal spec` should work.

guides/ci/travis.md
```diff
@@ -0,0 +1,364 @@
...
+language: bash
+
+services:
+  - docker
+
+script:
+  - docker run -v $PWD:/src -w /src crystallang/crystal:0.31.1 crystal spec spec/game_of_life_spec.cr
+```
+
+**Note:** We may find the list with the different official [Crystal docker images](https://hub.docker.com/r/crystallang/crystal/tags) at [DockerHub](https://hub.docker.com/r/crystallang/crystal)
```
bcardiff:
```suggestion
**Note:** We may find the list with the different official [Crystal docker images](https://hub.docker.com/r/crystallang/crystal/tags) at [DockerHub](https://hub.docker.com/r/crystallang/crystal).
```

guides/ci/travis.md
```diff
@@ -0,0 +1,364 @@
...
+   - language: bash
+     services:
+       - docker
+     script:
+       - docker run -v $PWD:/src -w /src crystallang/crystal:0.31.1 crystal spec spec/game_of_life_spec.cr
+```
+
+## Using external dependencies
+
+Before presenting some examples, it is important to mention that there are many ways to achieve this and **it will heavily depend on our development workflow**.
```
bcardiff: I think some intro is missing in this part. The goal is to install native libraries that might be linked by the app. `sqlite3` could be a better example than `git`. ( `sqlite3 --version` works ;-) )

And maybe a clarification, or short previous section, to mention that shards dependencies are installed directly in the `install` step thanks to the `language: crystal` support. Otherwise talking about dependencies could be misleading.

guides/ci/travis.md
```diff
@@ -0,0 +1,364 @@
...
+  - git --version # => 2.21.0
+
+script:
+  # run specs in the container
+  - docker run -v $PWD:/src -w /src testing crystal spec spec/game_of_life_spec.cr
+  # Git installed in Docker container (see Dockerfile)
+  - docker run -v $PWD:/src -w /src testing git --version # => 2.17.1
+```
+
+We may notice that GIT’s version running in the `testing` container is different from Git running by default in Travis CI.
```
bcardiff: I would focus on the fact that the external dependency needs to be in the docker images.

guides/ci/travis.md
```diff
@@ -0,0 +1,364 @@
...
+
+Again, these are only examples of two ways of running our application and installing (and using) dependencies.
```
bcardiff: I would add the following note.

> Note: Dockerfile arguments can be used to use the same Dockerfile for latest, nightly or a specific version.

There is no need for a full example.

guides/ci/travis.md
```diff
@@ -0,0 +1,364 @@
...
+
+## Using a database (example using MySQL)
+
+As you may notice, in the last configuration file we are using `docker` as a `service`. Travis CI may start different services as needed. For example, we may need  [MySQL](https://docs.travis-ci.com/user/database-setup/#mysql)
```
bcardiff:
```suggestion
As you may notice, in the last configuration file we are using `docker` as a `service`. Travis CI may start different services as needed. For example, we may need [MySQL](https://docs.travis-ci.com/user/database-setup/#mysql).
```

guides/ci/travis.md
```diff
@@ -0,0 +1,364 @@
...
+
+script:
+  - crystal spec
+```
+
+To continue this example, let’s add a new test that uses the database.
+(**important:** the new test case is only for testing Travis CI database service, and it’s not a good design or test case by any means)
+
+```crystal
+# game_of_life_spec.cr
```
bcardiff:
```suggestion
# spec/game_of_life_spec.cr
```

guides/ci/travis.md
```diff
@@ -0,0 +1,364 @@
...
+  it "should be persisted" do
+    World.empty.save
+  end
+end
+```
+
+And the new implementation:
+
+```crystal
+# game_of_life.cr
```
bcardiff:
```suggestion
# src/game_of_life.cr
```

guides/ci/travis.md
```diff
@@ -0,0 +1,364 @@
...
+
+  def self.empty
+    World.new([] of Location)
+  end
+
+  def initialize
+    initialize([] of Location)
+  end
+
+  def initialize(living_cells)
```
bcardiff:
```suggestion
  def initialize(living_cells = [] of Location)
```

Use default args and the empty initialize can be removed.

guides/ci/travis.md
```diff
@@ -0,0 +1,364 @@
...
+  def set_living_at(a_location)
+    @living_cells << a_location
+  end
+
+  def is_empty?
+    @living_cells.size == 0
+  end
+
+  def save
+    DB.connect ENV["DATABASE_URL"], do |cnn|
```
bcardiff:
```suggestion
    DB.connect ENV["DATABASE_URL"] do |cnn|
```

guides/ci/travis.md
```diff
@@ -0,0 +1,364 @@
...
+If we read Travis CI job log, we will find that every time the job runs, Travis CI needs to fetch the libraries needed to run the application:
+
+```log
+Fetching https://github.com/crystal-lang/crystal-mysql.git
+Fetching https://github.com/crystal-lang/crystal-db.git
+```
+
+This takes time and on the other hand these libraries might not change as often as our application, so it looks like we may cache them and save time.
+
+Travis CI [uses caching](https://docs.travis-ci.com/user/caching/) to improve some parts of the building path. For our example. here is the new configuration file **with cache enabled**:
```
bcardiff:
```suggestion
Travis CI [uses caching](https://docs.travis-ci.com/user/caching/) to improve some parts of the building path. For our example. Here is the new configuration file **with cache enabled**:
```

bcardiff: Someday this could be integrated directly in a `setup_cache` method at https://github.com/travis-ci/travis-build/blob/master/lib/travis/build/script/crystal.rb . But for now, this section is needed.

guides/continuous_integration.md
```diff
@@ -0,0 +1,96 @@
...
+
+In the next subsections, we are going to present 2 continuous integration tools: [Travis CI](https://travis-ci.org/) and [Circle CI](https://circleci.com/) and use them with a Crystal example application.
+
+These tools not only will let us build and test our code each time the source has changed but also deploy the result (if the build was successful) or use automatic builds, and maybe test against different platforms, to mention a few.
+
+## The example application
+
+We are going to use Conway’s Game of Life as the example application. More precisely, we are going to use only the first iterations in [Conway's Game of Life Kata](http://codingdojo.org/kata/GameOfLife/) solution using [TDD](https://martinfowler.com/bliki/TestDrivenDevelopment.html).
```
bcardiff: The guide itself does not use TDD but it mimics as if starting to.

guides/continuous_integration.md
```diff
@@ -0,0 +1,96 @@
...
+
+  def self.empty
+    World.new([] of Location)
+  end
+
+  def initialize
+    initialize([] of Location)
+  end
+
+  def initialize(living_cells)
```
bcardiff:
```suggestion
  def initialize(living_cells = [] of Location)
```

Use default args and the empty initialize can be removed.

guides/continuous_integration.md
```diff
@@ -0,0 +1,96 @@
...
+
+  def initialize(@x, @y)
+  end
+end
+
+class World
+  @living_cells : Array(Location)
+
+  def self.empty
+    World.new([] of Location)
```
bcardiff:
```suggestion
   new
```

guides/continuous_integration.md
```diff
@@ -0,0 +1,96 @@
...
+These tools not only will let us build and test our code each time the source has changed but also deploy the result (if the build was successful) or use automatic builds, and maybe test against different platforms, to mention a few.
+
+## The example application
+
+We are going to use Conway’s Game of Life as the example application. More precisely, we are going to use only the first iterations in [Conway's Game of Life Kata](http://codingdojo.org/kata/GameOfLife/) solution using [TDD](https://martinfowler.com/bliki/TestDrivenDevelopment.html).
+
+Here’s the implementation:
+
+```crystal
+# game_of_life.cr
```
bcardiff:
```suggestion
# src/game_of_life.cr
```

guides/continuous_integration.md
```diff
@@ -0,0 +1,96 @@
...
+  def is_empty?
+    @living_cells.size == 0
+  end
+end
+```
+
+And the specs:
+
+```crystal
+# game_of_life_spec.cr
```
bcardiff:
```suggestion
# spec/game_of_life_spec.cr
```

guides/continuous_integration.md
```diff
@@ -0,0 +1,96 @@
...
+
+In the next subsections, we are going to present 2 continuous integration tools: [Travis CI](https://travis-ci.org/) and [Circle CI](https://circleci.com/) and use them with a Crystal example application.
+
+These tools not only will let us build and test our code each time the source has changed but also deploy the result (if the build was successful) or use automatic builds, and maybe test against different platforms, to mention a few.
+
+## The example application
+
+We are going to use Conway’s Game of Life as the example application. More precisely, we are going to use only the first iterations in [Conway's Game of Life Kata](http://codingdojo.org/kata/GameOfLife/) solution using [TDD](https://martinfowler.com/bliki/TestDrivenDevelopment.html).
+
+Here’s the implementation:
```
bcardiff: There should be a mention that it was expanded with `crystal init` otherwise `spec_helper.cr` would not exist and the `crystal spec` commands would need explicit files.

guides/ci/travis.md
```diff
@@ -0,0 +1,358 @@
...
+language: crystal
+crystal:
+  - latest
+
+env:
+  - DATABASE_URL="mysql://root@localhost/test"
+
+cache:
+  directories:
+    - lib
```
bcardiff:
```suggestion
```

guides/ci/travis.md
```diff
@@ -0,0 +1,358 @@
...
+  - mysql
+
+before_install:
+  - mysql -u root --password="" < travisci/test.sql
+
+script:
+  - crystal spec
+```
+
+Let's push these changes. Travis CI will run, and it will install dependencies, but then it will cache the `lib/` and `$HOME/.cache/shards` folders. The following runs will use the cached dependencies.
```
bcardiff:
```suggestion
Let's push these changes. Travis CI will run, and it will install dependencies, but then it will cache the `$HOME/.cache/shards` folder. The following runs will use the cached dependencies.
```

guides/ci/travis.md
```diff
@@ -0,0 +1,358 @@
...
+       - docker run -v $PWD:/src -w /src crystallang/crystal:0.31.1 crystal spec
+```
+
+## Using external dependencies
+
+Before presenting some examples, it is important to mention that there are many ways to achieve this and **it will heavily depend on our development workflow**.
+
+With this in mind, let's continue!
+
+Using shards dependencies (i.e. libraries declared in `shard.yml`) will not be a problem, because thanks to the `language: crystal` support, TravisCI automatically runs `shards install`. But, how do we install dependencies outside `shards.yml`? Well, here is a first example installing SQLite3 using Travis CI configuration file:
```
straight-shoota: The fact about shards being installed should definitely be mentioned earlier.
How does this work out with Docker, though? It wouldn't test installing shards with the shards version provided in the dockerfile, but the one in the travis runner.

guides/ci/travis.md
```diff
@@ -0,0 +1,358 @@
...
+       - crystal spec
+
+   - language: bash
+     services:
+       - docker
+     script:
+       - docker run -v $PWD:/src -w /src crystallang/crystal:0.31.1 crystal spec
+```
+
+## Using external dependencies
```
straight-shoota: I would clarify this section. `shards` are also external dependencies, but not actually covered here. I suppose it's about binary dependencies?

guides/ci/travis.md
```diff
@@ -0,0 +1,358 @@
...
+    update: true
+
+script:
+  - sqlite3 --version
+  - crystal spec
+```
+
+**Note:** We are using `sqlite3` just for printing SQLite's version, as an example of external dependency.
+
+The same can be accomplished using a containerization service like Docker. Here is an example of a `Dockerfile` installing SQLite3:
```
straight-shoota: This is why I think running CI in Docker images should be separated topic. Also because it's not actually specific to travis (only the syntax for configuring it initially).

guides/ci/travis.md
```diff
@@ -0,0 +1,358 @@
...
+
+script:
+  - crystal spec
+```
+
+To continue this example, let's add a new test that uses the database.
+(**important:** the new test case is only for testing Travis CI database service, and it's not a good design or test case by any means)
+
+```crystal
+# game_of_life_spec.cr
```
straight-shoota: This example is way to complex for a simple demonstration.
Just putting `DB.connect` in a spec file is completely sufficient.

guides/ci/travis.md
```diff
@@ -0,0 +1,358 @@
...
+
+  def save
+    DB.connect ENV["DATABASE_URL"] do |cnn|
+      cnn.exec("insert into worlds (living_cells) values (?);", @living_cells.size)
+    end
+  end
+end
+```
+
+We may use this `docker-compose` configuration to have a local MySQL database for our test to run:
```
straight-shoota: This adds even more complexity and it doesn't even explain why you would need docker compose when you can install a mysql service in travis.

guides/ci/travis.md
bcardiff: Renaming it a binary dependencies sounds right.

bcardiff:
The goal is to allow a crystal user to pick the CI infrastructure and dig into how to add ci for latest, nightly and specific Crystal version. How to install dependencies. How to use external services. And if possible, how to improve ci time by caching shards. This is mentioned at the end of the main entry for CI included in the PR.

In the case of Travis this means mentioning docker currently. There are other ways to do it, of course, but I don't want to force the users to discover them by themself.

guides/ci/travis.md
```diff
@@ -0,0 +1,358 @@
...
+script:
+  - crystal spec
+  - crystal tool format --check
+```
+
+Although a basic configuration file, it will let's use Travis CI. Now, let's go to Travis CI dashboard to [add the GitHub repository](https://docs.travis-ci.com/user/tutorial/#to-get-started-with-travis-ci). From this moment, Travis CI will run the `specs` against the Crystal compiler (using both `latest` and `nightly` releases)
+
+### Using a specific Crystal release
+
+So, let's suppose we want to use version [0.31.1](https://github.com/crystal-lang/crystal/releases/tag/0.31.1)
```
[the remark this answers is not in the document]
bcardiff: Sometimes an app is unable to follow the upgrade of the underlying framework/language. It's better to have a way of not blocking the development of it.

guides/ci/travis.md
```diff
@@ -0,0 +1,358 @@
...
+  - docker run -v $PWD:/src -w /src crystallang/crystal:0.31.1 crystal spec
+```
+
+**Note:** We may find the list with the different official [Crystal docker images](https://hub.docker.com/r/crystallang/crystal/tags) at [DockerHub](https://hub.docker.com/r/crystallang/crystal).
+
+**Note:** This is a great trick that will let us use different Crystal versions without the need of actually installing the compiler and without the need of creating a `Dockerfile` configuration file.
+
+### Using `latest`, `nightly` and a specific Crystal release all together!
+
+For using 3 versions of the compiler we are going to use a different configuration file based on [Build Matrix](https://docs.travis-ci.com/user/customizing-the-build#build-matrix)
```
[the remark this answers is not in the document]
bcardiff: I see it as progressive support in Travis. It's better to use the native approach when possible.

straight-shoota:
I don't agree with the premise that this guide should show how to run CI against a specific Crystal release, when that makes essentially means duplicating every following step in the guide and confusing both variants in the process.
It's better to make two guides, dedicated to the specifics of each method and everyone can pick whatever they need.

guides/ci/travis.md
```diff
@@ -0,0 +1,358 @@
...
+services:
+  - docker
+
+script:
+  - docker run -v $PWD:/src -w /src crystallang/crystal:0.31.1 crystal spec
+```
+
+**Note:** We may find the list with the different official [Crystal docker images](https://hub.docker.com/r/crystallang/crystal/tags) at [DockerHub](https://hub.docker.com/r/crystallang/crystal).
+
+**Note:** This is a great trick that will let us use different Crystal versions without the need of actually installing the compiler and without the need of creating a `Dockerfile` configuration file.
```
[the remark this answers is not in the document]
bcardiff: Ok, we can remove the note

guides/ci/travis.md
bcardiff: The locally running examples will go away and reduce the length.

bcardiff:
There is no initial reason why someone will pick (a) Travis vs (b) Docker on Travis or (c) CI with docker.
The requirement is to use X provider and from there support scenarios (crystal version, dependencies, services). I intent the docs to be structured in that way to simplify things for the reader, even at the cost of duplication.

But for example on CircleCI there will be no distinction between them as it is in Travis. So there will be no duplication there.

If on the future the crystal version is available on travis natively then this can be redone. But today this is the way I prefer to advertise how to.

guides/ci/travis.md
ftarulla: Yep, it is a little more complex example than `DB.connect` but it has another purpose: it shows how to use `environment variables` (in this case `DATABASE_URL`)
For making the example shorter, now we only show the lines that need to be added to the example introduced in Continuous Integration.

straight-shoota: Really, I would just make the example be as tight as possible.
```cr
require "mysql"
require "spec"

it "connects do database" do
  DB.connect(ENV["DATABASE_URL"]) do |db|
    db.query("SELECT 'foo'", as: String). should eq "foo"
  end
end
```

There's no point in having anything else here. It's just noise.

straight-shoota:
> I intent the docs to be structured in that way to simplify things for the reader, even at the cost of duplication.

Sure, duplication is not an issue. But jumping between examples for plain travis runners and docker on travis doesn't seem simplified for the reader.

guides/ci/travis.md
ftarulla: oh! ok, that's a simpler example, but I forgot to mention that the original example has another goal: to show how to use a `setup.sql` file to initialize the database from TravisCI. This may seem trivial but it's not (that's why I even mention the source of the solution: https://andidittrich.de/2017/06/travisci-setup-mysql-tablesdata-before-running-tests.html)
So we wanted to create an example that the reader could almost cut&paste.

I think that, after merging this new section, we may iterate with a new and more simple example (oh! and we would update the example in the introduction as well!) and then create a new PR with those changes. 

@straight-shoota, @bcardiff:  Does it seem ok?

By the way: thank you for all the suggestions!!

straight-shoota: I seriously doubt that *setting up* a MySQL database fits in the scope of this guide. The goal of this section should be to demonstrate the concept of using services on travis.
How to actually use such a service is a highly specific topic. MySQL is really just a stand in for any travis service. You wouldn't want to dive into giving instructions for other services either.
It should be sufficient to refer to https://docs.travis-ci.com/user/database-setup/ for more specific documentation.

ftarulla: ok! I removed the current example and added the simpler DB example.

Let me know what you think!

guides/ci/travis.md
```diff
@@ -0,0 +1,238 @@
...
+       - crystal spec
+
+   - language: bash
+     services:
+       - docker
+     script:
+       - docker run -v $PWD:/src -w /src crystallang/crystal:0.31.1 crystal spec
+```
+
+## Using binary dependencies
```
bcardiff: Maybe we need to have a `## Installing shards dependencies` section where we said that 

Thanks to the `language: crystal` support, Travis CI automatically runs `shards install`. And reference `Caching` for faster ci times.

Then the binary dependencies section can focus only on that. Some shards require libraries and packages. You will need to install them using `apt` or similar.

guides/ci/travis.md
```diff
@@ -0,0 +1,238 @@
...
+Well, here is a first example installing SQLite3 using Travis CI configuration file:
+
+```yaml
+# .travis.yml
+language: crystal
+crystal:
+  - latest
+
+before_install:
+  - sudo apt-get -y install sqlite3
```
bcardiff:
```suggestion
  - sudo apt-get -y install libsqlite3-dev
```

this is the package that will normally be needed.

guides/ci/travis.md
```diff
@@ -0,0 +1,238 @@
...
+
+addons:
+  apt:
+    update: true
+
+script:
+  - sqlite3 --version
```
bcardiff:
```suggestion
  - apt show libsqlite3-dev
```

and this is how we can check the package installed 🤷‍♂

guides/ci/travis.md
```diff
@@ -0,0 +1,238 @@
...
+  - crystal spec
+```
+
+**Note:** We are using `sqlite3` just for printing SQLite's version, as an example of external dependency.
```
bcardiff:
```suggestion
**Note:** We are using `apt show` just for printing SQLite's version, as an example of binary dependency.
```

guides/ci/travis.md
```diff
@@ -0,0 +1,238 @@
...
+
+The same can be accomplished using a containerization service like Docker. Here is an example of a `Dockerfile` installing SQLite3:
+
+```dockerfile
+# Dockerfile
+FROM crystallang/crystal:latest
+
+RUN apt-get update && apt-get install -y sqlite3
```
bcardiff:
```suggestion
RUN apt-get update && apt-get install -y libsqlite3-dev
```

guides/ci/travis.md
```diff
@@ -0,0 +1,238 @@
...
+
+before_install:
+  # build image using Dockerfile:
+  - docker build -t testing .
+
+script:
+  # run specs in the container
+  - docker run -v $PWD:/src -w /src testing crystal spec
+  # sqlite3
+  - docker run -v $PWD:/src -w /src testing sqlite3 --version
```
bcardiff:
```suggestion
  - docker run -v $PWD:/src -w /src testing apt show libsqlite3-dev
```

guides/continuous_integration.md
```diff
@@ -0,0 +1,96 @@
...
+
+## Continuous Integration step by step
+
+Here's the list of items we want to achieve:
+
+1. Build and run specs using 3 different Crystal's versions:
+   * latest
+   * nightly
+   * 0.31.1 (using a Docker image)
+2. Use external dependencies
```
bcardiff:
```suggestion
3. Use binary dependencies
```

And if the shards deserves it's own section as proposed above

```
2. Install shards packages
```

guides/ci/travis.md
straight-shoota: I think the idea was that `sqlite3` is installed as an executable. Of course we don't necessarily need that., using `libsqlite` is more precise.
But the wording needs to be changed: using `apt show` to display the installed package is not exactly an example of a binary dependency.

ftarulla: Changing "using" with "installing"

guides/ci/travis.md
```diff
@@ -0,0 +1,232 @@
...
+     services:
+       - docker
+     script:
+       - docker run -v $PWD:/src -w /src crystallang/crystal:0.31.1 crystal spec
+```
+
+## Installing shards packages
+
+Thanks to the `language: crystal` support, Travis CI automatically runs `shards install`. And for faster ci times we may use [caching](#caching)
+
```
bcardiff: We should add:

When using a docker image you will need to run `shards install` explicitly. 

```
script:
  - docker run -v $PWD:/src -w /src crystallang/crystal:0.31.1 shards install
  - docker run -v $PWD:/src -w /src crystallang/crystal:0.31.1 crystal spec
```

Note: Since the shards will be installed in `./lib/` folder, it will be preserved for the second `docker run` command.

straight-shoota:
Sorry for this interruption, but why don't we update the guide at https://docs.travis-ci.com/user/languages/crystal/ with this content?
IMO this location has a few advantages. It's probably easier to discover for travis users trying to integrate Crystal projects and it's embedded into the travis documentation, making it easier to find related information.
On the downside we'd lose control of the content (because it's managed by travis). I don't expect that to be an issue, though. We're also sending readers away from our own documentation platform. That's not necessarily bad, but needs to be considered. In the end, it's even beneficial to have Crystal resources integrated with other services and discoverable from different angles.

Duplicating content could be an option, but doesn't sound like a great idea.

Most of the content of this guide I would also expect to find in travis' documentation for Crystal lang. Binary dependencies and services could probably be skipped there because it's not specific to using Crystal.

So maybe it's still better to put the Crystal-specific basics on travis doc and accept some kind of duplication. But our guide could be a more expressive walk through which combines separate steps from the travis documentation.

bcardiff:
I will try to update that guide later once they merge a PR I sent for some features.

I'm happy with some duplication of content. And I prefer that we can have the authorship of how to present things.

Note: Maybe after 1.0 the Docker story in travis will not be needed and there will be less friction. But now this is the best we have.

RX14:
I'd like for travis to support old versions ideally.

bcardiff:
That may happen after 1.0 thanks to snapcraft tracks. Yet that will not allow to pin to an exact patch version. Only the latest Major.minor.

RX14:
Why overcomplicate this with snapcraft when you can just use the old `.deb` files, which we already have on all the github releases. Depending on how the APT repository is set up, you might already have all the old package versions available.

bcardiff:
Because travis goal is to simplify their language support through snapcraft.

straight-shoota:
Okay, let's please not stray off course. We can currently only document using docker for pinned versions. If and how to change that doesn't belong here.

guides/ci/travis.md
```diff
@@ -0,0 +1,252 @@
...
+  apt:
+    update: true
+
+script:
+  - crystal spec
+```
+
+#### Using Docker
+
+The same can be accomplished using a containerization service like Docker. Here is an example of a `Dockerfile` installing SQLite3:
```
straight-shoota: Using a `Dockerfile` is a very different method and introduces a completely new concept without much context. An example more similar to the previous section using `docker run` to install a dependency in the container is probably not a good idea, because it is more complex. But maybe that should be mentioned here? And a few more words about the method `Dockerfile`? When using a `Dockerfile` anyway, it would also make sense to install shards there.

However, a `Dockerfile` for setting up dependencies is not specific to travis CI, so maybe there should rather be a separate guide for more details which would be referenced here.

guides/ci/travis.md
```diff
@@ -0,0 +1,252 @@
...
+  - docker build -t testing .
+
+script:
+  # run specs in the container
+  - docker run -v $PWD:/src -w /src testing crystal spec
+```
+
+**Note:** Dockerfile arguments can be used to use the same Dockerfile for latest, nightly or a specific version.
+
+## Using a database (example using MySQL)
```
straight-shoota: MySQL database is just an example, I'd call the section more general:
```suggestion
## Using services
```

guides/ci/travis.md
```diff
@@ -0,0 +1,252 @@
...
+
+As you may notice, in the last configuration file we are using `docker` as a `service`. Travis CI may start different services as needed. For example, we may need [MySQL](https://docs.travis-ci.com/user/database-setup/#mysql).
```
straight-shoota: I'm not sure the reference to the previous section is helpful. Readers who are only interested in travis runners have probably skipped it anyways. Maybe we could just swap dependencies and services section? This would avoid applying a concept before its introduction. Services like databases are probably more often used.

guides/ci/travis.md
```diff
@@ -0,0 +1,252 @@
...
+it "connects to the database" do
+  DB.connect ENV["DATABASE_URL"] do |cnn|
+    cnn.query_one("SELECT 'foo'", as: String).should eq "foo"
+  end
+end
+```
+
+When pushing this changes Travis CI will report the following error: `Unknown database 'test' (Exception)`, showing that we need to configure the MySQL service **and also setup the database**:
+
+> It's really **important** to notice that the lines we are adding to `.travis.yml` will depend exclusively on the development workflow we are using!
```
straight-shoota: I don't understand this.

guides/ci/travis.md
```diff
@@ -0,0 +1,252 @@
...
+> And remember that this is only an example using MySQL.
```
straight-shoota: This should be mentioned in the first sentence of this section, and also a link to travis' service documentation.

guides/ci/travis.md
```diff
@@ -0,0 +1,252 @@
...
+  - mysql -u root --password="" < test-data/setup.sql
+
+script:
+  - crystal spec
+```
+
+We are [using a `setup.sql` script](https://andidittrich.de/2017/06/travisci-setup-mysql-tablesdata-before-running-tests.html) to create a more readable `.travis.yml`. The file `./test-data/setup.sql` looks like this:
+
+```sql
+-- setup.sql
```
straight-shoota: The before_install step references this file at `test-data/setup.sql`. Maybe it should be `db/schema.sql` which is a very common path for this?

guides/ci/travis.md
```diff
@@ -0,0 +1,252 @@
...
+CREATE DATABASE IF NOT EXISTS test;
```
straight-shoota: Hardcoding the database name in the schema file is bad practice. It unnecessarily restricts the usage. For example, you can't have a production and test database on the same server.
Database creation should be handled with service setup so that shard finds the database at `DATABASE_URL` already existing.

guides/ci/travis.md
```diff
@@ -0,0 +1,252 @@
...
+-- CREATE TABLE ... etc ...
+```
+
+Pushing these changes will trigger Travis CI and the build should be successful!!
```
straight-shoota:
```suggestion
Pushing these changes will trigger Travis CI and the build should be successful.
```
Exclamation mark would be fine, but only a single one please 🙏

guides/ci/travis.md
```diff
@@ -0,0 +1,252 @@
...
+If we read Travis CI job log, we will find that every time the job runs, Travis CI needs to fetch the libraries needed to run the application:
+
+```log
+Fetching https://github.com/crystal-lang/crystal-mysql.git
+Fetching https://github.com/crystal-lang/crystal-db.git
+```
+
+This takes time and on the other hand these libraries might not change as often as our application, so it looks like we may cache them and save time.
+
+Travis CI [uses caching](https://docs.travis-ci.com/user/caching/) to improve some parts of the building path. For our example. Here is the new configuration file **with cache enabled**:
```
straight-shoota: `For our example` - is some part of the sentence missing?

guides/ci/travis.md
```diff
@@ -0,0 +1,252 @@
...
+
+```yml
+# .travis.yml
+language: crystal
+crystal:
+  - latest
+
+env:
+  - DATABASE_URL="mysql://root@localhost/test"
```
straight-shoota: MySQL doesn't make sense here, it has nothing to do with caching.

guides/ci/travis.md
```diff
@@ -0,0 +1,252 @@
...
+
+cache:
+  directories:
+    - $HOME/.cache/shards
```
straight-shoota: `$HOME/.cache/shards` caches the git repos, but they still need to be checked out on every run. `$TRAVIS_BUILD_DIR/lib` should be cached as well. Or even only `$TRAVIS_BUILD_DIR/lib`.

bcardiff: I disagree. If the local lib directory is cached then it will not mimic a fresh checkout of the repo. The caching of the `.cache/shards` directory will prevent the down of the shards which should be the main source of time penalty.

In scenarios where a new version of the dependecy is released, caching the `$TRAVIS_BUILD_DIR/lib` will prevent checking the new release.

straight-shoota: I guess you're right. When using a `shard.lock`, this shouldn't matter, though. But it's probably okay to cache only the shards cache, not checked out lib.

PR #380, opened 2020-01-02
channel usage example between classes

btihen:
**Purpose:** To document the usage of Channels for communication between classes and waiting for callbacks (to wait to end after all messages are delivered). 

**Reason:** Concurrency & lightweight Fibers are a very intersting aspect of Crystal - I found new working examples to get me started with realistic uses of Channels and Fibers.  So I thought the Crystal Git book might be a good place to add a working example using:

Crystal 0.32.1 (2019-12-18)

LLVM: 9.0.0
Default target: x86_64-apple-macosx

### DIFF
--- a/guides/concurrency.md
+++ b/guides/concurrency.md
@@ -377,3 +377,132 @@ After send
 ```
 
 Note that the first 2 sends are executed without switching to another fiber. However, in the third send the channel's buffer is full, so execution goes to the main fiber. Here the two values are received and the channel is depleted. At the third `receive` the main fiber blocks and execution goes to the other fiber, which sends more values, finishes, etc.
+
+
+### Channel usage example
+
+Here is an example of using channels to send messages between objects.
+Closing the channels & waiting for the `status` closed call-back from all User channels before ending.
+```
+class User
+  getter channel : Channel(String)
+  private getter name : String, email : String, status : Channel(Nil)
+
+  def initialize(@name, @email, @status)
+    @channel = Channel(String).new(1)
+    listen_for_messages
+  end
+
+  def send_message(message : String, receiver : User = self)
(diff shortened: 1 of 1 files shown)


straight-shoota:
Thanks for contributing this.

However, I don't know if this is the right place for it. The concurrency guide serves to give an overview about concurrency features and tries to use minimal examples to explain the concepts.
The suggested example can certainly help people understand how to use channels in a more complex example case. But it's also very specific and contains a lot of features not strictly related to concurrency, making it harder to grasp the integral parts. Actual explanations are also quite sparse.

I'm not sure how to fit this in. In the existing concurrency guide it seems to be way to specific. Maybe we could add a separate guide showcasing an exemplary concurrency application. But I'm not even sure if *this* use case makes for a good example.

btihen:
Ok - sounds good

Best wishes, 

Bill

bcardiff:
I agree with @straight-shoota . 

For this PR I think the content is more suitable for now as a post in dev.to or similar.

For usages of channels in different scenarios, I would probably like to see something as https://blog.golang.org/pipelines . Where they implement various patterns of communications.

Likely in 0.33 there will be select with timeouts and it will be a good change to document them `select` in the guide which I believe is still missing. And from there the above can be done probably better.

Closing the PR. Again thanks @btihen, and I encourage to share the code & experience still.

btihen:
I am writing my own notes of what I learn and will try to keep up with Crystal as it evolves.  
I wrote this - since I had a hard time collecting the info and knowing how to use it.  (I mostly 
Have experience with Ruby and Elixir), where this stuff either doesn’t exist or is all done for me.

I’ll definitely check the golang blog.  

Thanks

PR #385, opened 2020-02-10
Fix format in Crystal code blocks

straight-shoota:
I applied `crystal tool format` to all ```` ```crystal ```` code blocks. Plus a few manual fixes related to code blocks.


rdp:
LGTM

Fryguy:
Out of curiousty, how did you do this...was it manual, or did you run a script of some sort?  If automated, I'm wondering if it would be useful to have this as a CI task of some sort, so that new entries are auto-formatted (or at least fail CI if it detects that formatting is needed)

straight-shoota:
It's automated of course 😆 
I basically tapped into `Markd` to use a custom renderer which only handles code blocks and pipes their content to a formatter program depending on the specified language. All other lines are left in place.

The goal would be to make a generic tool from this. Such a feature sounds pretty useful for any markdown documents with code in them. It's not limited to formatting Crystal code.

I published this as a WIP to see if someone spots some issues with the results.

On a more broader scope, I'd also like to format other Markdown features as well and potentially integrate the code formatter with that. Markd currently can't be used as a basis for that because its syntax tree doesn't contain source references which would be necessary for example to fix whitespace issues and preserve existing formating when there are multiple variants.
There don't seem to be many existing tools with such a feature set. There are several linters for markdown, but most of them only identify issues and can't autofix them. There's [prettier.js](https://prettier.io/) which formats Markdown but its opinionated target style seems very questionable to me and diverges from (in my experience) typical conventions in many instances. It's also not configurable. And written in JavaScript 🤷‍♂ 

So, I'm not sure how to proceed best, but the (probably) ideal solution would be to build a Crystal tool for formatting markdown. It could be based on an enhanced Markd or a new parser which includes token info in the syntax tree. I might try to take a stab at this, but I'm not sure about prioritizing it. The implementation could be similar to `crystal tool format`, but might need at least some level of configuration like ameba or rubocop. 
The end result could be a nice and useful tool for many markdown editors and help promote Crystal =)

Fryguy:
@straight-shoota That is really awesome!

> ideal solution would be to build a Crystal tool for formatting markdown

I was thinking something very similar, then the tool can be used directly in CI for this repo, but also could be used by authors to verify their code.  I'd probably even set my editor to auto-execute this tool on saving Markdown files.

straight-shoota:
It's probably going to take some time until the automated tool is polished.  So I just merged this before it gets out of sync.

PR #386, opened 2020-02-12
Update macros.md

rdp:
Try and explain how they are useful.


rdp:
That CI failure seems unrelated to the change in question FWIW...

straight-shoota:
I don't think this addition is much helpful. There is already a link to the macro docs which includes `puts` and `run`. `puts` is also used in several examples. Maybe a section with more details about `run` could be added.

PR #390, opened 2020-02-23
Add section for inheriting documentation

Blacksmoke16:
Updates docs to include https://github.com/crystal-lang/crystal/pull/6989

### DIFF
--- a/conventions/documenting_code.md
+++ b/conventions/documenting_code.md
@@ -117,7 +117,7 @@ end
 
 ### Inheriting Documentation
 
-When using inheritance, if an instance method is only documented on the parent type; the method on the child type will inherit the documentation of the parent with `Description copied from class/struct ParentType` prepended.
+When an instance method has no doc comment, but a method with the same signature exists in a parent type, the documentation is inherited from the parent method.
 
 For example:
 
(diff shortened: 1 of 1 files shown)


conventions/documenting_code.md
```diff
@@ -115,6 +115,63 @@ class Unicorn
 end
 ```
 
+### Inheriting Documentation
+
+When using inheritance, if an instance method is only documented on the parent type; the method on the child type will inherit the documentation of the parent with `Description copied from class/struct ParentType` prepended.
```
straight-shoota: "When using inheritance" is unnecessary. This applies to every type, even for implicit inheritance from `Object`.

My suggestion: `When an instance method has no doc comment but a method with the same signature exists in a parent type, the documentation is inherited from that method.`

conventions/documenting_code.md
```diff
@@ -115,6 +115,63 @@ class Unicorn
...
+end
+
+class Unicorn < Animal
+  def name : String
+    "unicorn"
+  end
+end
+```
+
+The documentation that would appear on `Unicon#name` would be:
```
straight-shoota:
```suggestion
The documentation for `Unicorn#name` would be:
```

PR #392, opened 2020-02-27
Update concurrency.md

rdp:
Mention that waiting on channels can be tricky, with a workaround, the previously unmentioned Future.  Or is it deprecated?


rdp:
Unrelated but also it feels like that concurrency page is a bit large somehow...

bcardiff:
Noted that the exception handling could have a better general description. 
If a code example is added, the following paragraph loses context 

> In the above example we used `nil` ...

Also, using Future here kind of changes the focus and the whole example.

If you want to add a section at the end to introduce Future as useful abstraction, I'm fine with it. But I don't agree with the proposed change as it is.

rdp:
I might work on this someday, feel free anybody to clean it up in the meantime :)

PR #393, opened 2020-03-07
Update blocks_and_procs.md

rdp:
mention Tuple is different

### DIFF
--- a/syntax_and_semantics/blocks_and_procs.md
+++ b/syntax_and_semantics/blocks_and_procs.md
@@ -445,6 +445,16 @@ array.each do |number, word, bool|
 end
 ```
 
+Because [Hash#each](http://crystal-lang.org/api/Hash.html#each(&):Nil-instance-method) passes Tuples to its blocks, this means you can iterate over Hash key values with auto-splatting:
+
+```crystal
+h = {"foo" => "bar"}
+h.each do |key, value|
+  key   # => "foo"
+  value # => "bar"
+end
+```
+
 ## Performance
 
 When using blocks with `yield`, the blocks are **always** inlined: no closures, calls or function pointers are involved. This means that this:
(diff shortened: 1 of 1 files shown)


syntax_and_semantics/blocks_and_procs.md
```diff
@@ -436,6 +436,25 @@ end
 
 That means that any type that responds to `[]` with integers can be unpacked in a block argument.
 
+For [Tuple](http://crystal-lang.org/api/Tuple.html) arguments you can take advantage of auto-splatting and not need parentheses:
```
straight-shoota:
```suggestion
For [Tuple](http://crystal-lang.org/api/Tuple.html) arguments you can take advantage of auto-splatting and do not need parentheses:
```

syntax_and_semantics/blocks_and_procs.md
```diff
@@ -436,6 +436,25 @@ end
...
+
+```crystal
+array = [{1, "one", true}, {2, "two", false}]
+array.each do |number, word, bool|
+  puts "#{number}: #{word} #{bool}"
+end
+```
+
+Because [Hash#each](http://crystal-lang.org/api/Hash.html#each(&):Nil-instance-method) passes Tuples to its blocks, this means you can iterate over Hash key values with auto-splatting:
```
straight-shoota:
```suggestion
[Hash(K, V)#each](http://crystal-lang.org/api/Hash.html#each(&):Nil-instance-method) passes `Tuple(K, V)` to the block so iterating key-value pairs works with auto-splatting:
```

Fryguy:
"auto-splatting" sounds strange to me...I've always heard the term as "destructuring assignment".

bcardiff:
@Fryguy but the block does not have an assignment. And I think destructuring is a more general concept.

Issue #384, opened 2020-02-05
Reorganize content structure


straight-shoota:
While working on porting this site to Jekyll (https://github.com/crystal-lang/crystal-book/issues/261#issuecomment-582609390) it became evident that the content in this book needs to be restructured.
The current navigation structure ([SUMMAR.md](https://github.com/crystal-lang/crystal-book/raw/master/SUMMARY.md)) is very confusing. The individual pages and sections should be better arranged.

## Sections

Looking at the current content, I would roughly categorize it into three section. The contents in each section build a collection and can have different properties / visuals from others. For a clearly arranged structure, the navigation menu should only provide links to other pages in the same section.

* **Manuals:** *Using the compiler* and *The shards command* are manuals for the CLI programs coming with Crystal
* **Guides:** Besides the obvious *Guides* section, *Database* and *Overview and Examples* are essentially also guides and should be merged into this section.
* **Language:** *Syntax and semantics* contains what I would consider the actual language reference. That's obviously the biggest chunk.

I'm not sure what to do with what's currently labeled *Conventions*. *Documenting code* is a mixture of language reference, compiler manual and style guide. I suppose this should be more streamlined, the respective aspects moved to *Coding style* and the compiler manual. The heart of what's left clearly belongs to the language reference unless we consider the documentation syntax not part of the core language itself.
*Coding style* is primarily an internal resource for contributors to the compiler and stdlib. As such it could even be referenced more prominently. I think most of it is pretty much undisputed standard anyway and some even enforced by the formatter (indentation for example).
It could fit into the *Guides*  section, but maybe it could also be placed in a **Contribute**/**Developer** section, together with some pages ported from the [GitHub Wiki](https://github.com/crystal-lang/crystal/wiki) which would help centralize documentation in one place (see [Overhauling the GitHubWiki](https://forum.crystal-lang.org/t/overhauling-the-github-wiki/1303)).

## Navigating *Syntax and Semantics*

Even when reduced to only a single section, the navigation menu for *Syntax and Semantics* is still pretty huge and hard to grasp. It currently lists 108 pages, stacked up to 4 layers deep with very different sizes.

### General outline

The structure has probably grown from more humble beginnings, but now it's at a place where certain things don't make much sense. The top-level items:

* [Comments](https://github.com/crystal-lang/crystal-book/tree/master/syntax_and_semantics/comments.md)
[...]

PR #402, opened 2020-04-21
Added 'Getting started' section

ftarulla:
Related #384 

- Added `Getting started` section with `cli` and `http_server` examples (and removed `overview` section)
- Move `Using the compiler` section under `guides` section.
- Added Introduction (for Book) and updated README (for Github project)


getting_started/README.md
```diff
@@ -0,0 +1,74 @@
...
+
+Let's try it!
+
+## Crystal version
+
+We may check the Crystal compiler version. If Crystal is installed correctly then we should see something like this:
+
+```terminal-session
+$ crystal --version
+Crystal 0.33.0 (2020-02-14)
```
straight-shoota, line 17: This is already outdated at print time 😄

getting_started/cli.md
```diff
@@ -280,7 +281,7 @@ When the user presses `Enter`, then the execution will continue and `user_input`
 But what happen if the user doesn’t enter any value? In that case, we would get an empty string (if the user only presses `Enter`) or maybe a `Nil` value (if the input stream id closed, e.g. by pressing `Ctrl+D`).
 To illustrate the problem let’s try the following: we want the input entered by the user to be sang loudly:
 
-```crystal
+ ```crystal
```
bcardiff, line 284:
```suggestion
```crystal
```

getting_started/cli.md
```diff
@@ -380,7 +381,7 @@ Right now the code containing the logic of each of the applications always gets
 
 ## Using `Readline` and `NCurses`
 
-In case we want to build richer CLI applications, there are libraries that can help us. Here we will name two well-known libraries: `Readline` and `NCurses`.
+In case we want to build richer CLI applications, there are libraries that can help us. Here we will name two well-known libraries: `Readline` and` NCurses`.
```
bcardiff, line 384:
```suggestion
In case we want to build richer CLI applications, there are libraries that can help us. Here we will name two well-known libraries: `Readline` and `NCurses`.
```

getting_started/README.md
bcardiff, line 17:
```suggestion
Crystal 0.34.0 (2020-04-07)
```

getting_started/README.md
```diff
@@ -0,0 +1,74 @@
...
+
+LLVM: 9.0.1
```
bcardiff, line 19:
```suggestion
LLVM: 10.0.0
```

getting_started/README.md
ftarulla: 🙈

introduction.md
```diff
@@ -0,0 +1,15 @@
...
+
+Crystal is a programming language with the following goals:
+
+* Have a syntax similar to Ruby (but compatibility with it is not a goal).
+* Be statically type-checked, but without having to specify the type of variables or method arguments.
+* Be able to call C code by writing bindings to it in Crystal.
+* Have compile-time evaluation and generation of code, to avoid boilerplate code.
+* Compile to efficient native code.
+
+From here you can jump to anywhere you want is this document although, if you are new to the Crystal Language, we suggest to continue reading the [Getting started](./getting_started/README.md) section.
```
z64, line 13:
```suggestion
From here you can jump to anywhere you want is this document. Although, if you are new to the Crystal Language, we suggest that you continue reading the [Getting started](./getting_started/README.md) section.
```

straight-shoota:
> This is an improvement for first timers.

Sure it improves some things. Others it doesn't or only partially when it could be better. It doesn't make any sense to move things around, while we're still discussing how we want them to be moved in #384. Let's create and act on a coordinated plan, not just arbitrary changes.
Every change to page paths breaks existing deep links. Let's only break them once when we're sure how we want to structure the content (and then install redirects from the old ones).

It's also inconclusive to move compiler manual, but not shards. They obviously belong together.

> It make more sense to know how-to start coding first than having a full description of the compiler options.

That can be achieved by the following diff.
```diff
 # Summary

 * [Introduction](README.md)
-* [Using the compiler](using_the_compiler/README.md)
-* [The shards command](the_shards_command/README.md)
 * [Overview and Examples](overview/README.md)
    * [Hello World!](overview/hello_world.md)
    * [HTTP Server](overview/http_server.md)
    * [Command Line Application](overview/cli.md)
+* [Using the compiler](using_the_compiler/README.md)
+* [The shards command](the_shards_command/README.md)
 * [Syntax and semantics](syntax_and_semantics/README.md)
    * [Comments](syntax_and_semantics/comments.md)
    * **[Literals](syntax_and_semantics/literals.md)
```

> This is also improving the first page of the book.

I'm fine with that. I'm also fine with adding the new *Getting started* page. But please let's not move the other stuff around until there's a conclusion on where we want everything to be.

getting_started/README.md
```diff
@@ -0,0 +1,73 @@
+# Getting started
+
+Hi and welcome to Crystal's Reference Book!
+
+First we are going to check if we have [installed the language](https://crystal-lang.org/install/) correctly so that we may try all the examples listed in this book.
```
straight-shoota, line 5: The text shouldn't assume readers have already installed the language, might have come here first. Let's guide them to the installation instructions.

```suggestion
First let's make sure to [install the compiler](https://crystal-lang.org/install/) correctly so that we may try all the examples listed in this book.
```

getting_started/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+Let's try it!
+
+## Crystal version
+
+We may check the Crystal compiler version. If Crystal is installed correctly then we should see something like this:
+
+```terminal-session
+$ crystal --version
+Crystal 0.34.0 (2020-04-07)
```
straight-shoota, line 17: @bcardiff Can we establish a step in the dist process to make sure this is updated on a new release to always show the latest version? I'm not sure if that should just be an item on the manual TODO list or could be an automatic step.

getting_started/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+LLVM: 10.0.0
+Default target: x86_64-apple-macosx
+```
+
+Great!
+
+## Crystal help
+
+Now, if we want to list all the options given by the compiler, we may run `crystal` and that's it:
```
straight-shoota, line 27:
```suggestion
Now, if we want to list all the options given by the compiler, we may run `crystal` program without any arguments:
```

getting_started/README.md
```diff
@@ -0,0 +1,73 @@
...
+    spec                     build and run specs (in spec directory)
+    tool                     run a tool
+    help, --help, -h         show this help
+    version, --version, -v   show version
+
+Run a command followed by --help to see command specific information, ex:
+    crystal <command> --help
+```
+
+We have a section with some examples about [using the compiler](../using_the_compiler/README.md).
```
straight-shoota, line 50: The manpage should also be mentioned.
```suggestion
More details about using the compiler can be found on the manpage `man crystal` or in our [compiler manual](../using_the_compiler/README.md).
```

bcardiff:
Ok, I moved the compiler back together with the shards command after the getting started.

I agree they belong together. But since they are more a reference / glossary of all the options of the commands instead of explaining things as the user will need, I would eventually push them both down in the sidebar. But that is a discussion for another time in #384

getting_started/README.md
bcardiff: Sure, but once we move out from gitbook. To avoid some solution that is might not be the best fit on a future stack.

straight-shoota: Hm, updating a markdown file should be workable in any case. Or do you have some client-side expansion with current data in mind? That seems like overkill.

bcardiff: Currently there is no publish of the book after the release.

For a couple of releases I'm fine manually updating this.

At some point it is another case of how to validate the code samples are still valid.

Having a script that will update the .md in the repo will imply some magic string or ad-hoc process. Between doing that for one case and manually editing it, I prefer the latter.

PR #403, opened 2020-04-23
Allow external links to fail

bcardiff:
Same policy as in website


bcardiff:
This is causing the build to hang on #402 due to https://github.com/crystalshards/crystalshards/issues/2

PR #406, opened 2020-05-06
Added entry for CamelCase enum value

crush-157:
The approved style for enum values is CamelCase, as per https://github.com/crystal-lang/crystal/issues/7270

However this is not stated in the style guide, although the approved style for constants is.


straight-shoota:
Thanks for your initiative, but this is not decided yet. crystal-lang/crystal#7270 is still an open discussion.
Closing for now. We can come back to this when there is a decision.

PR #408, opened 2020-05-16
Remove incorrect statement

gerrywastaken:
The final yield expression includes a second argument of `nil`, so I'm removing this statement as it was incorrect.

Unless I'm missing something?


straight-shoota:
This has been resolved by #462

PR #409, opened 2020-05-16
Add anchor links for documentation headers

gerrywastaken:
This makes it easier to link to a specific part of a page without having
to hunt through the source to find the id for a particular title.

### DIFF
--- a/book.json
+++ b/book.json
@@ -3,7 +3,7 @@
     "links": {
         "contributePrefix": "https://github.com/crystal-lang/crystal-book/"
     },
-    "plugins": ["ga", "edit-link", "offline"],
+    "plugins": ["anchors", "ga", "edit-link", "offline"],
     "pluginsConfig": {
         "ga": {
             "token": "UA-42353458-1"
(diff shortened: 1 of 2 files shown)


straight-shoota:
I'm not sure it's worth investing further into Gitbook and focus on #261 instead

z64:
I often go out of my way to extract the header ID so I can make my own anchor link when answering questions. This change would be very much appreciated in the meantime, #261 has been open for a very long time now. Whatever we move to should support this anyways.

gerrywastaken:
@straight-shoota I'm glad to see you are looking into a better alternative.

In this case I'm not sure what the investment is. It seems like a quick easy win until #261 goes through.

Fryguy:
While I think #261 is a good long term solution, this change seems really simple and brings a very useful change :+1:

straight-shoota:
This is obsolete with #433

z64:
It would be appreciated to have the anchors directly on the headers and not have to use the right hand ToC.

straight-shoota:
Right, that was still not configured. => #446

PR #410, opened 2020-06-07
Clarify the range of octal string codepoints

zaa:
A codepoint can be represented by octal numbers in `\000` to `\377` range only.
Otherwise the compiler returns `Error: octal value too big`. See, for example, https://play.crystal-lang.org/#/r/9864

### DIFF
--- a/syntax_and_semantics/literals/string.md
+++ b/syntax_and_semantics/literals/string.md
@@ -24,7 +24,7 @@ Available escape sequences:
 "\r"                  # carriage return
 "\t"                  # tab
 "\v"                  # vertical tab
-"\888"                # octal ASCII character
+"\377"                # octal ASCII character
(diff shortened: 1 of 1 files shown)


syntax_and_semantics/literals/string.md
```diff
@@ -24,15 +24,15 @@ Available escape sequences:
 "\r"                  # carriage return
 "\t"                  # tab
 "\v"                  # vertical tab
-"\888"                # octal ASCII character
+"\377"                # octal ASCII character
```
straight-shoota, line 27: I'd rather not use the max value here. `\x888` is supposed to illustrate it's base 8. That's probably fine.

zaa: `\888` does not represent an octal number. Max would be `\777`, but even this value is misleading, if in this context the language accepts \377 as the max

PR #411, opened 2020-06-07
Provide an example of a symbol with Unicode codepoint

zaa:
As explained in the text below, a symbol can contain Unicode codepoints.
Provide an example to prove the point.


zaa:
@straight-shoota Thank you for the review. Could you please merge the changes, GitHub UI does not allow me to do that.

straight-shoota:
Thanks @zaa

PR #413, opened 2020-06-20
feat: added Int128 and UInt128 to integers docs

krthr:
Ref: https://forum.crystal-lang.org/t/docs-need-updating-for-128-bit-ints/2266


straight-shoota:
The documentation is actually missing on purpose: 128-bit literals don't work correctly. The implementation is based on Int64, so the valid range of these literals is only `Int64::MIN..Int64::MAX`.

```cr
# literal for Int64::MAX + 1
9223372036854775808_i128 # Argument error (compiler bug): Invalid Int64: 9223372036854775808
```

Until the literals are working correctly, it shouldn't be documented. (We could consider documenting the limitations, but there's no effective benefit over 64-bit literals, so there's no point in using actual 128-bit literals).

straight-shoota:
Alternative PR: #361

PR #416, opened 2020-06-30
Improve wording of finished macro description

jhass:
Not sure it's good yet, but better than the current one, suggesting that instance variables are known in the finished hook, which they are not.


straight-shoota:
Thanks @jhass

PR #420, opened 2020-08-19
Add associativity column to list of operators

radhakrishna-greyamp:
Though precedence, arity and associativity of operators have some conventional defaults, it is good to explicitly call out crystal behavior.
I know compatibility with Ruby is not a goal, but i wanted to list the differences i have noticed so far
1. unary negation has higher precedence than exponentiation (i like this)
2. exponentiation is left associative (i would have liked it to be right associative)
3. equality operator can be chained (`a == b == c` is a syntax error in ruby)


straight-shoota (changes requested):
I didn't include associativity in the original because I think it's not super useful, but I'm sure it can be helpful in some cases. In any way it makes the documentation more complete. So looks good overall 👍 

However, the explicit `<table>` code is necessary because the markdown engine used for building the HTML breaks at operators containing `|`. Github's preview renderes it correctly, but Gitbook does not. Please revert those changes.

radhakrishna-greyamp:
> I didn't include associativity in the original because I think it's not super useful, but I'm sure it can be helpful in some cases. In any way it makes the documentation more complete. So looks good overall +1
> 
> However, the explicit `<table>` code is necessary because the markdown engine used for building the HTML breaks at operators containing `|`. Github's preview renderes it correctly, but Gitbook does not. Please revert those changes.

Sorry about that. I should have guessed, they are there for a reason.

straight-shoota (approved):

straight-shoota:
Since the move from gitbook to mkdocs the HTML tables where transformed to markdown tables in #445. So I've reverted to the original commit 1fb52bd and fixed a few merging errors. Should be good to go now.

oprypin (approved):

PR #423, opened 2020-09-09
Update unless.md, change then/else block terms

jilleJr:
Was counterintuitive in the `if` statement about the naming of the "else_expression" in the "then" block


straight-shoota (approved):

straight-shoota:
THanks @jilleJr

PR #424, opened 2020-10-01
Cross compiling example

carlhoerberg:
Took a lot of googling to find out how to cross compile for an aarch64 machine, wanted to share the script I now use. It's not the most generic but for Ubuntu machines it works..


bcardiff (approved):
I think is a good starting point to give more visibility of the process

straight-shoota:
I merged this into a feature branch to combine with #177

PR #428, opened 2020-11-21
Fix list syntax for stricter Markdown parsers

oprypin:
This is needed to migrate to MkDocs but is also good on its own


straight-shoota (approved):

bcardiff:
@oprypin with this change the on the fly summary is not nested as it should. I guess this is something on the mkdocs-literate-nav plugin since it was tailored for the previous summary.md , right?

I no longer see the database section grouping their articles with this change.

oprypin:
@bcardiff Sorry, I don't understand what you mean. I don't see anything wrong in GitBook or MkDocs in the current state.

oprypin:
Oh if you're on the branch of PR #433, then yes, it doesn't have this commit in it, so it's wrong there.

bcardiff:
Oh, I see. thanks!

PR #432, opened 2020-11-21
Remove duplication between README.md and introduction.md

oprypin:
This reverts a part of #402. Hiding contribution instructions from the doc site itself is not a clear advantage, let's just merge it back.

Regarding a potential MkDocs migration: well, it complains about the extra file (never linked to), that's what prompted me.


straight-shoota:
I don't think this is a good idea. Yes, contribution instructions should not be hidden. But I'd still separate the concerns of describing the repo `crystal-book` in a `README.md` from the home page of the documentation it provides.

bcardiff:
I agree, README.md of the repo and intro of the book are different things.

Either turn off the warning for not referencing README.md or, dunno... referenced at the very bottom as an appendix (?).

straight-shoota:
I'd suggest to move all doc content to `docs/` folder as per mkdocs default (cf https://github.com/crystal-lang/crystal-book/pull/433#discussion_r528210807). Then the root folder would only contain config and meta files and content is clearly separated.

oprypin:
Yes, that's exactly what I realized as well.

PR #429, opened 2020-11-21
Restore the page overview/README.md, delete all unused & empty pages

oprypin:
Pages that are never linked to are impossible to view, and MkDocs would've complained about that.

PR #433, opened 2020-11-21
Migrate book building to MkDocs

### LINKED ISSUE #261: Move away from Gitbook

oprypin:
Fixes #261.

GitBook building still keeps working.

Required dependent PRs:

* [x] [Fix list syntax for stricter Markdown parsers](https://github.com/crystal-lang/crystal-book/pull/428)
* [x] [Replace inner links: .html -> .md](https://github.com/crystal-lang/crystal-book/pull/431)

Semi-related PRs:

* [x] [Restore the page overview/README.md, delete all unused & empty pages](https://github.com/crystal-lang/crystal-book/pull/429)
* [x] ~~[Remove duplication between README.md and introduction.md](https://github.com/crystal-lang/crystal-book/pull/432)~~

[Things to come immediately after this is merged](https://github.com/oprypin/crystal-book/commits/mkdocs) (included in the [preview](https://oprypin.github.io/crystal-book/)):

* [ ] [Move almost everything into a docs/ subdirectory](https://github.com/crystal-lang/crystal-book/pull/434)
* [ ] Finish deleting Gitbook configs
* [ ] Add CI
* [ ] Properly indent paragraphs that should be part of a list
* [ ] Use actual admonition syntax for "NOTE:"
* [ ] More consistently use section index pages

## [Site preview](https://oprypin.github.io/crystal-book/)


mkdocs.yml
```diff
@@ -0,0 +1,25 @@
...
+    accent: blue
+  icon:
+    repo: fontawesome/brands/github
+
+plugins:
+  - search
+  - same-dir
+  - literate-nav:
+      nav_file: SUMMARY.md
+  - section-index
```
oprypin, line 20: Disclaimer:
The plugins that this uses are my own.

* https://github.com/oprypin/mkdocs-same-dir
  Without it, we'd have to move every single file into a *docs* subdirectory for mkdocs to work. 
* https://github.com/oprypin/mkdocs-literate-nav
  Without it, we'd have to re-write the nav (*SUMMARY.md*) in YAML.
* https://github.com/oprypin/mkdocs-section-index
  Without it, we wouldn't be able to keep sections clickable with their own associated index pages. ([read more](https://github.com/crystal-lang/crystal-book/issues/261#issuecomment-730028305))

None of these are a very big deal to get rid of if you ever feel the need to.

straight-shoota: I'd really prefer to keep it simple and try to avoid needing additional plugins for features that don't really make a difference one way or another.

* Moving all content into `docs/` even sounds like a good idea to keep it separate from all the configuration and stuff.
* I'd rather write navigation in a structured format like YAML than a custom markdown-based variant. SUMMARY.md is written in Markdown because that's what gitbook needs. When moving to a differen generator, we can switch to the default format which is also better.
* I don't think subsection index pages are that useful and as expressed in #261 the content structure might even be better without it. I'm fine with keeping it, but it's probably not gonna work with YAML, is it?

oprypin: In terms of minimizing the diff (even keeping GitBook working) this is better as is. Then there can be small PRs specifically doing those ugly changes.

straight-shoota: Oh, I see. As a temporary solution, that's all fine.

bcardiff: I'm fine with these custom plugins.

oprypin:
So, this PR is fully ready now. It has only the bare minimum (i.e. you can run `mkdocs build` and it will work), and merging it will allow us to move on to further details of other aspects. The dependency graph basically looks like this:

```
o o o  <- all done
 \|/
  @  <- we're here
 /|\
o o O  <- drop GitBook support
    |\
    o o
```

asterite:
Wow. At first I was a bit skeptical about how it looks, but it actually looks really nice.

And the docs contents changed so much since last time I saw them, I'm impressed by how much the community improved them. They even have quotes (like the one from [Archimedes](https://oprypin.github.io/crystal-book/database/connection.html))! 👏

requirements.txt
```diff
@@ -0,0 +1,5 @@
+mkdocs
```
bcardiff, line 1: Please use pinned version

```
mkdocs==1.1.2
mkdocs-material==6.1.6
mkdocs-same-dir==0.1.0
mkdocs-literate-nav==0.2.0
mkdocs-section-index==0.2.0
```

oprypin: Hmm OK. It's just that there's no easy mechanism to auto upgrade versions in such a file.
Would you agree with `~=1.1.2`? (`>=1.1.2, ==1.1.*`)

oprypin: Oh I found a good way. Done.

bcardiff:
I was checking how to make a docker or nix environment for this and see how the livereload works.

I did the following. The livereload needs in docker 20s, and in nix 15s. Neither is fast :-(. 

The docker-compose or nix can be used later for the CI I guess.
Also, we should move the CI to github when doing this change.

```yaml
# docker-compose.yml
version: '2.0'

services:
  mkdocs:
    image: python:3.9
    volumes:
      - .:/src
      # because installing packages with --user leave stuff in /root/.local/bin
      - python-packages:/root/.local
    environment:
      - PATH=/root/.local/bin:/usr/local/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
    ports:
      - "8000:8000"
    working_dir: /src
    command: |
      /bin/sh -c " \
        pip install --user -r requirements.txt && \
        mkdocs serve --dev-addr 0.0.0.0:8000 \
      "

volumes:
 python-packages:
```


```nix
# shell.nix
{ pkgs ? import (builtins.fetchTarball {
    name = "nixpkgs-20.03";
    url = "https://github.com/NixOS/nixpkgs/archive/2d580cd2793a7b5f4b8b6b88fb2ccec700ee1ae6.tar.gz";
    sha256 = "1nbanzrir1y0yi2mv70h60sars9scwmm0hsxnify2ldpczir9n37";
  }) {} }:

let
  mach-nix = import (builtins.fetchGit {
    url = "https://github.com/DavHau/mach-nix/";
    ref = "3.0.2";
  }) {
    python = "python38";
    pypiDataRev = "62252c0d58ae1454a23dfed7c0ddcc699c729046";
    pypiDataSha256 = "1rrikdviqqdz6ivm30ij82d9m5s4xhhcwpqvd336ri085hrs1fhw";
  };

  pythonWithPackages = mach-nix.mkPython {
    requirements = (builtins.readFile ./requirements.txt); # NOTE: setuptools neeeds to be added to requirements.txt
  };
in

pkgs.mkShell {
  buildInputs = [ pythonWithPackages ];
}
```

oprypin:
@bcardiff I don't think speed of Docker is relevant because it's also the more difficult way to use this. Just build directly in 3.5 sec.

(BTW also make sure you've deleted `node_modules` and such things, because in the current setup it looks at every single file and that'd make it super slow.)

So is there any blocker? This change is "backwards-compatible" and unlocks many further changes.

straight-shoota (approved):

bcardiff:
deleting node_modules get the livereload down to 7 sec in nix-env. I rarely use python, that's why using docker or nix is appealing to me, I don't even have python3 installed :-P. 

The only change that would be needed to use nix is to add setuptools in the requirements.

But this is good to go. We can adjust later things for the CI.

bcardiff (approved):

oprypin:
@bcardiff Could you create the pull request for Docker?

bcardiff:
Yup, will do during the week. And doing the last part of the deploy, probably

mkdocs.yml
oprypin: 1. Removed in https://github.com/crystal-lang/crystal-book/pull/434 :+1: 
2. Sent a pull request to remove: https://github.com/crystal-lang/crystal-book/pull/440 :-1: 
3. Requires actual content changes, will not do :-1:

PR #435, opened 2020-11-28
Build the site with MkDocs in CI using GitHub Actions

oprypin:


.github/workflows/deploy-book.yml
```diff
@@ -0,0 +1,28 @@
+name: Deploy book
+on:
+  push:
+  pull_request:
+  schedule:
+    - cron: '0 6 * * 1'  # Every Monday 6 AM UTC
```
j8r, line 6: Why having a CRON, nothing is supposed to change if all versions are pinned.

oprypin: The real question is why not add cron

bcardiff (approved):
The cron is good. It can check that everything is still working despite this repo not getting updates.

j8r:
I don't think much people will be impacted if the repository doesn't work anymore, if it happens for some reason... unless if one run the site locally.

Also, what's prevents the scheduled cron to execute and publish a broken site? This is a double-edged sword.
Just build and check for any diff with the actual working site will be better.

oprypin:
Publishing happens only on an actual push to master.

A build failure will just be a red cross on the latest commit, nothing else.

I am not sure if it will also notify someone by email (don't know how it works in orgs, normally it notifies the owner). I would gladly personally subscribe to such emails, but I don't think that's possible.

PR #437, opened 2020-11-28
Add Google Analytics to MkDocs like it was in Gitbook

oprypin:


oprypin:
(Wasn't sure if this should be added immediately, in case it'd mess up some existing tracking, but here it is)

bcardiff:
Yes, it should :-) I going to say it afterwards. Thanks

bcardiff (approved):

j8r:
Is it really necessary? Most of use likely have some blockers that makes the stats inaccurate.
Also, they may be good to have, but are they really used by someone?

Having unique visitors, by country, pages visited, all of this can already be known by parsing logs. Apparently, can be done on S3 (https://docs.aws.amazon.com/AmazonS3/latest/dev/ServerLogs.html, https://docs.aws.amazon.com/AmazonS3/latest/dev/using-s3-access-logs-to-identify-requests.html). Another option is to use a non intrusive alternative to GA.

Finally, Amazon S3 is cheap, but still costs. Why not migrating to a free GH Pages?

j8r:
Note: for GitHub Pages, it means having a subdomain, which I think should be anyway because it a good practice for different sites https://crystal-lang.org != https://crystal-lang.org/reference. This also means better for SEO as far as I know.

PR #440, opened 2020-11-29
Migrate away from mkdocs-literate-nav plugin

oprypin:
**I am mostly against this change**, but as promised.

I made the plugin ~~dig its own grave~~ print this config out :D
```diff
diff --git a/mkdocs_literate_nav/plugin.py b/mkdocs_literate_nav/plugin.py
index 8207c38b2..fdbb9281a 100644
--- a/mkdocs_literate_nav/plugin.py
+++ b/mkdocs_literate_nav/plugin.py
@@ -47,6 +47,8 @@ class LiterateNavPlugin(mkdocs.plugins.BasePlugin):
             # Not found, return None.
 
         config["nav"] = parser.markdown_to_nav(read_index_of_dir)
+        import yaml
+        print(yaml.dump(config["nav"]))
         resolve_wildcards(config["nav"], files)
 
         return mkdocs.structure.nav.get_navigation(files, config)
```


straight-shoota:
I prefer not having an additional dependency for a solely stylistic alternative.
@oprypin Why are you (mostly) against it?

I do see a benefit in having navigation in a separate file from config. bit YAML is IMO better suited format than custom markdown semantics (so my ideal solution would be `nav.yml`).

oprypin:
Don't need to follow the strange way to specify an index page of a section.

Don't need to think about YAML escaping rules.

The format has cleaner visual separation between title and link.

Can go to [SUMMARY.md](https://github.com/crystal-lang/crystal-book/blob/master/docs/SUMMARY.md) and you can find all pages there with real links.
(Actually that page could even be published as part of the site if wanted)

oprypin:
Oh also the current format doesn't let you forget to add a section index page. Those really should be kept consistent.

straight-shoota:
I'm really not that fond of those section index pages.
Staying with `SUMMARY.md` is probably fine, still.

Blacksmoke16:
If this moves forward we'll want to also update https://github.com/crystal-lang/crystal-book#adding-a-page.

bcardiff:
I prefer to keep the SUMMARY.md as it is now. I vote closing this PR.

( I would even prefer if the summary is not needed at all, letting file names with a prefix number set the order, and a frontmater set the title 🙊 . But that would be harder to explain probably)

straight-shoota:
@bcardiff Experience shows central config is so much easier to manage. Automatic nav generation is quirky.

So let's stay literate with SUMMARY.md 👍

PR #444, opened 2020-12-01
Show off the idea of full-on literate nav

oprypin:
Let me just entertain you for a bit more with my plugin. Its current deployment is certainly not its final form.

Maybe it's controversial, and somewhat limited due to how un-nested the `syntax_and_semantics` directory is.

Anyway, I'm deleting SUMMARY.md but **the nav doesn't change**.

These lists *do* get added to the bottom of the actual page, though.
Sometimes they integrate naturally into the language of the section index (e.g. docs/getting_started/README.md), sometimes not.


straight-shoota:
👎 I don't see *any* benefit from this feature. Having the nav structure in a single location is strictly better than dispersing it across all the section index files.

oprypin:
:upside_down_face:

PR #454, opened 2021-01-18
Improve documentation on heredocs

straight-shoota:
Improves documentation on using heredocs as expression.

See crystal-lang/crystal#10268

### DIFF
--- a/docs/syntax_and_semantics/literals/string.md
+++ b/docs/syntax_and_semantics/literals/string.md
@@ -159,7 +159,7 @@ In this case, leading whitespace is not included in the resulting string.
 ## Heredoc
 
 A *here document* or *heredoc* can be useful for writing strings spanning over multiple lines.
-A heredoc is denoted by `<<-` followed by an heredoc identifier which is an alphanumeric sequence starting with a letter (and may include underscores). The heredoc starts in the following line and ends with the next line that contains *only* the heredoc identifier, optional preceeded by whitespace.
+A heredoc is denoted by `<<-` followed by an heredoc identifier which is an alphanumeric sequence starting with a letter (and may include underscores). The heredoc starts in the following line and ends with the next line that contains *only* the heredoc identifier, optionally preceeded by whitespace.
 
 ```crystal
 <<-XML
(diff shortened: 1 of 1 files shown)


bcardiff (approved):

docs/syntax_and_semantics/literals/string.md
```diff
@@ -183,20 +183,31 @@ Leading whitespace is removed from the heredoc contents according to the number
   STRING
 ```
 
-It is possible to directly call methods on heredoc string literals, or use them inside parentheses:
+A heredoc can be used like any other expression but only the first line is treated as such and essentially holds its value. The content and final line may not contain any other Crystal code.
```
asterite, line 186: This is a bit confusing. I can use interpolation inside heredó.

docs/syntax_and_semantics/literals/string.md
```diff
@@ -159,7 +159,7 @@ In this case, leading whitespace is not included in the resulting string.
 ## Heredoc
 
 A *here document* or *heredoc* can be useful for writing strings spanning over multiple lines.
-A heredoc is denoted by `<<-` followed by an heredoc identifier which is an alphanumeric sequence starting with a letter (and may include underscores). The heredoc starts in the following line and ends with the next line that starts with the heredoc identifier (ignoring leading whitespace) and is either followed by a newline or a non-alphanumeric character.
+A heredoc is denoted by `<<-` followed by an heredoc identifier which is an alphanumeric sequence starting with a letter (and may include underscores). The heredoc starts in the following line and ends with the next line that contains *only* the heredoc identifier, optional preceeded by whitespace.
```
asterite, line 162: optional - > optionally

docs/syntax_and_semantics/literals/string.md
```diff
@@ -183,20 +183,31 @@ Leading whitespace is removed from the heredoc contents according to the number
   STRING
 ```
 
-It is possible to directly call methods on heredoc string literals, or use them inside parentheses:
+A heredoc can be used like any other expression but only the first line is treated as such and essentially holds its value. The final line may not contain any Crystal code.
+It is possible to directly call methods on heredoc string literals, or use them as arguments.
```
asterite, line 187: I think it's still not clear what are the rules. Just examples aren't enough (that's why the issue was opened). I would say something like "after the heredoc identifier, and in that same line, anything that follow continues the original expression that came before the heredoc. It's as if the end of the heredoc identifier is the end of the string. However, the string contents come in subsequent lines.

straight-shoota: Yeah, that puts it much better.

asterite (approved):
Thank you!

PR #461, opened 2021-02-02
Markdown lint fixes

oprypin:
* Nested lists not recognized as such
* Multiple H1 headings
* Jumps in headings more than 1 level
* Too many newlines
* No newline around code block
* Consistency in code language identifiers
    * Tip: `git grep -hEo '```.+' | sort | uniq -c`

### DIFF
--- a/docs/SUMMARY.md
+++ b/docs/SUMMARY.md
@@ -1,4 +1,5 @@
 # Summary
+
 * [Introduction](README.md)
 * [Getting started](getting_started/README.md)
     * [An HTTP Server](getting_started/http_server.md)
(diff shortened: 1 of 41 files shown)


straight-shoota (approved):

bcardiff:
Lovely! Should we setup a lint in the CI?

oprypin:
In CI doesn't seem feasible. The lint tool had over 50% not useful findings. And some of these are something I noticed myself anyway.

j8r:
I think it will be worth it. I don't known which linter in particular you used, usually it is possible to ignore rules.

PR #462, opened 2021-02-08
Clarify usage of "arguments" and "parameters"

HertzDevil:
See crystal-lang/crystal#10374.


docs/syntax_and_semantics/blocks_and_procs.md
```diff
@@ -132,19 +132,19 @@ end
...
+It's an error specifying more block parameters than the arguments yielded:
 
 ```crystal
 def twice
   yield
   yield
 end
 
-twice do |i| # Error: too many block arguments
+twice do |i| # Error: too many block parameters
```
straight-shoota, line 143: This change is not correct until the compiler error message is updated.

docs/syntax_and_semantics/default_and_named_arguments.md
```diff
@@ -1,6 +1,6 @@
-## Default values
+## Default arguments
```
straight-shoota, line 1: I would leave this section title as is per https://github.com/crystal-lang/crystal/issues/10374#issuecomment-775140541 ff.

docs/syntax_and_semantics/blocks_and_procs.md
HertzDevil: A _lot_ other error messages are also outdated, so I simply assume that they do not correspond to exact error messages and that we should revisit them later.

straight-shoota: Well the idea is to represent the actual error message.

straight-shoota: But I suppose we can apply this change in expectation that the compiler error message is going to change as well as suggested in crystal-lang/crystal#10374.

HertzDevil: The error messages are now covered by crystal-lang/crystal#10378 (although this particular one might also be affected by the changes in crystal-lang/crystal#10251 so I updated that as well).

straight-shoota (approved):

PR #463, opened 2021-02-10
Add docs for exhaustive case expressions

HertzDevil:
I don't see it being mentioned in this repo at all, so I added some minimal examples for `case in` expressions.

### DIFF
--- a/docs/syntax_and_semantics/case.md
+++ b/docs/syntax_and_semantics/case.md
@@ -146,3 +146,84 @@ when {String, Int32}
   # and the type of value2 is known to be an Int32
 end
 ```
+
+## Exhaustive case
+
+Using `in` instead of `when` produces an exhaustive case expression; in an exhaustive case, it is a compile-time error to omit any of the required `in` conditions. An exhaustive `case` cannot contain any `when` or `else` clauses.
+
+The compiler supports the following `in` conditions:
+
+### Union type checks
+
+If `case`'s expression is a union value, each of the union types may be used as a condition:
+
+```crystal
+# var : (Bool | Char | String)?
+case var
+in String
+  # var : String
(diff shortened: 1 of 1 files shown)


asterite (approved):
Thanks again!

docs/syntax_and_semantics/case.md
```diff
@@ -146,3 +146,84 @@ when {String, Int32}
...
+
+## Exhaustive case
+
+Using `in` instead of `when` produces an exhaustive case expression; in an exhaustive case, it is a compile-time error to omit any of the required `in` conditions. An exhaustive `case` cannot contain any `when` or `else` clauses.
+
+The compiler supports the following `in` conditions:
+
+### Union type checks
+
+If `case`'s expression is a union value, each of the union types may be used as a condition:
```
straight-shoota, line 158: Type-based expressions are not limited to union types but also work with type hierarchies.

docs/syntax_and_semantics/case.md
```diff
@@ -146,3 +146,84 @@ when {String, Int32}
...
+in true
+  do_something
+in false
+  do_something_else
+end
+```
+
+### Enum values
+
+If `case`'s expression is a non-flags enum value, its enumerators may be used as conditions:
```
straight-shoota, line 190:
```suggestion
If `case`'s expression is a non-flags enum value, its members may be used as conditions, either as constant or predicate method.
```

docs/syntax_and_semantics/case.md
HertzDevil: They don't work in virtual type hierarchies?

straight-shoota:
Thank you @HertzDevil

PR #464, opened 2021-02-10
Add docs for offsetof(Tuple, NumberLiteral)

HertzDevil:
Rewrites the `offsetof` section for crystal-lang/crystal#10218.


docs/syntax_and_semantics/offsetof.md
```diff
@@ -1,9 +1,19 @@
 # offsetof
 
-An `offsetof` expression returns the byte offset of an instance variable in a struct or class type. It accepts the type as first argument and the instance variable name prefixed by an `@` as second argument:
+An `offsetof` expression returns the byte offset of a subobject in an instance of a class or struct.
```
straight-shoota, line 3: Not sure if `subobject` is an approriate term here. Maybe `property` is better?

docs/syntax_and_semantics/offsetof.md
```diff
@@ -1,9 +1,19 @@
...
+
+There are two forms of `offsetof` expressions. The first form accepts any type as first argument and an instance variable name prefixed by an `@` as second argument:
+
+```crystal
+# x : Type
+offsetof(Type, @ivar) # the byte offset of x.@ivar relative to x
```
straight-shoota, line 9: Maybe the examples could be a tiny bit larger and self-contained?

I would move the explanation that `offsetof` returns the byte offset of x.@ivar relative to x (with x : Type) to prose text outside the example.
Dito below.

docs/syntax_and_semantics/offsetof.md
```diff
@@ -1,9 +1,19 @@
...
+```
+
+The second form accepts any tuple type as first argument and an integer literal index as second argument:
```
straight-shoota, line 12:
```suggestion
The second form accepts a tuple type as first argument and an integer literal index as second argument:
```
"any" could be interpeted to include named tuples, but this only works for (non-named) tuples.

docs/syntax_and_semantics/offsetof.md
HertzDevil: "Instance property" doesn't carry the same connotation here because a "property" needn't be tied to an instance's memory representation. Maybe "instance member" or "instance field" is better.

docs/syntax_and_semantics/offsetof.md
HertzDevil: What about "`Tuple` type"? (with a link to `Tuple`'s API docs page)

docs/syntax_and_semantics/offsetof.md
straight-shoota: I suppose `field` is a common term for this.

HertzDevil:
Pinging @hugopl in case something else needs to be changed.

straight-shoota (approved):

docs/syntax_and_semantics/offsetof.md
```diff
@@ -1,20 +1,27 @@
...
 offsetof(Foo, @y) # => 8
+offsetof(Foo, @z) # => 10
 ```
+
+The second form accepts any [`Tuple`](https://crystal-lang.org/api/latest/Tuple.html) instance type as first argument and an integer literal index as second argument, and returns the byte offset of the corresponding tuple element relative to an instance of the given type:
+
+```crystal
+offsetof(Tuple(Int64, Int8, UInt16), 0) # => 0
+offsetof(Tuple(Int64, Int8, UInt16), 1) # => 8
+offsetof(Tuple(Int64, Int8, UInt16), 2) # => 10
```
oprypin, line 24: This is very curious. If I had to guess, the initial intention probably was for this syntax to look like
`offsetof({Int64, Int8, UInt16}, 0)` - which *does* work.
Of course this is an example of the duality of the type grammar/syntax.

Whether you want to represent that in the example is up to you. Because writing it your way *is* actually clearer, now that I think about it.

oprypin (approved):

oprypin (approved):

PR #470, opened 2021-03-11
Import *Crystal for Rubyists* from Github Wiki

straight-shoota:
This imports the wiki pages to this repo in the guides section, including their version history.

* https://github.com/crystal-lang/crystal/wiki/Crystal-for-Rubyists
* https://github.com/crystal-lang/crystal/wiki/Metaprogramming-Help


bcardiff:
I might've missed something here, but why this section is a good fit for the reference? It kind of goes in the opposite direction of trying to present the language by itself instead of comparing heavily on others.

I think is a better fit to remain on the wiki. Unless we expect to have a whole sections of coming from other languages.

straight-shoota:
It helps someone coming from Ruby to get acquainted with the language and it's differences from Ruby. It's not part of the language specification of course, but neither are any of the other guides in this section.

Having such resources on the wiki has several downsides. It's no good if it's not visible, doesn't get updated, or even erroneously updated and nobody notices (the wiki is free to edit for everyone but it's hard to keep track without having notifications).
Moving all documentation content to a central location largely improves discoverability and maintenance.

I'd rather move it here. Alternatively, I'd suggest to delete it. If it's nothing we want to present, then we shouldn't.

I've suggested all of this in [this forum topic](https://forum.crystal-lang.org/t/overhauling-the-github-wiki/1303) and most contents have already been moved out of the wiki and organized into different sections. Ultimately, I'd like to get rid of the wiki entirely. It's just an unnecessary extra. Public documentation can be placed with the other public documentation here. Internal documentation can be integrated into to the crystal repository proper.

straight-shoota:
For reference, Ruby has similar guides for developers coming from other languages: https://www.ruby-lang.org/en/documentation/ruby-from-other-languages/

(And that's not in the Github wiki)

straight-shoota:
I don't understand why it shouldn't be in the guides section. It is surely a guide and it shows how to do stuff in crystal.

The top-level nav is already very confusing and we should try to condense it instead of adding yet another concept. I'm fine with placing it in a sub section for guides from other languages, but until we have more of them, I'd only use that it in the URL and skip the nav item when it has only a single guide in it.

bcardiff:
I think it's a better way to classify it and would lead to a more stable url than the generic "guides".

straight-shoota:
Sorry, I don't follow what exactly is supposed to be a "better way". What section would you want to place this in?

bcardiff:
Sorry for my half-way sentences.

I mean to create a new top level catagory. "Coming from other language" and place there the Ruby guide.

straight-shoota:
Okay, thanks for clarifying. I really think the top level nav is alreaday too much crowded. We should probably continue with #384.

I'll move the URL's to the top level. That's the important factor for stability.
But for now I wouldn't introduce a new category for a single item. This is just bad for usability. There's no harm in reorganizing the nav menu later. URLs don't need to represent the full nav structure with intermediary sections, so it can definitely stay at `/crystal_for_rubyists`.

docs/crystal_for_rubyists/README.md
```diff
@@ -0,0 +1,406 @@
...
+
+### Bool
+
+`true` and `false` are of type _Bool_ rather than instances of classes _TrueClass_ or _FalseClass_.
+
+### Integers
+
+For Ruby's `Fixnum` type, use one of Crystal's integer types `Int8`, `Int16`, `Int32`, `Int64`, `UInt8`, `UInt16`, `UInt32`, or `UInt64`.
+
+If any operation on a Ruby `Fixnum` exceeds its range, the value is automatically converted to a `ignum`.
```
Fryguy, line 53:
```suggestion
If any operation on a Ruby `Fixnum` exceeds its range, the value is automatically converted to a `Bignum`.
```

beta-ziliani:
I leave it here. There are two pending issues:
 - The linter doesn't like the bullets inside the Tip block.
 - The `for` macro is broken.

oprypin:
I fixed the linter complaint. It's that that I removed all usages of this old syntax because it's not lintable.

 Though the commit message I just wrote doesn't make sense, both halves say the same thing and the other half of the meaning is gone 😅

straight-shoota:
I would just remove the example for implementing a `for` loop with macros.

straight-shoota:
Showing off the power of macros is nice. But leading devs to a non-idiomatic way of rolling their own syntax extension...😦 I don't think it has much impact though, because `for` is already pretty frowned upon in the Ruby community.
So I think it's fine. Let's merge it!

beta-ziliani:
I added it in the Getting started. If you think it's good to go, feel free to merge it @straight-shoota

beta-ziliani (approved):

PR #480, opened 2021-03-17
Improve program page in Syntax and Semantics

straight-shoota:
This adds a new page about the program to the *Syntax and Semantics* section. It's a brief description, but fills a gap of a major concept missing from the language reference.
Also moves the note about source code being UTF-8 to the new page instead of the section index.


straight-shoota:
Depends on #495

beta-ziliani (commented):
I like this refactor. WDYT about adding forward references to proper places? (what's a method, a variable, etc.)

straight-shoota:
We can add references where it makes sense. But I don't think we should link every mention.

straight-shoota:
There is also a practical challenge that we're missing clear definitions for most of these terms which we could link to. The doc tree is a bit convoluted.

docs/syntax_and_semantics/the_program.md
```diff
@@ -1,98 +1,80 @@
...
-def add(y)
-  x + y # error: undefined local variable or method 'x'
-end
+  CONST = "A::Const"
 
-add(2)
+  p! CONST   # => "A::CONST"
+  p! ::CONST # => "::CONST"
+end
 ```
```
beta-ziliani, line 63: How about adding:
```suggestion
!!! tip
   You can check the list of methods and constants included accessible from the top-level in the [API documentation](https://crystal-lang.org/api/toplevel.html).

```

straight-shoota: Yeah, I thought about that earlier as well. But it's confusing. The API docs show only what's in the standard library. Most typical program have a lot more stuff.
Perhaps this could be rephrased somehow, but I'm not sure if such a reference even makes sense here at all.

beta-ziliani: Yeah, today I don't like this either 😅

beta-ziliani (approved):

PR #482, opened 2021-03-18
Add snippet for crystal-version

straight-shoota:
The current output of `crystal version` is placed in crystal-version.txt and only this location needs to be updated when a new version is released. The content is automatically included in all places where `crystal version` output is used as an example.


bcardiff (approved):
🤩

PR #484, opened 2021-03-19
Add first tutorial lessons

straight-shoota:
This is a draft of a few first tutorial lessons. They provide a low-requirements, easy to follow introduction to the language and stdlib.

Preview: https://straight-shoota.github.io/crystal-book/feature/tutorials-initial/tutorials/01_basics/index.html

See https://forum.crystal-lang.org/t/interactive-tutorial/3019 for a primer and discussion about tutorial contents.

I'd like to get some feedback on the course of the tutorial.

Based on #479 for the embedded playground.

Resolves https://github.com/crystal-lang/crystal/issues/4462


Fryguy:
Reminds me a lot of those introductory "tours" like Tour of Go.  I really like it!

docs/tutorials/01_basics/02_variables.md
```diff
@@ -0,0 +1,89 @@
...
+message = "Hello Penny!"
+
+puts message
+puts message
+puts message
+```
+
+This program prints the string `Hello Penny!` three times to the standard output, each followed by a line break.
+
+The name of a local variable always starts with a lowercase [Unicode](https://en.wikipedia.org/wiki/Unicode) letter (or an underscore, but that's reserved for special use cases) and can otherwise consist of alphanumeric characters or underscores. As a typical convention, upper-case letter are avoided and names are written in `snake_case`.
```
Fryguy, line 18:
```suggestion
The name of a local variable always starts with a lowercase [Unicode](https://en.wikipedia.org/wiki/Unicode) letter (or an underscore, but that's reserved for special use cases) and can otherwise consist of alphanumeric characters or underscores. As a typical convention, upper-case letter are avoided and names are written in [`snake_case`](https://en.wikipedia.org/wiki/Snake_case).
```

docs/tutorials/01_basics/03_math.md
```diff
@@ -0,0 +1,130 @@
...
+```{.crystal, .crystal-play}
+p! 1 + 1,  # addition
+   1 - 1,  # subtraction
+   2 * 3,  # multiplication
+   2 ** 4, # exponentiation
+   2 / 3,  # division
+   2 // 3, # floor division
+   3 % 2,  # modulus
+   -1,     # negation (unary)
+   !4      # faculty (unary)
```
Fryguy, line 77: Not sure what this is...  `!4` == "not 4" == `false`.  I think this might have been going for factorial, but I don't think there's an operator for that?

docs/tutorials/01_basics/03_math.md
```diff
@@ -0,0 +1,130 @@
...
+
+### Number Methods
+
+Some less common math operations are not operators, but named methods.
+
+```{.crystal, .crystal-play}
+p! -5.abs,    # absolute value
+   4.3.round, # round to nearest integer
+   5.even?,   # odd/even check
+   10.gcd    # greatest common divisor
```
Fryguy, line 99: alignment

```suggestion
   10.gcd     # greatest common divisor
```

docs/tutorials/01_basics/03_math.md
```diff
@@ -0,0 +1,130 @@
...
+    A full list of number methods is available in [the Number API docs](https://crystal-lang.org/api/latest/Number.html) (also check subtypes).
+
+### Math Methods
+
+Some arithmetic methods are not defined on the number types directly, but in the `Math` namespace.
+
+```{.crystal, .crystal-play}
+p! Math.cos(1),      # cosine
+   Math.sin(1),      # sine
+   Math.tan(1),      # tangens
```
Fryguy, line 112: Typo

```suggestion
   Math.tan(1),      # tangent
```

docs/tutorials/01_basics/05_control_flow.md
```diff
@@ -0,0 +1,68 @@
...
+boolean values:
+
+```{.crystal, .crystal-play}
+a = true
+b = false
+
+p! a && b, # AND
+   a || b, # OR
+   !a,     # NOT
+   a != b, # XOR
```
Fryguy, line 35: Technically `^` is the XOR operator, though that's a bitwise operator.  `!=` works, but that's usually interpreted as "not equals", which, while accurate, seems strange.  If people search the docs for XOR, they're going to find `^`

Fryguy:
This is really great and IMO is super important for the 1.0 release!

docs/tutorials/01_basics/05_control_flow.md
straight-shoota: Yeah, I didn't want to bring bitwise operators into the picture at this point, because it's another concept and not relevant for boolean respective truthy/falsey. Not sure what the best thing to do here. Maybe keep `!=` and comment it as `inequality (=XOR)`?

Fryguy: Yeah that seems like a nice way to put it.

docs/tutorials/01_basics/05_control_flow.md
```diff
@@ -50,19 +52,151 @@ b = nil
...
+Up until now, every program in this tutorial has been a sequential series of expressions.
+Now this is going to change.
+
+### Conditionals
+
+A conditional clause puts a branch of code behind a gate that only opens if the condition is met.
+
+In a most basic form it consists of a keyword `if` followed by an expression serving as condition.
+The condition is met when the return value of the expression is *truthy*.
+All subsequent expressions are part of the branch until closes with the keyword `end`.
```
Fryguy, line 82:
```suggestion
All subsequent expressions are part of the branch until closed with the keyword `end`.
```

docs/tutorials/01_basics/05_control_flow.md
```diff
@@ -50,19 +52,151 @@ b = nil
...
+
+Per convention we indent nested branches by two spaces.
```
Fryguy, line 84:
```suggestion
Per convention, we indent nested branches by two spaces.
```

docs/tutorials/01_basics/05_control_flow.md
```diff
@@ -50,19 +52,151 @@ b = nil
...
+```
+
+!!! note
+    Technically, this program still runs in a predefined order. The fixed message always matches and makes the condition truthy.
+    But let's assume we don't define the value of the message in the source code. It could just as well come from user input,
+    for example a chat client.
+
+If the message has a value that does not start with `Hello`, the conditional branch skips and the program prints nothing.
+
+The condition expression can be more complex. With boolean algebra we can construct a condition that accepts either `Hello`
```
Fryguy, line 103: link the to the Boolean Algebra section of the tutorial?

docs/tutorials/01_basics/05_control_flow.md
```diff
@@ -50,19 +52,151 @@ b = nil
...
+if message.starts_with?("Hello")
+  puts "Hello to you, too!"
+end
+
+if !message.starts_with?("Hello")
+  puts "I didn't understand that."
+end
+```
+
+This works but there are two drawbacks: The condition expression `message.starts_with?("Hello")` evaluates twice, that's inefficient.
```
Fryguy, line 146:
```suggestion
This works but there are two drawbacks: The condition expression `message.starts_with?("Hello")` evaluates twice, which is inefficient.
```

docs/tutorials/01_basics/05_control_flow.md
```diff
@@ -50,19 +52,151 @@ b = nil
...
+And if we later change the condition in one place (maybe allow `Hi` as well), we might forget changing the other one as well.
```
Fryguy, line 147:
```suggestion
Later, if we change the condition in one place (maybe allowing `Hi` as well), we might forget changing the other one as well.
```

docs/tutorials/01_basics/05_control_flow.md
```diff
@@ -50,19 +52,151 @@ b = nil
...
+else
+  puts "I didn't understand that."
+end
+```
+
+The `else` branch still only executes if neither of the previous conditions is met. It can always be omitted, though.
+
+Note that the different branches are mutually exclusive and conditions evaluate from top to bottom.
+In the above example that doesn't matter because both conditions can't be truthy at the same time (the message can't start with both `Hello` and `Bye`).
+But we can add an alternative condition that is not exclusive to demonstrate this:
```
Fryguy, line 183:
```suggestion
However, we can add an alternative condition that is not exclusive to demonstrate this:
```

Fryguy (commented):
Some minor grammar stuff.

sdogruyol:
OMG! Seriously @straight-shoota you're just awesome! Thank you very much for this ❤️

straight-shoota:
This first set of lessons should be good to go. Ready for another round of reviews!

docs/tutorials/01_basics/02_variables.md
```diff
@@ -0,0 +1,63 @@
...
+
+The name of a variable always starts with a lowercase [Unicode](https://en.wikipedia.org/wiki/Unicode) letter (or an underscore, but that's reserved for special use cases) and can otherwise consist of alphanumeric characters or underscores. As a typical convention, upper-case letter are avoided and names are written in [`snake_case`](https://en.wikipedia.org/wiki/Snake_case).
+
+!!! note
+    The kind of variables this lesson discusses are called *local variables*.
+    Other kinds will be introduced later. For now we focus on local variables only.
+
+## Type
+
+The type of a variable is automatically inferred by the compiler. In the above example, it's [`String`](https://crystal-lang.org/api/String.cr).
```
bcardiff, line 26:
```suggestion
The type of a variable is automatically inferred by the compiler. In the above example, it's [`String`](https://crystal-lang.org/api/latest/String.html).
```

docs/tutorials/01_basics/02_variables.md
```diff
@@ -0,0 +1,63 @@
...
+You can verify this with [`typeof`](https://crystal-lang.org/api/toplevel.html#typeof(*expression):Class-class-method):
+
+```{.crystal .crystal-play}
+message = "Hello Penny!"
+
+p! typeof(message)
+```
+
+!!! note
+    [`p!`](https://crystal-lang.org/api/toplevel.html#p!(*exps)-macro) is similar to `puts` as it prints the value to the standard output, but it also prints the expression in code. This makes it a useful tool for inspecting the state of a Crystal program and debugging.
```
bcardiff, line 37:
```suggestion
    [`p!`](https://crystal-lang.org/api/latest/toplevel.html#p!(*exps)-macro) is similar to `puts` as it prints the value to the standard output, but it also prints the expression in code. This makes it a useful tool for inspecting the state of a Crystal program and debugging.
```

docs/tutorials/01_basics/02_variables.md
```diff
@@ -0,0 +1,63 @@
...
+
+p! message
+```
+
+This also works with values of a different type. The type of the variable changes when a value of a different type is assigned. The compiler is smart enough to know which type it has at which point in the program.
+
+```{.crystal .crystal-play}
+message = "Hello Penny!"
+
+p! message, typeof(message)
```
bcardiff, line 58: (idea/optional) Maybe we should include the expected output as inline comments we could use later to verify the example are still valid.

docs/tutorials/01_basics/04_strings.md
```diff
@@ -0,0 +1,254 @@
...
+
+The value of the expression doesn't need to be a string. Any type will do and it gets converted to a string representation by calling the `#to_s` method. This method is defined for any object. Let's try with a number:
+
+```{.crystal .crystal-play}
+name = 6
+puts "Hello #{name}!"
+```
+
+!!! note
+    An alternative to interpolation is concatenation. Instead of `"Hello #{name}!"` you could write `"Hello " + name + "!"`. But that's bulkier and has some gotchas with non-string types. Interpolation is generally preferred over concatenation.
```
bcardiff, line 32: Can we be specific about the gotchas? Or just suggest that interpolation is preferred?

docs/tutorials/01_basics/04_strings.md
```diff
@@ -0,0 +1,254 @@
...
+```{.crystal .crystal-play}
+message = "Hello World!"
+
+p! message.sub("World", "Crystal")
+```
+
+`#sub` only replaces the first instance of a search string. It's big brother `#gsub` applies to all instances.
+
+```{.crystal .crystal-play}
+message = "Hello World! Bye World."
```
bcardiff, line 247: Can we do another happier phrase? 🙈

docs/tutorials/01_basics/05_control_flow.md
```diff
@@ -0,0 +1,202 @@
+# Control Flow
+
+## Primitive Types
+
+### Nil
```
bcardiff, line 5: A note regarding nil not been a value of String and other reference types will make sense here.

As a motivational example we could use finding the occurrence of a string inside another, which could return nil when not present. This could replace the car analogy.

docs/tutorials/01_basics/05_control_flow.md
```diff
@@ -0,0 +1,202 @@
...
+   a != b, # inequivalence (XOR)
+   a == b  # equivalence
+```
+
+You can try flicking the values of `a` and `b` to see the operator behaviour for different input values.
+
+### Truthiness
+
+Boolean algebra however is not only defined for boolean types proper. All values have an implicit truthiness: `nil`, `false`
+and null pointers (just for completeness, we cover that later) are *falsey*. Any other value is *truthy*.
```
bcardiff, line 44: We could add 

> Note: 0 is a truthy value.

Fryguy (approved):

docs/tutorials/01_basics/02_variables.md
```diff
@@ -0,0 +1,63 @@
...
+
+message = 73
+
+p! message, typeof(message)
```
bcardiff, line 62: I would like to mention the difference between `typeof(message)` and `message.class` but to do that we need a loop at least to make a meaningful example, but that is not covered yet. So, something to keep in mind for the future.

docs/tutorials/01_basics/02_variables.md
straight-shoota: Yeah, I thought about that, too. On the other hand, if we show the expected output directly, it's less fun to click the button and actually see the code work.
I'd suggest to leave it for now and revisit this later.

docs/tutorials/01_basics/04_strings.md
straight-shoota: Yeah, I don't know. I wouldn't just want to say interpolation is preferred, without a reason. But this is also not the place to get into details.

docs/tutorials/01_basics/04_strings.md
straight-shoota: Oh, yeah 😆

docs/tutorials/01_basics/05_control_flow.md
```diff
@@ -0,0 +1,202 @@
...
+
+Controlling the flow of a program means to take different paths based on conditions.
+Up until now, every program in this tutorial has been a sequential series of expressions.
+Now this is going to change.
+
+### Conditionals
+
+A conditional clause puts a branch of code behind a gate that only opens if the condition is met.
+
+In a most basic form it consists of a keyword `if` followed by an expression serving as condition.
```
da1nerd, line 80:
```suggestion
In the most basic form it consists of a keyword `if` followed by an expression serving as the condition.
```

docs/tutorials/01_basics/05_control_flow.md
```diff
@@ -0,0 +1,202 @@
...
+message = "Hello World"
+
+if message.starts_with?("Hello") || message.starts_with?("Hi")
+  puts "Hey there!"
+end
+```
+
+Let's turn the condition around: Only print the message if it does *not*  start with `Hello`.
+That's just a minor deviation from the previous example: We can use the negation operator (`!`) to turn the condition
+into the opposite.
```
da1nerd, line 116:
```suggestion
That's just a minor deviation from the previous example: We can use the negation operator (`!`) to turn the condition
into the opposite expression.
```

da1nerd (commented):
Mostly grammatical fixes.
I really like what you've done so far!

docs/tutorials/01_basics/03_math.md
```diff
@@ -0,0 +1,141 @@
...
+As with equality, comparability is also independent of type.
+
+```{.crystal, .crystal-play}
+p! 2 > 1,
+   1 >= 1,
+   1 < 2,
+   1 <= 2
+```
+
+The main comparison operator is `<=>`, also called *Spaceship operator* for its appearance. It compares its operands and returns a value that is either zero (both operands are equal),
```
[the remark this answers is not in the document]
straight-shoota: The others are `<`, `<=`, `>=`, `>` and mentioned in the code sample directly before.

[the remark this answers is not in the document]
da1nerd: Of course. I was just questioning why the spaceship operator should stand out from among the rest, other than the fact that it's awesome :smile:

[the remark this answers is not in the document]
straight-shoota: Because it combines all of them? Maybe that should be more explicit.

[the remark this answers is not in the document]
da1nerd: Yes, I like the way you reworded it.

bcardiff (commented):
The only thing I think is worth changing is the Nil introduction as commented before.

Something like the following is enough to show how nil appears

```crystal
a = "Crystal is awesome".index("aw")
p! aw typeof(a)
b = "Crystal is awesome".index("meh")
p! b, typeof(b)
```

Then in the `if` section that same snippet could be repeated with an `if` that shows when the type narrowing is needed. `if a; str[0...a]; end` for example.

Other than that 👍

straight-shoota:
@bcardiff I wouldn't go into union types yet. It's a deeper, separate topic from an introduction to basic control directives.

docs/tutorials/01_basics/03_math.md
```diff
@@ -0,0 +1,141 @@
...
+
+A tool to structure operations are parentheses. An operator expression in parenthesis always takes precedence over external operators.
+
+```{.crystal, .crystal-play}
+p! 4 + 5 * 2,
+   (4 + 5) * 2
+```
+
+!!! info
+    The entire precedence rules are quite extensive in the [the Operator reference](../../syntax_and_semantics/operators.md#operator-precedence).
```
oprypin, line 100:
```suggestion
    All the precedence rules are detailed in the [the Operator reference](../../syntax_and_semantics/operators.md#operator-precedence).
```

docs/tutorials/01_basics/04_strings.md
```diff
@@ -0,0 +1,302 @@
...
+
+### Alternative Delimiters
+
+Some string literals may contain a lot of double quotes – think of HTML tags with quoted argument values for example. It would be cumbersome to have to escape each one with a backslash. Alternative literal delimiters are a convenient alternative. `%(...)` is equivalent to `"..."` except that the delimiters are denoted by parentheses (`(` and `)`) instead of double quotes.
+
+```{.crystal .crystal-play}
+puts %(I say: "Hello World!")
+```
+
+Escape sequences and interpolation still works the same way.
```
oprypin, line 61:
```suggestion
Escape sequences and interpolation still work the same way.
```

docs/tutorials/01_basics/04_strings.md
```diff
@@ -0,0 +1,302 @@
...
+
+!!! tip
+    You can find more info on alternative delimiters in the [string literal reference](../../syntax_and_semantics/literals/string.md#percent-string-literals).
+
+## Unicode
+
+Unicode is an international standard for representing text in many different writing systems. Besides letters of the latin alphabet used by English and many other languages, it includes several other character sets. Not just for plain text, but the Unicode standard also includes emojis and icons.
```
oprypin, line 68:
```suggestion
Unicode is an international standard for representing text in many different writing systems. Besides letters of the latin alphabet used by English and many other languages, it includes many other character sets. Not just for plain text, but the Unicode standard also includes emojis and icons.
```

docs/tutorials/01_basics/04_strings.md
```diff
@@ -0,0 +1,302 @@
...
+
+The following example uses the unicode character [`U+1F310` (*Globe with Meridians*)](https://codepoints.net/U+1F310) to address the world:
```
oprypin, line 70:
```suggestion
The following example uses the Unicode character [`U+1F310` (*Globe with Meridians*)](https://codepoints.net/U+1F310) to address the world:
```

docs/tutorials/01_basics/04_strings.md
```diff
@@ -0,0 +1,302 @@
...
+
+Working with unicode symbols can be a bit tricky sometimes. Some characters may not be supported by your editor font, some characters are not even printable. As an alternative, Unicode characters can be expressed as an escape sequence. A backslash followed by the letter `u` denotes a Unicode codepoint. The codepoint value is written as hexadecimal digits enclosed in curly braces. The curly braces can be omitted if the codepoint has exactly four digits.
+
+```{.crystal .crystal-play}
+puts "Hello \u{1F310}"
+```
+
+## Transformation
+
+Consider you want to change something about a string. Maybe scream the message and make it all uppercase?
```
oprypin, line 84:
```suggestion
Let's say you want to change something about a string. Maybe scream the message and make it all uppercase?
```

docs/tutorials/01_basics/04_strings.md
```diff
@@ -0,0 +1,302 @@
...
+puts "upcased: #{message.upcase}"
+puts "downcased: #{message.downcase}"
+puts "camelcased: #{message.camelcase}"
+puts "capitalized: #{message.capitalize}"
+puts "reversed: #{message.reverse}"
+puts "titleized: #{message.titleize}"
+puts "underscored: #{message.underscore}"
+```
+
+The methods `#camelcased` and `#underscored` don't change our string, but try them with `"snake_cased"` or `"CamelCased"`.
```
oprypin, line 102:
```suggestion
The methods `#camelcase` and `#underscore` don't change this particular string, but try them with the inputs `"snake_cased"` or `"CamelCased"`.
```

docs/tutorials/01_basics/04_strings.md
```diff
@@ -0,0 +1,302 @@
...
+`#includes?` method.
+
+```{.crystal .crystal-play}
+message = "Hello World!"
+
+p! message.includes?("Crystal"),
+   message.includes?("World")
+```
+
+What is particularly interesting is often at the beginning or end of a string. That's where the methods `#starts_with?` and `#ends_with?`
```
oprypin, line 166:
```suggestion
Sometimes the beginning or end of a string are of particular interest. That's where the methods `#starts_with?` and `#ends_with?`
```

docs/tutorials/01_basics/04_strings.md
```diff
@@ -0,0 +1,302 @@
...
+The `#sub` method is very versatile and can be used in different ways. We could also pass a search string as the first argument
+and it replaces that substring with the value of the second argument.
+
+```{.crystal .crystal-play}
+message = "Hello World!"
+
+p! message.sub("World", "Crystal")
+```
+
+`#sub` only replaces the first instance of a search string. It's big brother `#gsub` applies to all instances.
```
oprypin, line 292:
```suggestion
`#sub` only replaces the first instance of a search string. Its big brother `#gsub` applies to all instances.
```

docs/tutorials/01_basics/04_strings.md
```diff
@@ -0,0 +1,302 @@
...
+```{.crystal .crystal-play}
+message = "Hello World!"
+
+p! message[6..(message.size - 2)],
+   message[6..-2]
+```
+
+## Substitution
+
+In a very similar manner, we can alter a string. Let's make sure we properly greet Crystal and nothing else.
```
oprypin, line 273: Uh we literally can't alter a string. Maybe it's not important to stress that we *can't* but probably don't falsely say that we can.

oprypin, line 273: I think this tutorial would improve a lot from first using the already-known operations to perform the substitution manually, before introducing the method to do it.

docs/tutorials/01_basics/05_control_flow.md
```diff
@@ -0,0 +1,205 @@
...
+   !a,     # negation (NOT)
+   a != b, # inequivalence (XOR)
+   a == b  # equivalence
+```
+
+You can try flicking the values of `a` and `b` to see the operator behaviour for different input values.
+
+### Truthiness
+
+Boolean algebra however is not only defined for boolean types proper. All values have an implicit truthiness: `nil`, `false`,
```
oprypin, line 46:
```suggestion
Boolean algebra isn't limited to just boolean types, though. All values have an implicit truthiness: `nil`, `false`,
```

docs/tutorials/01_basics/05_control_flow.md
```diff
@@ -0,0 +1,205 @@
...
+   "foo" && false,
+   false || "foo",
+   "bar" || "foo"
+```
+
+The `NOT`, `XOR` and equivalence operators always return a `Bool` value (`true` or `false`).
+
+## Control Flow
+
+Controlling the flow of a program means to take different paths based on conditions.
```
oprypin, line 75:
```suggestion
Controlling the flow of a program means taking different paths based on conditions.
```

docs/tutorials/01_basics/05_control_flow.md
```diff
@@ -0,0 +1,205 @@
...
+  puts "Hello to you, too!"
+end
+
+if !message.starts_with?("Hello")
+  puts "I didn't understand that."
+end
+```
+
+This works but there are two drawbacks: The condition expression `message.starts_with?("Hello")` evaluates twice, which is inefficient.
+Later, if we change the condition in one place (maybe allowing `Hi` as well), we might forget changing the other one as well.
```
oprypin, line 150:
```suggestion
Later, if we change the condition in one place (maybe allowing `Hi` as well), we might forget to change the other one as well.
```

docs/tutorials/01_basics/06_methods.md
```diff
@@ -0,0 +1,109 @@
...
+
+!!! tip
+    Method calls are unambiguously indicated by parenthesis after the name, but it can be omitted. It would only be
+    necessary for disambiguation, for example if `say_hello` was also a local variable.
+
+## Arguments
+
+What if we want to greet different people, but all in the same manner?
+Instead of writing individual messages, we can define a method that allows customization through a parameter.
+A parameter is like a local variable inside the method body. Parameters are declared after the method name in parenthesis.
```
oprypin, line 27:
```suggestion
A parameter is like a local variable inside the method body. Parameters are declared after the method name in parentheses.
```

**And I think all of the usages in this file should be "parentheses" - plural**

docs/tutorials/01_basics/06_methods.md
```diff
@@ -0,0 +1,109 @@
...
+# This method greets *times* times.
+def say_hello(times : Int32)
+  puts "Hello " * times
+end
+
+say_hello "World"
+say_hello 3
+```
+
+Overloading is not only defined by type restrictions. The number of arguments as well as named arguments are also
```
oprypin, line 108:
```suggestion
Overloading isn't defined just by type restrictions. The number of arguments as well as named arguments are also
```

oprypin (commented):
Very nice! Thoughtfully written as *learning* material

docs/SUMMARY.md
```diff
@@ -135,3 +135,11 @@
         * [Using CircleCI](guides/ci/circleci.md)
     * [Static Linking](guides/static_linking.md)
 * [Platform Support](platform_support.md)
+* [Tutorials](tutorials/README.md)
+    * [Basics](tutorials/01_basics/README.md)
+        * [Hello World](tutorials/01_basics/01_hello_world.md)
+        * [Variables](tutorials/01_basics/02_variables.md)
+        * [Math](tutorials/01_basics/03_math.md)
```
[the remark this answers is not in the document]
straight-shoota: This clearly expresses the order of lessons. I like to see that order while editing to know which contents have already been established in previous lessons.
After the content is published I don't expect to insert new lessons in the middle because every lesson builts on the previous ones (to some degree) and new contents should be added to the end. In case there is a need to insert a new lesson somewhere in between, it could just get an existing number, maybe with a order suffix.

docs/tutorials/01_basics/04_strings.md
straight-shoota: Hm, what would be a better term then?

straight-shoota: What manual operations do you mean? Extracting substrings and stitching them back together?

oprypin: Yes. `index`, `[..]`, `+`

bcardiff:
@straight-shoota I don't think there is a need to explain unions to use values that are unions.

For the nil section if the typeof an union wants to be avoided the following example is enough to show that an operation might return numbers or nil.

```crystal
a = "Crystal is awesome".index("aw")
p! a
b = "Crystal is awesome".index("meh")
p! b
```

For the if section we could include the following example that will fail until the user uncomment the `if` because we need to be sure that `" " * a` is invoked with a number and not a `nil`. (And later we can hint that `if a` can be used instead of `if !a.nil?`).

```crystal
str = "Crystal is awesome"
a = str.index("aw")

# if !a.nil?
  puts str
  print " " * a 
  puts "^^"
# end
```

I think that is enough to show a real example of Nil, get the usual error message on nilable types and how to solve it. No need to mention union.

docs/SUMMARY.md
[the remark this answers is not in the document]
bcardiff: We can use 10 / 20 / 30 numbers to insert things in the middle if we need. I usually do that. Leftovers of GW Basic 🙈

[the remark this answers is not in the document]
straight-shoota: That sounds like a good idea. We should avoid having more than 10 lessons in a course, so that fits.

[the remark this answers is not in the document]
straight-shoota: And I suppose the courses don't need to be ordered, so we can remove the prefix from `01_basics`.

bcardiff (approved):

docs/SUMMARY.md
```diff
@@ -135,3 +135,11 @@
...
     * [Static Linking](guides/static_linking.md)
 * [Platform Support](platform_support.md)
+* [Tutorials](tutorials/README.md)
+    * [Basics](tutorials/basics/README.md)
+        * [Hello World](tutorials/basics/10_hello_world.md)
+        * [Variables](tutorials/basics/20_variables.md)
+        * [Math](tutorials/basics/30_math.md)
+        * [Strings](tutorials/basics/40_strings.md)
+        * [Control Flow](tutorials/basics/50_control_flow.md)
+        * [Methods](tutorials/basics/60_methods.md)
```
bcardiff, line 145: I think you missed to commit the rename of 06_methods to 60_methods

docs/tutorials/01_basics/04_strings.md
straight-shoota: I'll add that in a follow-up.

straight-shoota:
The first lessons are online: https://crystal-lang.org/reference/tutorials/basics/index.html

Thank you for all the detailed and helpful reviews! ❤️ 

I'll follow up with the suggested step of manual string replacement as suggested in https://github.com/crystal-lang/crystal-book/pull/484#discussion_r600846945

PR #487, opened 2021-03-21
Add a CI guide for GitHub Actions, update Hosting on GitHub

oprypin:
I have quite generously sprinkled this with links to my own (relevant) projects, hope that's OK....


bcardiff (approved):

docs/guides/ci/gh-actions.md
```diff
@@ -0,0 +1,216 @@
...
+Most projects will have external dependencies, ["shards"](https://github.com/crystal-lang/shards#usage). Having declared them in `shard.yml`, just add the installation step into your workflow (after `install-crystal` and before any testing):
+
+```yaml
+      - name: Install shards
+        run: shards install
+```
+
+### Latest or locked dependencies?
+
+If your repository has a checked in `shard.lock` file (typically good for applications), consider the effect that this has on CI: `shards install` will always install the exact versions specified in that file. But if you're developing a library, you probably want to be the first to find out in case a new version of a dependency breaks the installation of your library -- otherwise the users will, because the lock doesn't apply transitively. So, strongly consider running `shards update` instead of `shards install`. And then it makes sense to add [scheduled runs](https://www.jeffgeerling.com/blog/2020/running-github-actions-workflow-on-schedule-and-other-events) to your repository.
```
straight-shoota, line 109: Perhaps mention that for a library you maybe shouldn't check in `shard.yml` in the first place?

docs/guides/ci/gh-actions.md
```diff
@@ -0,0 +1,216 @@
...
+## Enforcing code formatting
+
+If you want to verify that all your code has been formatted with [`crystal tool format`](../writing_shards.md#coding-style), add the according check as a step near the end of the workflow. If someone pushes code that is not formatted correctly, this will break the build just like failing tests would.
+
+```yaml
+      - name: Check formatting
+        run: crystal tool format --check
+```
+
+Consider also adding this check as a *Git pre-commit hook* for yourself.
```
straight-shoota, line 131: Not sure this is is much helpful without further explanation. Is there any resource it could link to?

oprypin: > Is there any resource it could link to?

Uhh... no, actually! I looked.

The official documentation is super dry, just technical details about them. On the other hand, random blog posts are quite ephemeral and incomplete. The best option here is for the user to search this phrase with a search engine and get an overview that way, which is the intent here. I just didn't want to literally link to a particular search engine.

straight-shoota: Oh, actually there is https://github.com/crystal-lang/crystal/blob/master/scripts/git/pre-commit

I remembered I did something with git hooks back in the day. It shoud work with any crystal repo, not just crystal-lang/crystal.

straight-shoota: Also https://github.com/crystal-lang/crystal/blob/master/CONTRIBUTING.md#git-pre-commit-hook

straight-shoota: Maybe there could be a guide in this book? Or mentioned in a guide. I'll look into that.

oprypin: The instruction is specific to the repo itself, though:

https://github.com/crystal-lang/crystal/blob/dd40a2442fa186add8a82b74edb14a90aa1dae05/scripts/git/pre-commit#L10

docs/guides/ci/gh-actions.md
oprypin: OK. I didn't want to express an explicit opinion on that within this guide, but done now.

docs/guides/ci/gh-actions.md
straight-shoota: Yes, for others it's `curl -sL https://github.com/crystal-lang/crystal/raw/master/scripts/git/pre-commit > .git/hooks/pre-commit`.
Maybe should add that.

oprypin: I would like to not add details about commit hooks in this guide. If in the future this is described in a different guide then this can be changed to a link.

straight-shoota: Yes, just leave it like that. I'll update the script and add it to the book separately.

straight-shoota (approved):

PR #488, opened 2021-03-23
Added shebang scripting guide to Getting Started

NightMachinery:
Please squash the commits yourself, I used the web interface which is not very powerful for these kinds of stuff.

### DIFF
--- a/docs/getting_started/README.md
+++ b/docs/getting_started/README.md
@@ -70,5 +70,6 @@ Hello World!
 
 Here we have two more examples to continue our first steps in Crystal:
 
+- [Scripting](./scripting.md)
 - [HTTP Server](./http_server.md)
 - [Command Line Application](./cli.md)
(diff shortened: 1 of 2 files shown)


straight-shoota:
Thanks for this contribution!

I'm not sure, where it fits best into the document structure. It pretty certainly shouldn't be the first topic after the getting started page. I'd rather see this in the guides section, probably.

If it's to be a guide, there should definitely be some introductory text, explaining what it is about and why it's useful.

I don't think the benchmark section is very useful. Percieved performanc of CLI applications is pretty subjective. When `crystal run` is fast enough for your use case, you don't need to worry if a cached tool could be a bit more performant. And there are more considerations to this than just a simple benchmark. Like how often do you run that code, how often does it change, how do you install the dependency etc. So I'd probably just leave that section out.

I'm not familiar scriptista, but I'd suggest to add an example for a more purpose-specific tool like https://github.com/Val/crun I don't know if scriptista has any benefit over that. It might offer more flexibility but it requires configuration which makes it more difficult to use.

beta-ziliani:
I'm closing this given that there was no movement for a year. @NightMachinery feel free to ask us to reopen it if you'd like to address it again.

PR #494, opened 2021-03-26
Fix gcd example call

straight-shoota:
Fixes https://forum.crystal-lang.org/t/interactive-tutorial/3019/13?u=straight-shoota


asterite (approved):
I saw this one too! :-)

PR #496, opened 2021-03-26
Add example of primitiv string substitution

straight-shoota:
Follow-up on https://github.com/crystal-lang/crystal-book/pull/484#discussion_r600847339

/cc @oprypin

### DIFF
--- a/docs/tutorials/basics/40_strings.md
+++ b/docs/tutorials/basics/40_strings.md
@@ -222,6 +222,8 @@ p! b, typeof(b)
 
 !!! tip
     We'll cover `nil` more deeply in the next lesson.
+    For now it should be sufficient to know that calling `.not_nil!` on the result
+    lets the program compile. It allows to use the value if it's not `nil`, and errors otherwise.
 
 ## Extracting Substrings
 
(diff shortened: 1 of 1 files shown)


bcardiff:
I don't think not_nil! should be shown/encouraged in the basic tutorial.

I prefer if/nop or if/raise.

In the methods/classes tutorial we could define a `#trust_me_is_not_nil` and show the reader there is nothing special about not_nil and then mention its existence in the std-lib.

But presenting it as in this PR sounds like we are encouraging it's use a lot.

straight-shoota:
I'm not comfortable with promoting `.not_nil!` either.

But what else could it be? We need some way to make `String#index` usable. Maybe there should just be `String#index!` 😆 
Conditionals are only introduced in the next lesson... 

I suppose we might be able to do something like `index || abort "Search string not found"`. Boolean algebra is also introduced in the next lesson, but maybe it's okay to just say "use this, it'll be explained later".

Any other ideas?

oprypin:
What's wrong with `if !search_index.nil?`

straight-shoota:
`if` gets introduced in the next lesson.

bcardiff:
If the tutorials are to be read in strict order, then the things introduced in this PR should not go intro Basic. 

Maybe a followup as Text manipulation technique section is a better fit than a basic tutorial. This way you can assume that basic knowledge is covered. I think that from "Indexing Substrings" until the end could be moved out of basic if we do this.

Otherwise I would use `if search_index` with a reference to the next section. Any other alternative is harder to understand for a newcomer. 

My vote is to keep the basics short and have another section to describe more use cases were we can assume all basic tutorial are covered.

PR #498, opened 2021-03-26
Fix link to md file in README

straight-shoota:
Follow-up on https://github.com/crystal-lang/crystal-book/pull/497#issuecomment-808499065

PR #499, opened 2021-03-26
Disallow unchecked intra-site links

oprypin:
In CI, use regular expressions to check against the kinds of links that could be made safer by letting MkDocs check their validity:

* Links not ending with '.md' (but the check is specifically for '.html' and '/' endings)
* Links starting with '/' or containing the site URL itself.

Closes #498

### DIFF
--- a/.github/workflows/deploy-book.yml
+++ b/.github/workflows/deploy-book.yml
@@ -21,6 +21,15 @@ jobs:
           restore-keys: pip-
       - name: Build book
         run: make build
+      - name: Disallow unchecked intra-site links
+        run: |
+          if grep -P '\[\w.*?\]\((?!http)[^ )]*?(\.html|/)(#[^ )]*?)?\)' docs/**/*.md; then
+            echo "Links within the site must end with .md"; exit 1
+          fi
+          if grep -P '\]\(/|//crystal-lang.org/reference/' docs/**/*.md; then
+            echo "Absolute links within the site are disallowed, use relative links instead"; exit 1
+          fi
+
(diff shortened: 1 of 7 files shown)


bcardiff:
Should this be a bin/ci script that can be run locally directly also?

straight-shoota:
Makefile recipe?

oprypin:
Ideally this would be a MkDocs plugin :D
Actually this, along with my previous ideas (namely, making links starting with `/` work correctly), is finally enough to justify a plugin

But yeah this is wishful thinking.

Back to reality, ... I don't know where to put this and make it look nicely.

@straight-shoota if you know how to put this conditional into makefile nicely, do you want to send a PR to my PR :smile:

oprypin:
Ah by the way, `grep -P` is not normally available on Mac, so that's a significant downside of forcibly running this

straight-shoota (approved):

bcardiff (approved):
Next step should be to check for valid relative links as it's done in the website, right? Or is mkdocs already preventing bad internal links?

oprypin:
Well that's the point - MkDocs checks internal links, so we prevent situations where it doesn't check them

straight-shoota:
The website actually only checks for external links. https://github.com/crystal-lang/crystal-website/blob/af07adfccb686e5890b2dd6f687458d55eede2a6/Makefile#L26
But we could add that here, too. It just adds a dependency on a Ruby tool, which was trivial to sneek in the Gemfile for crystal-website :D

PR #503, opened 2021-03-30
Check Crystal code blocks for syntax validity, formatting

oprypin:
and, only for `crystal-play` blocks, compilability.

I made a plugin to accomplish this: https://github.com/oprypin/mkdocs-code-validator


straight-shoota (approved):

straight-shoota:
So this increases build time in CI by about 1 minute (from ~30 seconds to 1:30). I suppose that's still acceptable.

oprypin:
Those numbers aren't correct, it's just that https://github.com/crystal-lang/crystal-book/pull/447 clumped dependency installation into the step.

oprypin:
And I think you're comparing a run with cached deps to one without

oprypin:
The site build time changes from ~6 sec to ~35 sec in CI.
(But both numbers are halved for me locally.)

So I'd expect with all things equal that this adds +30 sec to CI, not +60 sec.

PR #504, opened 2021-04-03
Add docs for values of `while` expressions

HertzDevil:
Adds docs and examples for crystal-lang/crystal#10566.


straight-shoota (approved):

HertzDevil:
ping

docs/syntax_and_semantics/until.md
```diff
@@ -13,4 +13,4 @@ while !some_condition
 end
 ```
 
-`break` and `next` can also be used inside an `until`.
+`break` and `next` can also be used inside an `until`, and like `while` expressions, `break`s may be used to return values from an `until`.
```
oprypin, line 16:
```suggestion
`break` and `next` can also be used inside an `until`, and like in `while` expressions, `break`s may be used to return values from an `until`.
```

docs/syntax_and_semantics/while.md
```diff
@@ -46,3 +44,78 @@ loop do
...
+The value of a `while` is the value of the `break` expression that exits the `while`'s body:
+
+```crystal
+a = 0
+x = while a < 5
+  a += 1
+  break "four" if a == 4
+  break "three" if a == 3
+end
+x # => "three"
```
oprypin, line 59: This is still `nil` on the latest version of Crystal, so this PR must not be merged yet.

And I think we'll need to define some kind of policy for that:

1. When do such doc updates get merged?
    1. Should they be merged into a branch first?
2. How to indicate that a feature is new since some version?

docs/syntax_and_semantics/while.md
```diff
@@ -46,3 +44,78 @@ loop do
...
+```
+
+If the `while`'s condition fails, `nil` is returned instead:
```
oprypin, line 62:
```suggestion
If the `while` loop ends normally (because its condition became false), the value is `nil`:
```

docs/syntax_and_semantics/while.md
straight-shoota: Good questions! But I think they should be asked in a dedicated issue, not in the comment thread on an affected pull request. 🙇‍♂️

oprypin (approved):

oprypin (approved):
The release happened, so this is good to merge.

Couldn't find a way to re-run CI.

PR #505, opened 2021-04-03
Add local variables and constants to the assignments 

ShalokShalom:
Local variables and Constants are assignments, this is where people are looking for them. 

Thanks


straight-shoota:
I don't think it's a good idea to cramp local variables and constants into the page about assignments. They're independent concepts and there's more to them than just assignments. So they should stay as separate pages.

However, I do agree that the assignment page should help readers learn about variables and constants. That's currently very poor. But I think it's a far better solution to just add references to those concepts, instead of incorporating them entirely.

ShalokShalom:
I am fine with both. 

They are still assignments, and the implementation-specific details are something that happen to fit into the developer handbook.

PR #508, opened 2021-04-19
Improve assignment page

straight-shoota:
Closes #505


docs/syntax_and_semantics/assignment.md
```diff
@@ -11,45 +20,58 @@ local = 1
...
 
-```crystal
-local += 1  # same as: local = local + 1
+This is an example of a call to a setter method in typical method notation and with assignment operator.
+Both expressions are equivalent.
 
-# The above is valid with these operators:
-# +, -, *, /, %, |, &, ^, **, <<, >>
+```crystal
+def name=(value);end
```
oprypin, line 42: Why is this all on one line? :o

docs/syntax_and_semantics/assignment.md
```diff
@@ -11,45 +20,58 @@ local = 1
...
+# Assigns to a constant
+CONST = 4
+
+# Assigns to a setter method
+foo.method = 5
+foo[0] = 6
 ```
 
-Each of the above kinds of variables will be explained later on.
+### Method as assignment target
```
oprypin, line 32: May I suggest putting the section "Method as assignment target"  below "Combined assignments"

docs/syntax_and_semantics/assignment.md
```diff
@@ -11,45 +20,58 @@ local = 1
...
+Both expressions are equivalent.
 
-# The above is valid with these operators:
-# +, -, *, /, %, |, &, ^, **, <<, >>
+```crystal
+def name=(value);end
 
-local ||= 1 # same as: local || (local = 1)
-local &&= 1 # same as: local && (local = 1)
+name=("John")
```
oprypin, line 44: Sadly, this example is disingenuous, because top-level calls don't actually pick up any methods. So these are still just `name  =  ("John")`.
Probably need to wrap this into a class and maybe even mention the point I'm bringing up.

```crystal
def f=(x)
  p x
end

f=("John")
```
(no output)

docs/syntax_and_semantics/assignment.md
```diff
@@ -11,45 +20,58 @@ local = 1
...
+Some syntax sugar that contains the `=` character is available:
 
-# The above can be written as:
-objects[2, 3]
+```crystal
+local += 1  # same as: local = local + 1
 ```
 
-The `=` operator syntax sugar is also available to setters and indexers. Note that `||` and `&&` use the `[]?` method to check for key presence.
+The `=` operator syntax sugar is also available to setter and index assignment methods.
```
oprypin, line 73: Add to this sentence something like "(and this assumes that the corresponding getter methods exist)"

docs/syntax_and_semantics/assignment.md
straight-shoota: I guess I wanted to emphasize that the implementation of the method is irrelevant.

docs/syntax_and_semantics/assignment.md
straight-shoota: Why? Setter methods are a more basic feature IMO.

docs/syntax_and_semantics/assignment.md
straight-shoota: Wondering if that's intended... Even the explicit reference `::f=("John")` doesn't work.

oprypin: Yes, I'm quite sure it's intended. No receiver -> always means assignment to a local variable. Same as with `self.a = 5` not having an equivalent `a = 5` even though getting `a` works.

docs/syntax_and_semantics/assignment.md
```diff
@@ -58,6 +61,9 @@ list.[]=(2, 3)
 list[2] = 3
 ```
 
+NOTE: This does not apply to top-level methods. An assignment `x = y` in the top-level scope is always an assignment to
```
oprypin, line 64: NOTE: This will just be the literal text `NOTE:`

Unless you're ready to enable https://oprypin.github.io/markdown-callouts/ :D

oprypin (approved):

docs/syntax_and_semantics/assignment.md
straight-shoota: Yeah I was completely ignorant of that aspect :D

oprypin (approved):

PR #511, opened 2021-04-30
Add documentation for chaining of comparisons and equalities

### LINKED ISSUE #510: Documentation Request: Compiler rewrites Comparisons like a <= b <= c to a <= b && b <= c
I came along some Crystal Code where someone used a comparison like `a == b == c` where `a,b,c` are Numbers.
I was very confused that this worked. Initially, I would have thought that the compiler does not allow this or that at least it is always `false` since even if `a == b` is `true`, `true == c` should be `false` if `c` is a Number.

I did not find anything about that in the Documentation but I finally discovered the Code for the Compiler that handles this:  
https://github.com/crystal-lang/crystal/blob/a7544b5f0f3e23d05cb4622ff4db3cbb7f1b996a/src/compiler/crystal/semantic/normalizer.cr#L100

I like the feature very much as it is really nice syntactic sugar. However, without documentation I find it rather confusing.
Maybe I just did not find this behavior in the documentation but if this is indeed not documented, I think it would be important to add it. Probably, https://github.com/crystal-lang/crystal-book/blob/master/docs/syntax_and_semantics/operators.md#comparison is a good place.  
Let me know what you think about this. I'm also happy to prepare a PR if it is indeed not documented yet and you also think that this should be added to the Docs :)

Maybe there are also some other "rewritings" done by the compiler that should be documented as well.

monsterkrampe:
fixes #510 

As discussed in #510, the compiler allows to chain some of the comparisons and equality together but this was not documented yet.  
In this PR, I now added a little section about this.

I also investigated this a little more and noticed that in some cases compile time errors occur, e.g. for `a <= b == c`.
This is because the precedence handling (I think this is done here: https://github.com/crystal-lang/crystal/blob/a7544b5f0f3e23d05cb4622ff4db3cbb7f1b996a/src/compiler/crystal/syntax/parser.cr#L537) interferes with the normalization (https://github.com/crystal-lang/crystal/blob/a7544b5f0f3e23d05cb4622ff4db3cbb7f1b996a/src/compiler/crystal/semantic/normalizer.cr#L100).
I tried to describe this a little bit in the documentation as well.

Probably it is worthwhile to discuss this issue also in https://github.com/crystal-lang/crystal since I think that this behavior is far from being obvious and it seems to depend on the implementation details of the compiler currently.
Maybe this can be fixed by adjusting the parsing or normalization process but I'm not sure right now which implications this might have.

Any feedback is very much appreciated :)

### DIFF
--- a/docs/syntax_and_semantics/operators.md
+++ b/docs/syntax_and_semantics/operators.md
@@ -229,7 +229,7 @@ For example `a <= b <= c` is treated as `a <= b && b <= c`
 and it is even possible to mix operators of the same 
 [operator precedence](#operator-precedence) 
 like `a >= b <= c > d`. 
-It is not advised to chain operators of different precedences since this may lead to compile time errors.
+Operators with different precedences can be chained too, however, it is advised to avoid it, since it is makes the code harder to understand. For instance `a == b <= c` is interpreted as `a == b && b <= c`, while `a <= b == c` is interpreted as `a <= (b == c)`. 
 
 ### Logical
 
--- a/docs/syntax_and_semantics/operators.md
+++ b/docs/syntax_and_semantics/operators.md
@@ -224,20 +224,12 @@ proven faster than equality).
 #### Chaining Equality and Comparison
 
 The compiler allows for `==`, `!=`, `===`, `<`, `>`, `<=`, and `>=` 
-to be chained together.  
+to be chained together. 
 For example `a <= b <= c` is treated as `a <= b && b <= c`
-and it is even possible to mix operators like `a == b <= c > d`.
-
-There are cases where this does not work due to 
-[operator precedence](#operator-precedence),
-e.g. `a <= b == c` results in a compile time error
-while `a == b <= c` works fine.
-To ensure precedence, the compiler 
-treats `a <= b == c` as `a.<=(b.==(c))` and throws an error since `<=` cannot be used with a `Bool`. The same thing happens for `(a == b) <= c` or `(a.==(b)).<=(c)`.
-However, `a == b <= c` is just treated as `a.==(b).<=(c)` where the compiler can infer that this is a chain of comparisons/equalities and produces `a == b && b <= c`.
-
-Because of this rather confusing behavior, it is advised to only make use of this feature
-in obvious cases, e.g. `a <= b <= c`.
+and it is even possible to mix operators of the same 
+[operator precedence](#operator-precedence) 
+like `a >= b <= c > d`. 
+It is not advised to chain operators of different precedences since this may lead to compile time errors.
 
 ### Logical
 
--- a/docs/syntax_and_semantics/operators.md
+++ b/docs/syntax_and_semantics/operators.md
@@ -223,8 +223,8 @@ proven faster than equality).
 
 #### Chaining Equality and Comparison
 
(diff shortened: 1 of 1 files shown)


beta-ziliani (changes requested):
Thanks for your PR! However, it needs some fixes (let me know if you need help):

Use a positive voice: instead of saying that it's confusing, note that it works with operators having the same precedence, as in `a <= b < c`, and this is useful. This is a better hint than "obvious cases". Then, I would say that it is _not advisable_ to mix operators precedences. 

Sounds good?

monsterkrampe:
Thanks for your feedback!

I agree that it's better to focus on what works and not so much on what does not work, especially in the Docs.  
I have now clarified that only operators of the same precedence should be mixed and I do not go into detail anymore of why mixing different precedences may lead to erros. This may just be confusing in the docs and depends very much on the implementation of the compiler.

Let me know what you think about the refined version :)

docs/syntax_and_semantics/operators.md
```diff
@@ -219,6 +221,16 @@ proven faster than equality).
 | `>=` | greater or equal | `1 >= 2` | yes |
 | `<=>` | comparison | `1 <=> 2` | yes |
 
+#### Chaining Equality and Comparison
+
+The compiler allows for `==`, `!=`, `===`, `<`, `>`, `<=`, and `>=` 
+to be chained together.
```
straight-shoota, line 227: This is a documentation of the language, not the compiler (that's just implementation), so it's better to not refer to compiler behaviour, but language behaviour. I'd also suggest adding more explanation to "chaining" together.

```suggestion
Equality and comparison operators `==`, `!=`, `===`, `<`, `>`, `<=`, and `>=` 
can be chained together and are interpreted as a compound expression. 
```

docs/syntax_and_semantics/operators.md
```diff
@@ -219,6 +221,16 @@ proven faster than equality).
...
 
+#### Chaining Equality and Comparison
+
+Equality and comparison operators `==`, `!=`, `===`, `<`, `>`, `<=`, and `>=` 
+can be chained together and are interpreted as a compound expression. 
+For example `a <= b <= c` is treated as `a <= b && b <= c`
+and it is even possible to mix operators of the same 
+[operator precedence](#operator-precedence) 
+like `a >= b <= c > d`. 
+It is not advised to chain operators of different precedences since this may lead to compile time errors.
```
beta-ziliani, line 232: Minor suggestion:

```suggestion
Operators with different precedences can be chained too, however, it is advised to avoid it, since it is makes the code harder to understand. For instance `a == b <= c` is interpreted as `a == b && b <= c`, while `a <= b == c` is interpreted as `(a <= b) == c`. 
```

monsterkrampe: This really makes things more clear :)

Minor thing:
`a <= b == c` is interpreted as `a <= (b == c)`, right?

beta-ziliani: Nice catch! Fixed

beta-ziliani (approved):
Thanks @monsterkrampe !

straight-shoota (approved):

monsterkrampe:
My pleasure :)
Thanks for your remarks as well!

oprypin:
@beta-ziliani FYI pull requests are always or almost-always squash-merged in Crystal

PR #387, opened 2020-02-14
Update coding_style.md

Piterden:

PR #515, opened 2021-05-14
Replace "underscore-case" in coding_style.md

NeoLight1010:
I think "snake-case" is the more popular term.


straight-shoota:
Duplicate of #387 which, unfortunately, has been stalled for way too long 😢

oprypin:
#387 is merged now. Thanks anyway, as you likely bumped it enough to be merged :)

PR #516, opened 2021-05-23
Remove Void outside of references to C code

mattrberry:
`Void` is a construct that is only relevant to C bindings. Referencing it outside of that explicit context could be confusing, and could lead people to use it where it's not meant to be used. (Evidence: me, about a year and a half ago :stuck_out_tongue:)

### DIFF
--- a/docs/syntax_and_semantics/capturing_blocks.md
+++ b/docs/syntax_and_semantics/capturing_blocks.md
@@ -45,7 +45,7 @@ def some_proc(&block : Int32 ->)
 end
 
 proc = some_proc { |x| x + 1 }
-proc.call(1) # nil
+proc.call(1) # => nil
 ```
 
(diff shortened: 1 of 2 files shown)


straight-shoota:
Related: https://github.com/crystal-lang/crystal/issues/10657

docs/syntax_and_semantics/capturing_blocks.md
```diff
@@ -45,7 +45,7 @@ def some_proc(&block : Int32 ->)
 end
 
 proc = some_proc { |x| x + 1 }
-proc.call(1) # void
+proc.call(1) # nil
```
straight-shoota, line 48: Could use the notation for inspecting the result value here:
```suggestion
proc.call(1) # => nil
```

straight-shoota (approved):

oprypin (approved):

PR #518, opened 2021-05-27
Fix z-index of carcin-play buttons

### LINKED ISSUE #517: Incorrect displaying button "Run code" on tutorial pages
There is some issue with the button "Run code" on all tutorial pages. I think it has to do with using the wrong z-index CSS prop.
The following link has this issue as an example:
https://crystal-lang.org/reference/tutorials/basics/40_strings.html

зображення

зображення

зображення

Solving issue by using z-index: 9:
зображення
зображення

Software and hardware I use:
1. Mozilla Firefox 88.0.1 (64-bit)
2. 8 Gb RAM DDR-4
3. macOS Big Sur 11.2.3
4. 2.3 GHz Quad-Core Intel Core i5

straight-shoota:
Resolves #517

### DIFF
--- a/docs/assets/style.css
+++ b/docs/assets/style.css
@@ -3,8 +3,8 @@
      * original value is z-index: 2 */
     z-index: 20;
 }
-.md-overly {
-  z-index: 15;
+.md-overlay {
+    z-index: 15;
 }
 
 .md-header-nav__button.md-logo {
--- a/docs/assets/style.css
+++ b/docs/assets/style.css
@@ -59,3 +59,7 @@
     content: "·";
     margin: 0 0.3em;
 }
+
+.carcin-play__actions > button.md-button {
+  z-index: 9;
+}
--- a/docs/assets/style.css
+++ b/docs/assets/style.css
@@ -51,7 +51,3 @@
     content: "·";
     margin: 0 0.3em;
 }
-
-.carcin-play__actions > button.md-button {
-  z-index: 9;
-}
--- a/docs/assets/style.css
+++ b/docs/assets/style.css
@@ -1,7 +1,10 @@
-.md-header {
+.md-header, .md-sidebar--primary {
     /* pull z-index up to be above the codemirror gutters (z-index: 3).
      * original value is z-index: 2 */
-    z-index: 10;
+    z-index: 20;
+}
+.md-overly {
+  z-index: 15;
 }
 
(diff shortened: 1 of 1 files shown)


oprypin:
The file has all of this:

```css
.md-header {
    /* pull z-index up to be above the codemirror gutters (z-index: 3).
     * original value is z-index: 2 */
    z-index: 10;
}
.md-header {
    /* place it above codemirror gutter */
    z-index: 10;
}
@media screen and (max-width: 76.1875em) {
    .md-sidebar--primary {
        /* place it above codemirror gutter */
        z-index: 10;
    }
}
.carcin-play__actions > button.md-button {
        z-index: 9;
}
```

You can just do this:
```css
.md-header, .md-sidebar--primary  {
    /* pull z-index up to be above the codemirror gutters (z-index: 3) and carcin-play button (z-index: 10).
     * original value is z-index: 2 */
    z-index: 20;
}
```

The removal of `@media screen` is unrelated but I think it's wrong to have it here. If the theme changes to not use those specific measurements anymore, then what, the sidebar will be above the button only for some screen widths?

oprypin:
While we're at the topic, I also see that `.codemirror-hscrollbar` and the button are above `.md-overlay` still

image

So that'd perhaps be
```css
.md-overlay {
    /* original value is z-index: 3 */
    z-index: 15;
}
```

docs/assets/style.css
```diff
@@ -1,7 +1,10 @@
-.md-header {
+.md-header, .md-sidebar--primary {
     /* pull z-index up to be above the codemirror gutters (z-index: 3).
      * original value is z-index: 2 */
-    z-index: 10;
+    z-index: 20;
+}
+.md-overly {
+  z-index: 15;
```
oprypin, line 7: Typo + mismatch in indentation

```suggestion
.md-overlay {
    z-index: 15;
```

docs/assets/style.css
```diff
@@ -59,3 +51,7 @@
     content: "·";
     margin: 0 0.3em;
 }
+
+.carcin-play__actions > button.md-button {
+  z-index: 9;
+}
```
oprypin, line 57: Part of my comment was that with the higher numbers in other places this addition is no longer needed

```suggestion
```

oprypin (approved):

PR #520, opened 2021-06-01
Add documentation for splat expansions in array / tuple literals

HertzDevil:
Add wording for crystal-lang/crystal#10429.

### DIFF
--- a/docs/syntax_and_semantics/literals/array.md
+++ b/docs/syntax_and_semantics/literals/array.md
@@ -80,3 +80,32 @@ The type arguments can be explicitly specified as part of the type name:
 ```crystal
 Set(Int32){1, 2, 3}
 ```
+
+## Splat Expansion
+
+The splat operator can be used inside array and array-like literals to insert multiple values at once.
+
+```crystal
+[1, *coll, 2, 3]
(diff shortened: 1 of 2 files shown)


docs/syntax_and_semantics/literals/array.md
```diff
@@ -80,3 +80,32 @@ The type arguments can be explicitly specified as part of the type name:
 ```crystal
 Set(Int32){1, 2, 3}
 ```
+
+## Splat Expansion
+
+The splat operator can be used inside array and array-like literals to insert multiple values at once.
+
+```crystal
+[1, *coll, 2, 3]
```
oprypin, line 89: What does "coll" stand for? :o
Somehow the fact that I can't guess it makes me more confused than if it was a random combination of letters

oprypin (approved):
This should not be merged if CI fails. But it fails only because the syntax is too new. So it should be merged immediately after release.

If https://github.com/crystal-lang/crystal-book/pull/529 is resolved and then `master`-specific CI is added, we can remove this limitation.

oprypin (approved):
The release happened, so this is good to merge.

Couldn't find a way to re-run CI.

PR #521, opened 2021-06-01
Add instructions for gh-action status display

docelic:
Updates gh-action.md page to provide an example of adding status badge to project's README.


straight-shoota:
This seems reasonably useful, but I'm hesitant because it is not specific to Crystal. This is a pretty generic GHA feature. You can read about putting a badge in a readme anywhere on the web. IMO we should keep these instructions more focused on the essential parts for Crystal projects.

oprypin:
Same as @straight-shoota for me. Not strongly against it.

PR #533, opened 2021-07-30
Update crystal-version

straight-shoota:


asterite:
How about we remove the versión output? Do we expect to keep updating this on every release? It sounds cumbersome.

straight-shoota:
It's actually used in a couple of places. Not necessarily super important to have, but I think it's a nice and helpful addition.

I've added this to the release checklist, it's a rather simple step. https://github.com/crystal-lang/distribution-scripts/pull/107
It should also be easy to automate (which we should be looking into to simplify the release process).

oprypin (approved):

PR #535, opened 2021-08-09
docs: explain dir and tag name conflict;

ondreian:
because of [this](https://github.com/crystal-lang/crystal/blob/e2903dbbc42e5668b9a1f950869b0871efe592c3/src/compiler/crystal/command/spec.cr#L40-L45)
if there is a directory with the same name as your tagged tests, your tests will exit with status code zero and not actually run the tests tagged when using the currently documented `--tag <name>` syntax.

### DIFF
--- a/docs/guides/testing.md
+++ b/docs/guides/testing.md
@@ -137,6 +137,7 @@ it "is slow", tags: "slow" do
   true.should be(true)
 end
 
+# can be ran using --tag="fast"
 it "is fast", tags: "fast" do
   true.should be(true)
 end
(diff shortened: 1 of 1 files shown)


docs/guides/testing.md
```diff
@@ -146,6 +147,9 @@ Tagging an example group (`describe` or `context`) extends to all of the contain
 
 Multiple tags can be specified by giving an [`Enumerable`](https://crystal-lang.org/api/latest/Enumerable.html), such as [`Array`](https://crystal-lang.org/api/latest/Array.html) or [`Set`](https://crystal-lang.org/api/latest/Set.html).
 
+Be sure to use `--tag="my-tag"` syntax when there is a directory name that matches your tag name in your project or else they will be parsed as directories instead of tag filters.  For example, if you have some tests that run in a Terraform environment and you have put all of your terraform scripts in `terraform/example.tf` and have also tagged some tests as `tags: "terraform"`
```
oprypin, line 150: I don't think "Terraform" has enough relevance here to be used without any introduction

Fryguy:
Is there a bug opened in Crystal issues for this?  It feels weird to document this when it's possible we can just fix it.

ondreian:
i didn't open a bug in Crystal, but I certainly can.

docs/guides/testing.md
ondreian: thanks for the review.

I often find concrete examples useful in docs when dealing with corner cases, but i can definitely clean it up when I'm back from vacation. 

Unless it is decided this behavior is actually a crystal bug we should fix.

straight-shoota:
Yes, let's discuss this first in the Crystal repo. 

Initial diagnosis: https://libera.irclog.whitequark.org/crystal-lang/2021-08-09#30646868;

PR #550, opened 2021-09-25
Remove references to Travis CI

HertzDevil:
Following crystal-lang/crystal#10690 and https://forum.crystal-lang.org/t/migrate-ci-in-crystal-lang-projects/3363, we no longer officially support Travis CI anymore. This PR removes any references to it in the guides.


straight-shoota:
Just to be clear, there are two independent aspects and neither of them is really about Crystal supportin travis:

* The default `travis.yml` in `crystal init`. That has been removed. Thus, references should be removed from the book as well.
* The guide about using Crystal with travis. This requires travis to support Crystal which really isn't the case anymore, at least not with `language: crystal` (see https://github.com/travis-ci/travis-build/pull/1987). You could probably still use Crystal on travis with docker. But the guide isn't about that. So it's probably best to remove it.

oprypin (approved):

PR #555, opened 2021-11-03
Add docs for splats on the left-hand side of multiple assignments

HertzDevil:
Adds wording for crystal-lang/crystal#10410.

### DIFF
--- a/docs/syntax_and_semantics/assignment.md
+++ b/docs/syntax_and_semantics/assignment.md
@@ -171,6 +171,64 @@ objects[1] = temp1
 objects[2] = temp2
 ```
 
+## Splat assignment
+
+The left-hand side of an assignment may contain one splat, which collects any values not assigned to the other targets. A [range](literals/range.md) index is used if the right-hand side has one expression:
+
+```crystal
+head, *rest = [1, 2, 3, 4, 5]
+
+# Same as:
+temp = [1, 2, 3, 4, 5]
+head = temp[0]
+rest = temp[1..]
+```
+
+Negative indices are used for targets after the splat. [`Indexable`](https://crystal-lang.org/api/latest/Indexable.html) types support negative indices out of the box:
+
+```crystal
+*rest, tail1, tail2 = [1, 2, 3, 4, 5]
+
+# Same as:
+temp = [1, 2, 3, 4, 5]
+rest = temp[..-3]
+tail1 = temp[-2]
+tail2 = temp[-1]
+```
+
+If the expression does not have enough elements and the splat appears in the middle of the targets, [`IndexError`](https://crystal-lang.org/api/latest/IndexError.html) is raised:
+
+```crystal
+a, b, *c, d, e, f = [1, 2, 3, 4]
+
+# Same as:
+temp = [1, 2, 3, 4]
+if temp.size < 5 # number of non-splat assignment targets
+  raise IndexError.new("Multiple assignment count mismatch")
+end
+# note that the following assignments would incorrectly not raise if the above check is absent
+a = temp[0]
+b = temp[1]
+c = temp[2..-4]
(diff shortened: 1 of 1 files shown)


straight-shoota (approved):

oprypin (approved):
This is good. But more could be added to the bottom. In another PR if you prefer.

* That `*_` can be used to drop the values.
* That merely using the splat adds the size check.
    * That `a, b, c = foo` is the same as `a, b, c, *_ = foo` currently.

oprypin:
Oh wow, so you needed to go back to the PR itself and implement what I assumed was true 😅

Thanks

oprypin (approved):

oprypin:
Well, we can't merge this yet, though...

HertzDevil:
With the version selector now available, what would happen if we merge this right now?

oprypin:
@HertzDevil Sorry, it's not quite ready yet. I want to merge *all* the things about the version selector and then branch *1.2* in this repo and then merge your contributions to *master* branch.

What would happen, indeed, is that this would show up only at https://crystal-lang.org/reference/master/syntax_and_semantics/assignment.html and then, when 1.3 is branched off master, also at https://crystal-lang.org/reference/1.3/syntax_and_semantics/assignment.html

oprypin:
Now it's time to *merge master* into here and expect CI to pass.

docs/syntax_and_semantics/assignment.md
```diff
@@ -171,6 +159,97 @@ objects[1] = temp1
...
+name, age, source = "Crystal, 123, GitHub".split(", ")
+
+# The above is the same as this:
+temp = "Crystal, 123, GitHub".split(", ")
+name = temp[0]
+age = temp[1]
+source = temp[2]
+```
+
+Additionally, if the [`strict_multi_assign` flag](compile_time_flags.md) is provided, the number of elements must match the number of targets, and the right-hand side must be an [`Indexable`](https://crystal-lang.org/api/latest/Indexable.html):
```
straight-shoota, line 176:
```suggestion
Additionally, if the [`strict_multi_assign` flag](compile_time_flags.md) is provided, the number of elements must match the number of targets, and the right-hand side must be an [`Indexable`](https://crystal-lang.org/api/Indexable.html):
```

docs/syntax_and_semantics/assignment.md
```diff
@@ -171,6 +159,97 @@ objects[1] = temp1
...
+  raise IndexError.new("Multiple assignment count mismatch")
+end
+name = temp[0]
+age = temp[1]
+source = temp[2]
+
+a, b = {0 => "x", 1 => "y"} # Error: right-hand side of one-to-many assignment must be an Indexable, not Hash(Int32, String)
+```
+
+## Splat assignment
```
straight-shoota, line 193:
```suggestion
### Splat assignment
```

beta-ziliani: I don't think this change is correct. Splats also work if there are several targets on the right-hand side, as in `a, *_, b = 1, 2, 3, 4`. The previous section is for when you have just one expression on the right.

beta-ziliani: Ah, my comment is locally correct but globally wrong :-) We need to increase depth level of he previous one too.

docs/syntax_and_semantics/assignment.md
```diff
@@ -171,6 +159,97 @@ objects[1] = temp1
 objects[2] = temp2
 ```
 
+## One-to-many assignment
```
beta-ziliani, line 162: The following two sections are part of _Multiple assignments_
```suggestion
### One-to-many assignment
```

beta-ziliani (approved):

PR #562, opened 2021-12-10
Use the version of Crystal according to the current branch

oprypin:
* Downloaded version of Crystal depends on the branch

  So the syntax validity checks for `1.2` branch will use latest Crystal 1.2.*, and for `master` branch will use Crystal nightly.

* Links in API docs are populated based on the branch

  Links get rewritten from https://crystal-lang.org/api/latest/ to the actual current latest version that matches the branch.


straight-shoota:
What about instead of adjusting links to `/api/latest/*`, we do that for links to `/api/*` (when no version or keyword follows).
Benefit would be shorter URLs in code. And `/api/latest/*` is more explicit about linking to latest instead of the version that matches that of the docs.

oprypin:
I don't have a significant preference, just that that's what we already have in the vast majority of cases. Maybe do that in a separate PR, just in the interest of keeping this one cleaner?

straight-shoota:
Well, this PR changes the semantics of link URLs and I think it makes sense to figure these out beforehand instead of fixing it up later.

Since we're rewriting the URLs anyways, we could change them even more. Make them shorter and simpler. Maybe just `/api/*` without authority component? And leave all absolute URLs unprocessed.
I suppose that's more difficult to implement probably? Because as I understand it, the rewrite hooks operates on plain text and not on what markdown considers a URL.

oprypin:
It's really not difficult to implement but I honestly believe that keeping the links as is is by far the best option because

 1. the links aren't broken-by-default and can still be viewed on GitHub.
 2. it's simpler to copy the URL from the browser without having to then edit it and remember the special prefix.

I don't mind figuring out the semantics but I do mind squashing a mass-rewrite into a commit that has a lot of functional changes. If you want, we could do the rewrite *before* merging this.

oprypin:
What I suggest:

Keep the links like this for this PR
https://crystal-lang.org/api/latest/Crystal/Macros.html#flag%3F%28name%29%3ABoolLiteral-instance-method

Then immediately send another PR to change all links to
https://crystal-lang.org/api/Crystal/Macros.html#flag%3F%28name%29%3ABoolLiteral-instance-method

-And this will also need some CI enforcement.

---

I oppose changing the links to some special prefix. But even then I'd prefer to do it in a separate PR.
Unless you're willing to deal with a merge-merge rather than a squash-merge.

oprypin:
FYI my main plan now is actually to merge #565 *first* and then immediately this.

And is there any remaining concern for these PRs, in your view?

oprypin:
Ah actually this PR will need some updates after the other PR #565, so I would definitely *not* merge this one *immediately* after.

oprypin:
Now this one is ready

straight-shoota (approved):

PR #564, opened 2021-12-10
Migrate lint checks from Makefile to Python (within mkdocs build)

oprypin:


straight-shoota:
Just wondering: What's the benefit of this approach? It looks pretty similar to what we already have.

oprypin:
It integrates into mkdocs builds, including every rebuild with `mkdocs serve`, you don't need to run it manually.
And the way it is displayed looks very natural compared to other warnings that are built into MkDocs.
And I plan to add more to this Python file, so may as well put all similar things into it.

oprypin:
And uh Python is a better programming language than Bash

PR #565, opened 2021-12-10
Enforce all API docs links to not contain a version

oprypin:
Prefixes such as /latest/ /master/ /1.2.2/ will be dropped.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -73,7 +73,7 @@ clean_all: clean clean_deps clean_vendor
 
 .PHONY: format_api_docs_links
 format_api_docs_links:
-	echo $(DOCS_FILES) | xargs sed -i -E -e 's|https?://(www\.)?crystal-lang.org/api/([A-Z])|https://crystal-lang.org/api/latest/\2|g'
+	echo $(DOCS_FILES) | xargs perl -p -i -e 's@\bhttps?://crystal-lang\.org/api/(?!toplevel|[A-Z])[^ )/]+/([^ )]+\.html\b)@https://crystal-lang.org/api/\1@g'
 
 .PHONY: help
 help: ## Show this help
--- a/Makefile
+++ b/Makefile
@@ -73,7 +73,7 @@ clean_all: clean clean_deps clean_vendor
 
 .PHONY: format_api_docs_links
 format_api_docs_links:
-	echo $(DOCS_FILES) | xargs perl -p -i -e 's@\bhttps?://crystal-lang\.org/api/(?!toplevel|[A-Z])[^ )/]+/([^ )]+\.html\b)@https://crystal-lang.org/api/\1@g'
+	echo $(DOCS_FILES) | xargs sed -i -E -e 's@\bhttps?://(www\.)?crystal-lang\.org/api/([0-9]+(\.[0-9]+)+|latest|master)/([^ )]+\.html)\b@https://crystal-lang.org/api/\4@g'
 
 .PHONY: help
 help: ## Show this help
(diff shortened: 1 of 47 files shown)


Makefile
```diff
@@ -73,7 +73,7 @@ clean_all: clean clean_deps clean_vendor
 
 .PHONY: format_api_docs_links
 format_api_docs_links:
-	echo $(DOCS_FILES) | xargs sed -i -E -e 's|https?://(www\.)?crystal-lang.org/api/([A-Z])|https://crystal-lang.org/api/latest/\2|g'
+	echo $(DOCS_FILES) | xargs perl -p -i -e 's@\bhttps?://crystal-lang\.org/api/(?!toplevel|[A-Z])[^ )/]+/([^ )]+\.html\b)@https://crystal-lang.org/api/\1@g'
```
straight-shoota, line 76: I think the regex is working from a wrong angle. It was already like this before, but when we're already changing it, let's make it better.
It shouldn't make any assumptions about the internal paths that don't represent a version prefix. Those could easily change. If additional paths get into the docs generator, they might not be covered by this regex and would be incorrectly recognized as a version prefix.
Instead of excluding paths, the regex should include paths that match the structure of the versions setup.
Something like this:
```suggestion
	echo $(DOCS_FILES) | xargs perl -p -i -e 's@\bhttps?://crystal-lang\.org/api/(master|nightly|\d+(?:\.\d+)+/[^ )/]*([^ )]+\.html\b)@https://crystal-lang.org/api/\1@g'
```

Any reason you removed the `www` subdomain?

oprypin: Well I think this is the correct angle. The doc generator hasn't gained any other page names in 5+ years, and when it does, we'll be fully aware of it. Meanwhile the regex as is can catch a lot of things that are possibly even mistakes. Or branch names that weren't foreseen, like `main`. I mean, you already wrote `master|nightly`, only one of which exists -- that's somewhat strange. Anyway just not to argue here, I'll just change it.

Seems nobody in the history of the world has ever written `www.crystal-lang.org/api` so that's why I thought it wasn't relevant. Maybe it is better to keep it here, but I didn't want to keep it in the Python script. Well maybe I'll put it back.

straight-shoota: You're right that the doc generator's structure doesn't change a lot. Maybe I'd like to have that too much 😁
But there's still a differentiation of deployment structure vs. content paths.

`master|nightly` was intentional because they're sometimes confused and we might end up separating master and nightly builds at some point. We're not going to have either as a content path, so it's totally fine.

Outright incorrect paths are not really an issue, they'd be caught by the link checker.

oprypin: I think if we'll be changing to `nightly` urls then we'll notice this along with all the other places that need to be changed. So I didn't put that in. Otherwise, done.

Makefile
```diff
@@ -73,7 +73,7 @@ clean_all: clean clean_deps clean_vendor
 
 .PHONY: format_api_docs_links
 format_api_docs_links:
-	echo $(DOCS_FILES) | xargs perl -p -i -e 's@\bhttps?://crystal-lang\.org/api/(?!toplevel|[A-Z])[^ )/]+/([^ )]+\.html\b)@https://crystal-lang.org/api/\1@g'
+	echo $(DOCS_FILES) | xargs sed -i -E -e 's@\bhttps?://(www\.)?crystal-lang\.org/api/([0-9]+(\.[0-9]+)+|latest|master)/([^ )]+\.html)\b@https://crystal-lang.org/api/\4@g'
```
straight-shoota, line 76: I believe the regex won't match paths that contain a `/` in the path after the version prefix (such as `/api/latest/Foo/Bar.html`).

oprypin: Hm no, it definitely catches them.
There's the version, then 1 slash, then anything except space or parenthesis. That "anything" includes slashes

oprypin: Anyway I made sure to try it out on the site

straight-shoota: Ah totally right. I was confused by `[^ )/]` in the original code.

straight-shoota (approved):

PR #569, opened 2021-12-16
Document required libraries

### LINKED ISSUE #634: Port "required libraries" from wiki
[Required libraries](https://github.com/crystal-lang/crystal/wiki/All-required-libraries) are somewhat version dependent, so it makes sense for this page to be versioned in the book.

HertzDevil:
This is a first attempt at crystal-lang/crystal#11488 for the reference manual. It does not include Windows specifics. Fixes #634.

There is some overlap of information with https://github.com/crystal-lang/crystal/wiki/All-required-libraries; ideally we should move that wiki article into the reference manual too.


docs/required_libraries.md
```diff
@@ -0,0 +1,47 @@
+# Required libraries
+
+This is a list of known third-party libraries required by the Crystal compiler and the standard library.
+
+## Runtime dependencies
```
straight-shoota, line 5: I would suggest to call this `Core dependencies` because *runtime* is ambiguous.

However, I think it might actually be better to merge all stdlib dependencies into a single table and use some indicator for core dependencies. Maybe a separate column? This column could also hold information about the stdlib modules that use this dependency. But the current solution in prose text is also good.

docs/required_libraries.md
```diff
@@ -0,0 +1,47 @@
...
+
+The following libraries are required by Crystal's runtime, and must be present when building or running any program, unless the `--prelude=empty` compiler option is passed.
+
+| Library | Description | License |
+|---------|-------------|---------|
+| [PCRE][libpcre] | Perl Compatible Regular Expressions. Implements [`Regex`](https://crystal-lang.org/api/Regex.html). | [BSD](http://www.pcre.org/licence.txt)
+| [Boehm GC][libgc] | The Boehm-Demers-Weiser conservative garbage collector. Performs automatic memory management. | [MIT-style](https://github.com/ivmai/bdwgc/blob/master/LICENSE)
```
straight-shoota, line 12:
```suggestion
| [Boehm GC][libgc] | The Boehm-Demers-Weiser conservative garbage collector. Performs automatic memory management for the `GC` module (can be disabled with `-Dgc_none`). | [MIT-style](https://github.com/ivmai/bdwgc/blob/master/LICENSE)
```

docs/required_libraries.md
```diff
@@ -0,0 +1,47 @@
...
+| [Libevent][libevent] | An event notification library. Implements concurrency features such as [`Fiber`](https://crystal-lang.org/api/Fiber.html) and the event loop. | [Modified BSD](https://github.com/libevent/libevent/blob/master/LICENSE)
```
straight-shoota, line 13: libevent is really just the event loop and IO integrations of course. `Fiber` itself is not directly affected by libevent.

docs/required_libraries.md
```diff
@@ -0,0 +1,47 @@
...
+| [compiler-rt builtins][compiler-rt] | Provides optimized implementations for low-level routines required by code generation, such as integer multiplication. Several of these routines are ported to Crystal directly. | [MIT / UIUC][compiler-rt]
```
straight-shoota, line 14: Where are we actually still using compiler-rt builtins? Is there a way to find out?

straight-shoota (commented):
This is great 👍

The most important library is missing, though: `libc` (including `dl`, `pthreads`, `m`)

HertzDevil:
I no longer have time for this, feel free to take over the PR.

libffi and PCRE2 are currently missing

straight-shoota:
I have enhanced the lib list. It is really a lot 🙈

beta-ziliani (approved):

straight-shoota (approved):

PR #574, opened 2022-01-05
Reland: Use the version of Crystal according to the current branch

oprypin:
Currently pull requests don't see a valid branch name, so the selection of Crystal version (to validate against) doesn't work at all for them.

This is a fixup for #562


straight-shoota (approved):
Thanks ❤️

PR #576, opened 2022-01-05
Adding autocasting

beta-ziliani:
Perfection is the enemy of good (but comments are welcomed).

### DIFF
--- a/docs/syntax_and_semantics/autocasting.md
+++ b/docs/syntax_and_semantics/autocasting.md
@@ -21,8 +21,8 @@ end
 
 foo 0xFFFF_u16 # OK, an UInt16 always fit an Int32
 bar 0xFFFF_u16 # OK, an UInt16 always fit an Float32
-bar (foo 1)    # Fails, casting an Int32 to a Float32 might lose precision
-bar64 (bar 1)  # OK, a Float32 can be autocasted to a Float64
+bar(foo 1)     # Fails, casting an Int32 to a Float32 might lose precision
+bar64(bar 1)   # OK, a Float32 can be autocasted to a Float64
 ```
 
 Number literals are always casted. Expression are casted too (like in the last example above), unless the flag `no_number_autocast` is passed to the compiler (see [Compiler features](compile_time_flags.md#compiler-features)).
--- a/docs/syntax_and_semantics/autocasting.md
+++ b/docs/syntax_and_semantics/autocasting.md
@@ -1,0 +1,50 @@
+# Type autocasting
+
+Crystal transparently casts elements of certain types when there is no ambiguity.
+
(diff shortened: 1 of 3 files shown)


straight-shoota (commented):
I think there are a couple of important aspects missing:

* Autocasting happens at method calls (this one is definitely necessary to understand the topic, it's not stearing towards unnecessary perfection 😄 )
* Number literals are autocasted if the *concrete value* fits the target type. Non-literal values can't be checked at compile time, so they can only autocaste when their *type* fits the target type.
* Add an example for ambiguity that makes autocasting fail (could take the one from the release post)

docs/syntax_and_semantics/autocasting.md
```diff
@@ -0,0 +1,66 @@
+# Type autocasting
+
+At function calls, Crystal transparently casts elements of certain types when there is no ambiguity.
```
straight-shoota, line 3: Oh, I was wrong there. Autocasting is not limited to function calls. It also applies to assignments. Maybe I missed something else?
```suggestion
Crystal transparently casts elements of certain types when there is no ambiguity.
```

docs/syntax_and_semantics/autocasting.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Number autocasting
+
+When passing a number to a function, values of a type are autocasted to a larger one if no precision is lost:
```
straight-shoota, line 7:
```suggestion
Values of a numeric type autocast to a larger one if no precision is lost:
```

docs/syntax_and_semantics/autocasting.md
```diff
@@ -0,0 +1,66 @@
...
+  x
+end
+
+foo 0xFFFF_u16 # OK, an UInt16 always fit an Int32
+foo 0xFFFF_u64 # OK, this particular Int64 fit in an Int32
+bar(foo 1)     # Fails, casting an Int32 to a Float32 might lose precision
+bar64(bar 1)   # OK, a Float32 can be autocasted to a Float64
+```
+
+Number literals are always casted when the actual value of the literal fits the target type, despite of its type. Expressions are casted too (like in the last example above), unless the flag `no_number_autocast` is passed to the compiler (see [Compiler features](compile_time_flags.md#compiler-features)).
```
straight-shoota, line 28: I think the last sentence should be moved to a separate paragraph, it's unrelated to autocasting of literals.

docs/syntax_and_semantics/autocasting.md
beta-ziliani: I added an example for the other situation where this works.

straight-shoota (approved):

PR #578, opened 2022-01-06
Update crystal-version.txt

straight-shoota:
Update the contents for 1.3.0


oprypin (approved):

oprypin:
I think this was already brought up once -- it's not that important to update this version, it's just an example.

But if we do think it's important, might as well set up automation for it -- in the current setup we can actually just call `crystal -v > crystal-version.txt` in CI and it will always give the appropriate result.

Issue #542, opened 2021-08-26
Document the `loop` keyword


SamantazFox:
I've seen that used a ton of times across the code, but it's not mentionned in the Control expressions.

From the stdlib:
https://github.com/crystal-lang/crystal/blob/dd40a2442fa186add8a82b74edb14a90aa1dae05/spec/std/socket/unix_socket_spec.cr#L85
https://github.com/crystal-lang/crystal/blob/6a3f6b1e89be89dbe32a66a11da92d35e7790977/src/uri/punycode.cr#L71-L77

oprypin:
> "it is highlighted as such"

That's a bug in their highlighter, then.

SamantazFox:
> `loop` is not a keyword. It's a method implemented in stdlib: https://crystal-lang.org/api/latest/toplevel.html#loop(&)-class-method

Mmh, sure, but my remark was about the fact that `loop` is not documented in the reference book, where [`typeof`](https://crystal-lang.org/reference/syntax_and_semantics/typeof.html) is, for instance.

oprypin:
Yes, because `typeof` is a keyword 🤔 unlike `loop`

beta-ziliani:
I just realize that `loop` is in fact mentioned in the `while` section. I guess we just need to add a section to the tutorial with focus on iteration.

SamantazFox:
> Yes, because `typeof` is a keyword 🤔 unlike `loop`

`typeof` is a class method, apparently: https://crystal-lang.org/api/1.1.1/toplevel.html#typeof(*expression):Class-class-method

straight-shoota:
I think we can close this. `loop` is properly documented in the stdlib API docs.

PR #581, opened 2022-01-11
Add section on iteration

bkmgit:
Partially addressing https://github.com/crystal-lang/crystal-book/issues/542

### DIFF
--- a/docs/tutorials/basics/50_control_flow.md
+++ b/docs/tutorials/basics/50_control_flow.md
@@ -243,7 +243,7 @@ The first matching condition selects which branch executes.
 
 ## Iteration
 
-Crystal offers several ways to do iteration. If iterating over an object, a natural way is to use `each`:
+Crystal offers several ways to do iteration. If iterating over an _iterable_ object, like an array, a natural way is to use `each`:
 
 ```crystal-play
 vals = ["one", "two", "three", "four"]
--- a/docs/tutorials/basics/50_control_flow.md
+++ b/docs/tutorials/basics/50_control_flow.md
@@ -260,7 +260,7 @@ vals.each_with_index { |num, id| puts "#{num} is #{id}" }
 `upto` allows iterations in a range:
 
 ```crystal-play
-1.upto(5) { |i| puts "Crystal is awesom!" }
+1.upto(5) { |i| puts "Crystal is awesome #{i}!" }
 ```
 
 If one wants to increment the counter directly, `until` can be used:
--- a/docs/tutorials/basics/50_control_flow.md
+++ b/docs/tutorials/basics/50_control_flow.md
@@ -240,3 +240,33 @@ end
 
 Both clauses have branches with the same conditions but in a different order and they behave differently.
 The first matching condition selects which branch executes.
+
+## Iteration
+
+Crystal offers several ways to do iteration. If iterating over an object, a natural way is to use `each`:
+
+```crystal-play
+vals = ["one", "two", "three", "four"]
+vals.each { |num| puts num }
+```
+
+The `loop` construct can also be used:
+
+```crystal-play
+count = 0
+loop do
+  puts "Crystal is awesome!"
+  count = count + 1
+  break if count == 5
+end
+```
+
+Another way is to use `while`:
+
+```crystal-play
+count = 0
+while count < 6
+  puts "Crystal is awesome!"
+  count = count + 1
+end
+```
--- a/docs/tutorials/basics/50_control_flow.md
+++ b/docs/tutorials/basics/50_control_flow.md
@@ -243,7 +243,19 @@ The first matching condition selects which branch executes.
 
 ## Iteration
 
-Crystal offers several ways to do iteration. If iterating over an _iterable_ object, like an array, a natural way is to use `each`:
+Crystal offers several ways to do iteration. `upto` allows iterations in a range:
+
+```crystal-play
+1.upto(5) { |i| puts "Crystal is awesome #{i}!" }
+```
+
+or if the index is not required,
+
+```crystal-play
+1.upto(5) { puts "Crystal is awesome!" }
+```
+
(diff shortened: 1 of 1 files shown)


docs/tutorials/basics/50_control_flow.md
```diff
@@ -240,3 +240,33 @@ end
 
 Both clauses have branches with the same conditions but in a different order and they behave differently.
 The first matching condition selects which branch executes.
+
+## Iteration
+
+Crystal offers several ways to do iteration. If iterating over an object, a natural way is to use `each`:
```
beta-ziliani, line 246: Just to clarify that `each` is not part of every object.

```suggestion
Crystal offers several ways to do iteration. If iterating over an _iterable_ object, like an array, a natural way is to use `each`:
```

docs/tutorials/basics/50_control_flow.md
```diff
@@ -240,3 +240,33 @@ end
...
+
+```crystal-play
+vals = ["one", "two", "three", "four"]
+vals.each { |num| puts num }
+```
+
+The `loop` construct can also be used:
```
beta-ziliani, line 253: I'd start with `while` first (which is indeed a construct), and then with `loop`, which is a [function](https://crystal-lang.org/api/1.3.0/toplevel.html#loop%28%26%29-class-method). In case it isn't clear, to me a construct is like a primitive, and `loop` isn't.

beta-ziliani (commented):
Nice addition, thanks!

docs/tutorials/basics/50_control_flow.md
bkmgit: @beta-ziliani Thanks for the review. Reordered and add a few further examples.  Further suggestions are welcome.

docs/tutorials/basics/50_control_flow.md
```diff
@@ -240,3 +240,56 @@ end
...
+```
+
+It is also possible to get the index using `each_With_index`:
+
+```crystal-play
+vals = ["one", "two", "three", "four"]
+vals.each_with_index { |num, id| puts "#{num} is #{id}" }
+```
+
+`upto` allows iterations in a range:
```
beta-ziliani, line 260: What do you think of starting this part with an example of `(1..5).each` to introduce a range, and then this one? Just a thought.

beta-ziliani (commented):
Nice improvements! 👍

docs/tutorials/basics/50_control_flow.md
beta-ziliani: I was meaning to use range's `each` `(1..5).each` prior to `upto`, but somehow the reordering does flow nicely now, so 👍 👍

beta-ziliani (approved):
🙇  @bkmgit !

straight-shoota:
I'm terribly sorry to chip in late. I had meant to leave a comment earlier, but it got longer and I lost track.

Thank you very much for the initiative. The tutorial is missing quite a bit and the effort to enhance it, is very appreciated. However, this goes a bit off from what I had planned for the iteration topic.

The section doesn't come off as a tutorial in the way that it *explains* how these concepts work. It just showcases a number of code examples. That's totally fine and certainly a help, but I feel it doesn't catch the idea of this tutorial series.
A clear indicator for that is the short text between examples. Compare that to the first part of that file. A complex topic such as loops should clearly need more explanation.
Also, I'm confused why you start with a very specific iteration method (`upto`). I can't see any didactic purpose in that. The tutorial should start with the very basics, that is `while` and `until` and later introduce alternatives that are easier for specific use cases. At this point, the concept of blocks has not been introduced, so all other iteration methods introduce yet another concept without explanation.

So at this point, I think the tutorial should only cover the basic language features `while` and `until` and really explain how they work.
Other iteration methods can be covered later in an advanced section, after blocks and basic collection types are introduced.

straight-shoota:
Perhaps we could make a separate document listing all methods for iteration. Could be a guide, maybe?

> They can be good examples for explaining blocks in fact.

I think blocks should first be introduced with simple, non-iterative examples. Something like `try`, for example.

bkmgit:
Thanks for the suggestions. The explanations are nice, but have mostly used the tutorial as a quick reference, so the example code has usually been more helpful than the text. If the intention is to write something like [The C Programming Language](https://en.wikipedia.org/wiki/The_C_Programming_Language) for Crystal, then some choice in what to include is required. `upto` is so clean and concise starting with `while` or `until` is a disservice to people new to Crystal. `each` and `loop` can probably be put elsewhere. The [contents section](https://deploy-preview-581--crystal-book.netlify.app/tutorials/basics/index.html) indicates some information on loops is available in the control flow section. Can add a little more text for `upto`, `while` or `until` if that would be helpful.

beta-ziliani:
The main problem is the ordering of information. Up to this point, readers were not yet faced to understand what a `do ... end` code block does to a method call. So, they might get confused later on when they realized that's not part of iteration but a general concept on its own. In particular, it might come as a surprise that the `while` doesn't use `do`.

I suggest the following: stick to the keywords for iteration (`while`, `until`), and mention that there is no `for` for good reasons that will become obvious later on, after we explain `do ... end` blocks. The rest can go to an `Advance iteration` part in a coming file for blocks. WDYT?

straight-shoota:
I had already started on a loops section a couple months ago, but it's been stalled as a draft. I'll try to finish it up and make a PR for it.

bkmgit:
Ok. Will close this request since you already have a draft. Ruby in 20 minutes only introduces [do .. end](http://www.ruby-lang.org/en/documentation/quickstart/4/) and Learn to program only introduces [while](https://pine.fm/LearnToProgram/chap_06.html). It seems that loops are [heated topic](https://github.com/crystal-lang/crystal/issues/830) - but feel free to reopen if the draft will take a long time to complete.

PR #583, opened 2022-01-19
Add section about loops in basic tutorial

straight-shoota:
This section has been brewing for long and I finally managed to finish it up after a nudge in #581.

It only covers the basic building blocks `while`, `until`, `next` and `break`. The more convenient looping methods from stdlib will be introduced in a later lesson. The tutorial still misses a couple of basic concepts for that.

### DIFF
--- a/docs/tutorials/basics/50_control_flow.md
+++ b/docs/tutorials/basics/50_control_flow.md
@@ -243,7 +243,7 @@ The first matching condition selects which branch executes.
 
 ## Loops
 
-This section delves into writing code that it will repeat until a condition (aka the `loop condition`) is met
+This section introduces the basics of repeated execution of code.
 
 The basic feature is the `while` clause. Its structured quite similar to an `if` clause:
 The keyword `while` designates the beginning and is followed by an expression serving as the loop condition.
@@ -282,10 +282,10 @@ end
 
 ### Infinite loops
 
-When working with loops, it's important to care about the loop condition failing at some point.
+When working with loops, it's important to care about the loop condition being *falsey* at some point.
 Otherwise, it would continue forever or until you stop the program externally (for example <kbd>Ctrl+C</kbd>, `kill`, pull the plug or when armageddon arrives).
 
-In this example, not incrementing the counter it would be the same as writing: 
+In this example, not incrementing the counter it would be the same as writing:
 
 ```
 while true
@@ -299,8 +299,8 @@ A good practice for index variables (such as `counter` in our example) is to inc
 That makes it harder to forget that.
 
 !!! tip
-    Fortunately, there are many features in the language that relieve the burdon of writing loops manually
-    and also take care of ensuring valid break conditions. A few of them will be introduced in following lessons.
+    Fortunately, there are many features in the language that relieve the burden of writing loops manually
+    and also take care of ensuring valid breaking conditions. A few of them will be introduced in following lessons.
 
 In some cases, the intention is to really have an endless loop.
 An example would be a server that always repeats waiting for a connection, or
--- a/docs/tutorials/basics/50_control_flow.md
+++ b/docs/tutorials/basics/50_control_flow.md
@@ -246,7 +246,7 @@ The first matching condition selects which branch executes.
 This section delves into reducing repetition by letting code repeat itself in a loop.
 
 The basic feature is the `while` clause. Its structured quite similar to an `if` clause:
-The keyword `while` designates the begin and is followed by an expression serving as the loop condition.
+The keyword `while` designates the beginning and is followed by an expression serving as the loop condition.
 All subsequent expressions are part of the loop until the closing keyword `end`.
 The loop continues to repeat itself as long as the return value of the condition is *truthy*.
 
--- a/docs/tutorials/basics/50_control_flow.md
+++ b/docs/tutorials/basics/50_control_flow.md
@@ -240,3 +240,130 @@ end
 
 Both clauses have branches with the same conditions but in a different order and they behave differently.
 The first matching condition selects which branch executes.
+
+## Loops
+
+This section delves into reducing repetition by letting code repeat itself in a loop.
+
+The basic feature is the `while` clause. Its structured quite similar to an `if` clause:
+The keyword `while` designates the begin and is followed by an expression serving as the loop condition.
+All subsequent expressions are part of the loop until the closing keyword `end`.
+The loop continues to repeat itself as long as the return value of the condition is *truthy*.
+
+Let's try a simple program for counting from 1 to 10:
+
+```{.crystal .crystal-play}
+counter = 0
+
+while counter < 10
+  counter += 1
+
+  puts "Counter: #{counter}"
+end
+```
+
+The code between `while` and `end` is executed 10 times. It prints the current counter value and increases it by one.
+After the 10th iteration, the value of `counter` is `10`, thus `counter < 10` fails and the loop breaks.
+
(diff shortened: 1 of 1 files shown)


docs/tutorials/basics/50_control_flow.md
```diff
@@ -240,3 +240,130 @@ end
 
 Both clauses have branches with the same conditions but in a different order and they behave differently.
 The first matching condition selects which branch executes.
+
+## Loops
+
+This section delves into reducing repetition by letting code repeat itself in a loop.
+
+The basic feature is the `while` clause. Its structured quite similar to an `if` clause:
+The keyword `while` designates the begin and is followed by an expression serving as the loop condition.
```
Fryguy, line 249:
```suggestion
The keyword `while` designates the beginning and is followed by an expression serving as the loop condition.
```

docs/tutorials/basics/50_control_flow.md
```diff
@@ -240,3 +240,130 @@ end
...
+  puts "Counter: #{counter}"
+end
+```
+
+The code between `while` and `end` is executed 10 times. It prints the current counter value and increases it by one.
+After the 10th iteration, the value of `counter` is `10`, thus `counter < 10` fails and the loop breaks.
+
+An alternative is to replace `while` with the keyword `until` which expects just the opposite truthiness. `until x` is equivalent to `while !x`.
+
+```crystal-play
```
Fryguy, line 270:
```suggestion
```{.crystal .crystal-play}
```

docs/tutorials/basics/50_control_flow.md
```diff
@@ -240,3 +240,130 @@ end
...
+end
+```
+
+!!! tip
+    You can find more details on these expressions in the language specification: [`while`](../../syntax_and_semantics/while.md) and [`until`](../../syntax_and_semantics/until.md).
+
+### Infinite loops
+
+When working with loops, it's important to care about the loop condition failing at some point.
+Otherwise, it wouldn't stop ever and continue to all eternity – or until you stop the program externally (for example <kbd>Ctrl+C</kbd>, `kill`, pull the plug or when armageddon arrives).
```
Fryguy, line 286:
```suggestion
Otherwise, it would continue forever or until you stop the program externally (for example <kbd>Ctrl+C</kbd>, `kill`, pull the plug or when armageddon arrives).
```

docs/tutorials/basics/50_control_flow.md
```diff
@@ -240,3 +240,130 @@ end
 
 Both clauses have branches with the same conditions but in a different order and they behave differently.
 The first matching condition selects which branch executes.
+
+## Loops
+
+This section delves into reducing repetition by letting code repeat itself in a loop.
```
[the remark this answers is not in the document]
beta-ziliani: I would go here with something even simpler:
```suggestion
This section introduces the basics of repeated execution of code.
```

docs/tutorials/basics/50_control_flow.md
```diff
@@ -240,3 +240,130 @@ end
...
+
+The code between `while` and `end` is executed 10 times. It prints the current counter value and increases it by one.
+After the 10th iteration, the value of `counter` is `10`, thus `counter < 10` fails and the loop breaks.
+
+An alternative is to replace `while` with the keyword `until` which expects just the opposite truthiness. `until x` is equivalent to `while !x`.
+
+```{.crystal .crystal-play}
+counter = 0
+
+until counter > 9
```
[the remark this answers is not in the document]
straight-shoota:
```suggestion
until counter => 10
```

docs/tutorials/basics/50_control_flow.md
```diff
@@ -240,3 +240,130 @@ end
...
+```
+
+!!! note
+    This example is not an interactive playground by choice because the playground can't
+    handle non self-terminating programs, and processing user input.
+    It would just time out and print an error.
+    You can compile and run this code with a local compiler, though.
+
+    To stop the program, hit <kbd>Ctrl+C</kbd>. This sends a signal to the process asking it
+    to exit.
```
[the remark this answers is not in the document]
straight-shoota: That's not a problem. Line breaks are just rendered as space.

docs/tutorials/basics/50_control_flow.md
```diff
@@ -240,3 +240,130 @@ end
...
+
+This example could've easily be written without `next` by placing the `puts` expression in a conditional instead.
+The value of `next` becomes apparent when there are many more expressions in the method body to be skipped.
+
+Loop conditions can be difficult to calculate, for example because they require multiple steps or depend on input that needs to be determined.
+In such situations, it's not very practical to write all the logic in the loop condition.
+The keyword `break` can be used anywhere in a loop body and serves as an additional option to break from a loop regardless of its loop condition.
+Control flow immediately continues after the end of the loop.
+
+```{.crystal .crystal-play}
```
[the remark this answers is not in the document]
straight-shoota: I think it's better to use a program that can run in the online playground. This wouldn't work because it needs user input.

docs/tutorials/basics/50_control_flow.md
```diff
@@ -240,3 +240,136 @@ end
 
 Both clauses have branches with the same conditions but in a different order and they behave differently.
 The first matching condition selects which branch executes.
+
+## Loops
+
+This section delves into writing code that it will repeat until a condition (aka the `loop condition`) is met
```
beta-ziliani, line 246: This line doesn't work: what is _it_ in "it will repeat"? I repeat my suggestion:

```suggestion
This section introduces the basics of repeated execution of code.
```

docs/tutorials/basics/50_control_flow.md
```diff
@@ -240,3 +240,136 @@ end
...
+In this example, not incrementing the counter it would be the same as writing:
+
+```
+while true
+  puts "Counter: #{counter}"
+end
+```
+Or if the condition was `counter > 0`, it would match for all values: they only increase from `1`.
+This would not technically be infinite, as it will fail with a math error when the counter reaches the maximum value of a 32-bit integer. But conceptually that's similar to an infinite loop.
+Such logic errors can be easy to miss and so it's very important to pay attention when writing the loop condition and also taking care of meeting said breaking case
```
beta-ziliani, line 297:
```suggestion
Such logic errors can be easy to miss and so it's very important to pay attention when writing the loop condition and also taking care of meeting said breaking case.
```

docs/tutorials/basics/50_control_flow.md
```diff
@@ -240,3 +240,136 @@ end
...
+A good practice for index variables (such as `counter` in our example) is to increment them at the beginning of the loop.
+That makes it harder to forget that.
```
beta-ziliani, line 299:
```suggestion
That makes it harder to forget to update them.
```

docs/tutorials/basics/50_control_flow.md
```diff
@@ -240,3 +240,136 @@ end
...
+  if counter % 3 == 0
+    next
+  end
+
+  puts "Counter: #{counter}"
+end
+```
+
+This example could've easily be written without `next` by placing the `puts` expression in a conditional instead.
+The value of `next` becomes apparent when there are many more expressions in the method body to be skipped.
```
beta-ziliani, line 354: "value" is used as the value of an expression, so it can get a bit confusing.
```suggestion
The worth of `next` becomes apparent when there are many more expressions in the method body to be skipped.
```

beta-ziliani (approved):
Just nitpicking. After fixing these it's good to go.

PR #584, opened 2022-01-27
Automatically populate crystal-version.txt

oprypin:
Note:
* The original file is kept even though it won't be manually updated anymore.
* The `crystal` invocation is kept separate from the build process.

This ensures that 'crystal-book' doesn't have 'crystal' as a hard build dependency.

---

Ref https://github.com/crystal-lang/crystal-book/pull/578#issuecomment-1007216153

Made possible by https://github.com/crystal-lang/crystal-book/pull/562

### DIFF
--- a/.github/workflows/deploy-book.yml
+++ b/.github/workflows/deploy-book.yml
@@ -40,6 +40,8 @@ jobs:
           if [[ '${{ steps.branch.outputs.branch }}' =~ ^[0-9] ]]; then
             echo "CRYSTAL_VERSION=${{ steps.crystal.outputs.crystal }}" >> ${GITHUB_ENV}
           fi
+      - name: Populate sample version of Crystal
+        run: crystal --version | tee crystal-version.txt
       - name: Build book
         run: LINT=true make build
       - name: Build versions file


straight-shoota:
I was actually planning to update this file in the automated release update workflow. That would remove the odd delta between source file and the online version.
Updates only need to happen when a Crystal version is released, not on every build. This might make sense for the master version though. Not that we really need a completely up-to-date version text, but we get a `-dev` version instead of the latest stable release which would be checked in. And we wouldn't want to update the source file in `master` to avoid merge conflicts.

oprypin:
> I was actually planning to update this file in the automated release update workflow

Hmm that makes it slightly more annoying to manage several release branches (in terms of the manual decision which change goes to a branch and which to master).
And it's manual, which is a huge downside.

> That would remove the odd delta between source file and the online version.

Does this downside have any real significance whatsoever?

> but we get a -dev version instead of the latest stable release which would be checked in.

On release branches it will be the release version.

> we wouldn't want to update the source file in master to avoid merge conflicts.

Yes so let's just not update anything :)

straight-shoota:
> that makes it slightly more annoying to manage several release branches

How so? I would imagine the file gets only changed when a new release branch is created. There shouldn't be any issues from that.
And the idea is to have this automated, and maybe eventually fully automatic.

straight-shoota (approved):
I'm definitely fine with merging this, though.

oprypin:
Everything went as planned.

---


https://crystal-lang.org/reference/1.3/getting_started/

<details><summary>Used to be:</summary>

```console
$ crystal --version
Crystal 1.3.0 [a3ee70ca0] (2022-01-06)

LLVM: 10.0.0
Default target: x86_64-unknown-linux-gnu
```

</details>


Became:
```console
$ crystal --version
Crystal 1.3.2 [932f193ae] (2022-01-18)

LLVM: 10.0.0
Default target: x86_64-unknown-linux-gnu
```

---

https://crystal-lang.org/reference/master/getting_started/

Became:
```console
$ crystal --version
Crystal 1.4.0-dev [b69e71f44] (2022-01-30)

LLVM: 10.0.0
Default target: x86_64-unknown-linux-gnu
```

PR #601, opened 2022-04-06
Mention user:password for PostgreSQL

rdp:
Mention postgres can take a user/pass, though ideally wouldn't this just be described in the individual modules?


docs/database/README.md
```diff
@@ -7,7 +7,8 @@ The following packages are compliant with crystal-db
 * [crystal-lang/crystal-sqlite3](https://github.com/crystal-lang/crystal-sqlite3) for sqlite
 * [crystal-lang/crystal-mysql](https://github.com/crystal-lang/crystal-mysql) for mysql & mariadb
 * [will/crystal-pg](https://github.com/will/crystal-pg) for postgres
-
+And several others.
```
straight-shoota, line 10: Not sure if this is much help without referencing "others".

I think we should have a canonical location for listing compatible drivers.

rdp: OK added a link in c845507 thanks.  It's kind of weird that the docs have the big 3 but kind of like incomplete versions...that can be fixed in a later commit though.

straight-shoota (approved):

oprypin (approved):

PR #602, opened 2022-04-06
Update DB Readme

rdp:
Add example of query parameters in a query, mention there are more helper methods available for query, mention preparedstatements are at play (crystal-lang/crystal-db#164).

### DIFF
--- a/docs/database/README.md
+++ b/docs/database/README.md
@@ -76,6 +76,8 @@ db.exec "insert into contacts values (?, ?)", "John", 30
 # Postgres
 db.exec "insert into contacts values ($1, $2)", "Sarah", 33
 ```
+Query parameters are accomplished typically using PreparedStatements under the hood,
+though this can be disabled via a connection string option `prepared_statements=false`.
 
 ## Query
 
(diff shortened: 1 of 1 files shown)


rdp:
Note to self: after this is committed add an example of creating a PreparedStatement...

docs/database/README.md
```diff
@@ -76,6 +76,8 @@ db.exec "insert into contacts values (?, ?)", "John", 30
 # Postgres
 db.exec "insert into contacts values ($1, $2)", "Sarah", 33
 ```
+Query parameters are accomplished typically using PreparedStatements under the hood,
+though this can be disabled via a connection string option `prepared_statements=false`.
```
straight-shoota, line 80: Not sure if that statement is correct. I don't think query parameters are technically tied to prepared statements.

docs/database/README.md
```diff
@@ -91,6 +93,16 @@ db.query "select name, age from contacts order by age desc" do |rs|
 end
 ```
 
+You can also use query parameters:
```
straight-shoota, line 96: Query parameters are already mentioned in more depth a couple lines above. It would probably make sense to move the explanation of query parameters to a separate section and reference that from both `exec` and `query` sections.

rdp: OK attempted see a147195

docs/database/README.md
rdp: Also in a147195

straight-shoota (approved):

oprypin (approved):

PR #603, opened 2022-04-10
Add Generics with variable number of arguments

ftarulla:
After adding https://github.com/crystal-lang/crystal/pull/11906 to Crystal 1.4, it's a good time to add definitions and examples for "using splats in generics" to the Generics section.
The last example shows the _Support Generic nodes with no type variables_

### DIFF
--- a/docs/syntax_and_semantics/generics.md
+++ b/docs/syntax_and_semantics/generics.md
@@ -96,7 +96,7 @@ end
 
 We may define a Generic class with a variable number of arguments using the [splat operator](./operators.md#splats).
 
-Let's see an example where we define a Generic class called `Foo` and then we will use it with two type parameters `Int32` and `String`:
+Let's see an example where we define a Generic class called `Foo` and then we will use it with different number of type variables:
 
 ```crystal-play
 class Foo(*T)
@@ -106,10 +106,17 @@ class Foo(*T)
   end
 end
 
+# 2 type variables:
+# (explicitly specifying type variables)
 foo = Foo(Int32, String).new(42, "Life, the Universe, and Everything")
 
 puts typeof(foo) # => Foo(Int32, String)
 puts foo.content # => {42, "Life, the Universe, and Everything"}
+
+# 3 type variables:
+# (type variables inferred by the compiler)
+bar = Foo.new("Hello", ["Crystal", "!"], 140)
+puts typeof(bar) # => Foo(String, Array(String), Int32)
 ```
 
 In the following example we define classes by inheritance, specifying instances for the generic types:
@@ -146,14 +153,7 @@ class Parent(*T)
 end
 
 foo = Parent.new # => Error: can't infer the type parameter T for the generic class Parent(*T). Please provide it explicitly
-```
 
-```crystal
-class Parent(*T)
+class Foo < Parent # => Error: generic type arguments must be specified when inheriting Parent(*T)
 end
-
-class Foo < Parent
-end
-
-# => Error: generic type arguments must be specified when inheriting Parent(*T)
 ```
--- a/docs/syntax_and_semantics/generics.md
+++ b/docs/syntax_and_semantics/generics.md
@@ -101,6 +101,7 @@ When using Generics this is also possible, meaning a `class` can be defined with
 ```crystal
 class Foo(*T)
   getter content
+
   def initialize(*@content : *T)
   end
 end
--- a/docs/syntax_and_semantics/generics.md
+++ b/docs/syntax_and_semantics/generics.md
@@ -91,3 +91,43 @@ end
 class GenericChild(T) < Parent(T)
 end
 ```
+
+## Generics with variable number of arguments
+
+As we can read in [Splats and Tuples](./splats_and_tuples.md) a method _can receive a variable number of arguments_.
+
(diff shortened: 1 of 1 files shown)


docs/syntax_and_semantics/generics.md
```diff
@@ -91,3 +91,44 @@ end
...
+```
+
+And if we need to instantiate a `class` with 0 arguments? In that case we may do:
+
+```crystal
+class Parent(*T)
+end
+
+class NoArgumentChild < Parent()
+end
```
straight-shoota, line 133: The sub class is not really necessary to demonstrate this. You could just directly instantiate with empty generic arguments (`Parent().new`). And contrast that to an instantiation without generic arguments (`Parent.new`).

docs/syntax_and_semantics/generics.md
```diff
@@ -91,3 +91,69 @@ end
...
+
+```crystal-play
+class Foo(*T)
+  getter content
+
+  def initialize(*@content : *T)
+  end
+end
+
+foo = Foo(Int32, String).new(42, "Life, the Universe, and Everything")
```
straight-shoota, line 109: Can we also add examples with inferrred argument types, and a different arity of generic arguments? (could be in a single example)

docs/syntax_and_semantics/generics.md
```diff
@@ -91,3 +91,69 @@ end
...
+
+```crystal
+class Parent(*T)
+end
+
+class Foo < Parent
+end
+
+# => Error: generic type arguments must be specified when inheriting Parent(*T)
+```
```
straight-shoota, line 159: These examples show code that doesn't compile. So I think they don't need to be self-contained. This can be more  concise:
```suggestion
```crystal
Parent.new # Error: can't infer the type parameter T for the generic class Parent(*T). Please provide it explicitly
class Foo < Parent # Error: generic type arguments must be specified when inheriting Parent(*T)
end
```

docs/syntax_and_semantics/generics.md
```diff
@@ -91,3 +91,69 @@ end
...
+# We define `StringChild` inheriting from `Parent` class
+# using `String` for generic type argument:
+class StringChild < Parent(String)
+end
+
+# We define `Int32StringChild` inheriting from `Parent` class
+# using `Int32` and `String` for generic type arguments:
+class Int32StringChild < Parent(Int32, String)
+end
+```
```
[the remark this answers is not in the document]
ftarulla: I think it's good to have the `Generic types inheritance` basic example and then this one (based on the other) that shows `classes` using different number of type variables.

docs/syntax_and_semantics/generics.md
```diff
@@ -91,3 +91,69 @@ end
...
+
+  def initialize(*@content : *T)
+  end
+end
+
+# 2 type variables:
+# (explicitly specifying type variables)
+foo = Foo(Int32, String).new(42, "Life, the Universe, and Everything")
+
+puts typeof(foo) # => Foo(Int32, String)
```
straight-shoota, line 113: Technically, this is not correct because `# =>` indicates to represent the value of the expression in this line. The return value of `puts` would be `nil`. Usually, we don't use `puts` or any other method that actually prints the value in these examples. It's not really necessary because the annotation in the example code is already sufficient to demonstrate this.
If we want the code to actually execute and print the value, it should be `p` instead of `puts`. `p` uses `#inspect` instead of `#to_s`, which should always be used for these kind of inspection purposes and is implicitly assumed by the `# =>` notation. And `p` actually returns the the value of its argument, so the comment annotation is even valid in a strict sense.

docs/syntax_and_semantics/generics.md
```diff
@@ -91,3 +91,69 @@ end
...
+
+But we should not mistake 0 arguments with not specifying the generic type variables. The following examples will raise an error:
+
+```crystal
+class Parent(*T)
+end
+
+foo = Parent.new # => Error: can't infer the type parameter T for the generic class Parent(*T). Please provide it explicitly
+
+class Foo < Parent # => Error: generic type arguments must be specified when inheriting Parent(*T)
```
straight-shoota, line 157: Per convention, `# =>` indicates a value and compiler errrors are prefixed by `# Error:`.
```suggestion
foo = Parent.new # Error: can't infer the type parameter T for the generic class Parent(*T). Please provide it explicitly

class Foo < Parent # Error: generic type arguments must be specified when inheriting Parent(*T)
```

straight-shoota (commented):
I have two issues about minor details in the sample code, but except for that this looks good to me.

docs/syntax_and_semantics/generics.md
ftarulla: oh! Great suggestion!! Now I realize that I've been using `# =>` in an incorrect way 🙃 Thx!!

ftarulla: So, I'm removing `puts` from every example but the ones in a `crystal-play` block. So that they can run the examples but also have the expected result in the code itself. For these examples I'm using `p` instead of `puts` so that it's technically correct (as you mentioned in the suggestion)

straight-shoota (approved):

beta-ziliani (approved):
Nice!

PR #604, opened 2022-04-10
Fix removing clipboard button when creating crystal-play component

ftarulla:
When replacing a code component with the `crystal-play` component, the following error was thrown:
```
crystal-play.js:11 Uncaught TypeError: Failed to execute 'removeChild' on 'Node'
```
This error was stopping the execution making the others (if any) code snippets to not be replaced with the `crystal-play` component.

In this PR a typo in the query selector was fixed and we check if the clipboard button exists before removing it.


straight-shoota (approved):

straight-shoota:
I'm wondering how this wasn't noticed before... it clearly causes broken functionality on the page (then there are multiple code blocks).

PR #606, opened 2022-04-11
Add Return section to Tutorial > Methods

ftarulla:
All the examples in `Tutorial>Methods` show methods that don't return a value.
We added new examples of methods returning values and also, examples with return type information

### DIFF
--- a/docs/tutorials/basics/60_methods.md
+++ b/docs/tutorials/basics/60_methods.md
@@ -108,27 +108,28 @@ say_hello 3
 Overloading isn't defined just by type restrictions. The number of arguments as well as named arguments are also
 relevant characteristics.
 
-## Return value
+## Returning a value
 
 Methods may also return a value:
 
 ```crystal-play
-def adds_2(n : Int32 = 0)
+def adds_2(n : Int32)
   n + 2
 end
 
 puts adds_2 40
 ```
 
-The keyword `return` is not necessary and methods will always return the result of executing the last line.
+!!! note
+    The keyword `return` is not necessary and methods will always return the result of executing the last line.
 
-The following example usest the `return` keyword and _implicit_ `return`:
+The following example illustrates the use of an _explicit_ and an _implicit_ `return`:
 
 ```crystal-play
 # This method returns:
 # - the same number if it's even,
 # - the number multiplied by 2 if it's odd.
-def build_even_number(n : Int32 = 0)
+def build_even_number(n : Int32)
   return n if n.even?
 
   times_2(n)
@@ -164,3 +165,31 @@ end
 
 hello_message_for "Crystal" # => Error: method top-level hello_message_for must return String but it is returning Int32
 ```
+
+Let's try one more example, just to highlight how useful it is to provide type information:
+
+```crystal-play
+def hello_message_for(recipient : String)
+  "Hello #{recipient}!"
+  42
+end
+
+puts hello_message_for "Crystal"
+```
+
+The above example works! (although we were expecting to print `Hello Crystal` instead of `42`).
+
+Now let's see what happens when we try to use the returned value as a `String`:
+
+```crystal
+def hello_message_for(recipient : String)
+  "Hello #{recipient}!"
+  42
+end
+
+(hello_message_for "Crystal") + "!!" # => Error: no overload matches 'Int32#+' with type String
+```
+
+The compiler is telling us that there is no `Int32#+` method that takes a `String` value as an argument.
+
+As we can see, this last Error message is not as accurate as when using type information since the compiler cannot know that we intend the method `hello_message_for` to return a String.
--- a/docs/tutorials/basics/60_methods.md
+++ b/docs/tutorials/basics/60_methods.md
@@ -121,7 +121,7 @@ puts adds_2 40
 ```
 
 !!! note
-    The keyword `return` is not necessary and methods will always return the result of executing the last line.
+    In the previous example, the use of the keyword `return` in method `adds_2` would be redundant since methods always return the evaluation of the last line.
 
 The following example illustrates the use of an _explicit_ and an _implicit_ `return`:
 
@@ -192,4 +192,4 @@ end
 
 The compiler is telling us that there is no `Int32#+` method that takes a `String` value as an argument.
 
-As we can see, this last Error message is not as accurate as when using type information since the compiler cannot know that we intend the method `hello_message_for` to return a String.
+As we can see, this last error message is not as accurate as when using type information since the compiler cannot know that we intend the method `hello_message_for` to return a `String`.
--- a/docs/tutorials/basics/60_methods.md
+++ b/docs/tutorials/basics/60_methods.md
@@ -117,7 +117,7 @@ def adds_2(n : Int32 = 0)
   n + 2
 end
 
-puts adds_2 40 # => 42
+puts adds_2 40  # => 42
 puts adds_2(40) # => 42 syntactically equivalent method call with parentheses
 ```
 
@@ -139,7 +139,7 @@ def times_2(n : Int32) : Int32
   n * 2
 end
 
-puts build_even_number 7 # => 14
+puts build_even_number 7  # => 14
 puts build_even_number 28 # => 28
 ```
 
--- a/docs/tutorials/basics/60_methods.md
+++ b/docs/tutorials/basics/60_methods.md
@@ -107,3 +107,61 @@ say_hello 3
 
 Overloading isn't defined just by type restrictions. The number of arguments as well as named arguments are also
 relevant characteristics.
+
+## Return value
+
+Methods may also return a value:
+
+```crystal-play
+def adds_2(n : Int32 = 0)
+  n + 2
+end
+
+puts adds_2 40 # => 42
+puts adds_2(40) # => 42 syntactically equivalent method call with parentheses
+```
+
+The keyword `return` is not necessary and methods will always return the result of executing the last line.
+
+The following example usest the `return` keyword and _implicit_ `return`:
+
+```crystal-play
+# This method returns:
+# - the same number if it's even,
+# - the number multiplied by 2 if it's odd.
+def build_even_number(n : Int32 = 0)
+  return n if n.even?
+
+  times_2(n)
+end
+
+def times_2(n : Int32) : Int32
+  n * 2
+end
+
+puts build_even_number 7 # => 14
+puts build_even_number 28 # => 28
+```
+
+### Return type
+
+As we can see in the previous example, in `times_2` definition we provide type information for the returned value. Here is another example:
+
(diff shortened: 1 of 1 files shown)


docs/tutorials/basics/60_methods.md
```diff
@@ -107,3 +107,60 @@ say_hello 3
...
+
+```crystal-play
+def adds_2(n : Int32 = 0)
+  n + 2
+end
+
+puts adds_2 40
+```
+
+The keyword `return` is not necessary and methods will always return the result of executing the last line.
```
beta-ziliani, line 123:
```suggestion
Note how there's no `return` keyword: methods always return the result of executing the last line.
```

docs/tutorials/basics/60_methods.md
```diff
@@ -107,3 +107,60 @@ say_hello 3
 
 Overloading isn't defined just by type restrictions. The number of arguments as well as named arguments are also
 relevant characteristics.
+
+## Return value
+
+Methods may also return a value:
+
+```crystal-play
+def adds_2(n : Int32 = 0)
```
beta-ziliani, line 116: What's the point of the method having a default argument?

docs/tutorials/basics/60_methods.md
```diff
@@ -107,3 +107,60 @@ say_hello 3
...
+  n + 2
+end
+
+puts adds_2 40
+```
+
+The keyword `return` is not necessary and methods will always return the result of executing the last line.
+
+The following example usest the `return` keyword and _implicit_ `return`:
```
beta-ziliani, line 125:
```suggestion
The following example illustrates the use of an _explicit_ and an _implicit_ `return`:
```
Even better, join it to the previous paragraph, with something like "An _explicit_ return allows us to break earlier from the function. The following..."

docs/tutorials/basics/60_methods.md
```diff
@@ -107,3 +107,60 @@ say_hello 3
...
+
+```crystal-play
+# This method returns:
+# - the same number if it's even,
+# - the number multiplied by 2 if it's odd.
+def build_even_number(n : Int32 = 0)
```
beta-ziliani, line 131: Same as before, what's the use of the default arg here?

docs/tutorials/basics/60_methods.md
```diff
@@ -107,3 +107,60 @@ say_hello 3
...
+This is really useful for finding errors at compile time:
+
+```crystal
+def hello_message_for(recipient : String) : String
+  "Hello #{recipient}!"
+  42
+end
+
+hello_message_for "Crystal" # => Error: method top-level hello_message_for must return String but it is returning Int32
+```
```
beta-ziliani, line 166: Perhaps we can show what happens without type annotations: it works (prints 42), but if then we try to use it as a String we get an obscure error message:

```crystal
def hello_message_for(recipient : String)
  "Hello #{recipient}"
  42
end
puts (hello_message_for "Crystal") + "!" # => Error: no overload matches 'Int32#+' with type String
```

docs/tutorials/basics/60_methods.md
```diff
@@ -107,3 +107,60 @@ say_hello 3
 
 Overloading isn't defined just by type restrictions. The number of arguments as well as named arguments are also
 relevant characteristics.
+
+## Return value
```
beta-ziliani, line 111:
```suggestion
## Returning values
```
or
```suggestion
## Returning a value
```

beta-ziliani (commented):
I left several comments, some of which are somewhat stylistic in nature. Feel free to disagree and debate!

docs/tutorials/basics/60_methods.md
oprypin: I preferred the previous wording. Or at least the suggested wording can be misleading because people may see it as just "there's no `return` keyword" (whatsoever in the language)

ftarulla: That's right! I didn't want the reader to misread it as "the language don't have `return` keyword". That's why I went with _The keyword `return` is not necessary_

docs/tutorials/basics/60_methods.md
ftarulla: I'm pretty sure that is a leftover of another version of the example 🙈

docs/tutorials/basics/60_methods.md
ftarulla: I add a "Note block" to highlight the use of `return`.

Regarding the content maybe we can say something like:
_In the previous example, the use of the keyword `return` is redundant since methods always return the evaluation of the last line._

docs/tutorials/basics/60_methods.md
```diff
@@ -107,3 +107,89 @@ say_hello 3
...
+  "Hello #{recipient}!"
+  42
+end
+
+(hello_message_for "Crystal") + "!!" # => Error: no overload matches 'Int32#+' with type String
+```
+
+The compiler is telling us that there is no `Int32#+` method that takes a `String` value as an argument.
+
+As we can see, this last Error message is not as accurate as when using type information since the compiler cannot know that we intend the method `hello_message_for` to return a String.
```
beta-ziliani, line 195:
```suggestion
As we can see, this last error message is not as accurate as when using type information since the compiler cannot know that we intend the method `hello_message_for` to return a String.
```

beta-ziliani (approved):
Just a minor typo :+1:

beta-ziliani (approved):

docs/tutorials/basics/60_methods.md
```diff
@@ -107,3 +107,89 @@ say_hello 3
...
+  n * 2
+end
+
+puts build_even_number 7
+puts build_even_number 28
+```
+
+### Return type
+
+As we can see in the previous example, in `times_2` definition we provide type information for the returned value. Here is another example:
```
straight-shoota, line 148: Please don't introduce concepts into examples through the back door. What's not yet introduced, shouldn't be used. There is really no necessity for another example using a return type restriction before this.

docs/tutorials/basics/60_methods.md
```diff
@@ -107,3 +107,89 @@ say_hello 3
...
+
+```crystal-play
+def hello_message_for(recipient : String) : String
+  "Hello #{recipient}!"
+end
+
+puts hello_message_for "Crystal"
```
straight-shoota, line 155: This is a minor suggestion, but I don't think the parameter is really helpful for this example. We could just drop it and it works the same way:
```suggestion
def hello_message : String
  "Hello Crystal"
end

puts hello_message
```

docs/tutorials/basics/60_methods.md
```diff
@@ -107,3 +107,89 @@ say_hello 3
...
+
+This is really useful for finding errors at compile time:
+
+```crystal
+def hello_message_for(recipient : String) : String
+  "Hello #{recipient}!"
+  42
+end
+
+hello_message_for "Crystal" # => Error: method top-level hello_message_for must return String but it is returning Int32
```
straight-shoota, line 166: Ditto. This example even becomes much more concise by dropping the entire string literal.
```suggestion
def hello_message : String
  42
end

hello_message_for "Crystal" # Error: method top-level hello_message must return String but it is returning Int32
```

docs/tutorials/basics/60_methods.md
```diff
@@ -107,3 +107,89 @@ say_hello 3
...
+```crystal-play
+def hello_message_for(recipient : String)
+  "Hello #{recipient}!"
+  42
+end
+
+puts hello_message_for "Crystal"
+```
+
+The above example works! (although we were expecting to print `Hello Crystal` instead of `42`).
```
straight-shoota, line 180: IMO this example doesn't add anything. The next one drives the same point and is much more impressive.
````suggestion
````

File: docs/tutorials/basics/60_methods.md [hunk 12]
```diff
@@ -107,3 +107,89 @@ say_hello 3
...
+  "Hello #{recipient}!"
+  42
+end
+
+(hello_message_for "Crystal") + "!!" # => Error: no overload matches 'Int32#+' with type String
+```
+
+The compiler is telling us that there is no `Int32#+` method that takes a `String` value as an argument.
+
+As we can see, this last error message is not as accurate as when using type information since the compiler cannot know that we intend the method `hello_message_for` to return a `String`.
```
straight-shoota, line 195: I would move this failing example up as an introduction into the *Return type* section. Didactically, it's better to experience the *problem* first and derive a *solution* from that instead of going the other way around.

However, the concrete example should be improved. The made-up usage example isn't very comprehensible.
Maybe we could turn the types around, make the example call a numeric operation and have the method return `String`:
```cr
def lucky_number
  "slevin"
end

puts lucky_number + 3 # Error: no overload matches 'String#+' with type Int32
```

docs/tutorials/basics/60_methods.md
ftarulla: Sometimes I like to present some elements (without explaining them) to capture reader's attention (and in code examples, it's also a way to show how intuitive the language is)
But sometimes I understand it may not work, so I accept your suggestion 😃

docs/tutorials/basics/60_methods.md
ftarulla: ok, this example makes sense in contrast to the example that follows. But if we're moving the other example to the beginning of the section (to introduce the problem), then this example doesn't make sense on its own 👍

straight-shoota (approved):

beta-ziliani (approved):

docs/tutorials/basics/60_methods.md
```diff
@@ -107,3 +107,59 @@ say_hello 3
...
+
+puts build_even_number 7
+puts build_even_number 28
+```
+
+### Return type
+
+Let's begin defining a method that we expect it will return an `Int32` value, but mistakenly returns a `String`:
+
+```crystal
```
beta-ziliani, line 145:
```suggestion
```crystal-play
```

docs/tutorials/basics/60_methods.md
```diff
@@ -107,3 +107,59 @@ say_hello 3
...
+end
+
+puts life_universe_and_everything + 1 # Error: no overload matches 'String#+' with type Int32
+```
+
+Because we never told the compiler we were expecting the method to return an `Int32`, the best the compiler can do is to tell us that there is no `String#+` method that takes an `Int32` value as an argument (i.e. the compiler is pointing at the moment when we use the value but not at the root of the bug: the type of the method's return value).
+
+The error message can be more accurate if using type information, so let's try again the example but now specifying the type:
+
+```crystal
```
beta-ziliani, line 157:
```suggestion
```crystal-play
```

beta-ziliani (commented):
These two should be in the playground too, don't they?

PR #609, opened 2022-04-28
Add WASM to platform support list

straight-shoota:
Basic WebAssembly support has been added in https://github.com/crystal-lang/crystal/pull/10870 and was released in Crystal 1.4

The second commit applies uniform format for the libc description.


beta-ziliani (approved):
Shouldn't this land on `master`?

straight-shoota:
WASM support was released in 1.4, so I think it makes sense to add it there. We could also merge into master and backport to `release/1.4` (or not do that). IdK. We don't have any established workflows for this, so let's figure them out =)

PR #616, opened 2022-05-31
Add note about type coercion of instance variables

aarongodin:
This has tripped me up enough times that I assume others are experiencing the same issue, so I wanted to add more information to the page on type restriction through `if var.nil?`.


straight-shoota (approved):

beta-ziliani (approved):
🚀  Thanks!

PR #619, opened 2022-07-05
Adding mention that the parameters name of abstract defs must match

beta-ziliani:


docs/syntax_and_semantics/virtual_and_abstract_types.md
```diff
@@ -95,6 +95,6 @@ abstract class Animal
 end
 ```
 
-By marking a method as `abstract` the compiler will check that all subclasses implement this method, even if a program doesn't use them.
+By marking a method as `abstract` the compiler will check that all subclasses implement this method (matching the arguments types and names), even if a program doesn't use them.
```
straight-shoota, line 98:
```suggestion
By marking a method as `abstract` the compiler will check that all subclasses implement this method (matching the parameter types and names), even if a program doesn't use them.
```

oprypin (approved):

PR #623, opened 2022-07-28
Update `crystal-version.txt`

straight-shoota:
Also adds `scripts/release-update.sh` like in other repositories for easy automation.

### DIFF
--- a/crystal-version.txt
+++ b/crystal-version.txt
@@ -1,4 +1,4 @@
-Crystal 1.4.1 [b7377c041] (2022-04-22)
+Crystal 1.5.0 (2022-07-06)
 
-LLVM: 13.0.0
+LLVM: 14.0.6
 Default target: x86_64-unknown-linux-gnu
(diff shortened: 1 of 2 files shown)


beta-ziliani (approved):

oprypin:
😢
Please don't do this.
This file is never used for deployment. It doesn't need to be touched.
Why are you adding an unnecessary workflow which has no advantages and only disadvantages?
I have explained this before.
https://github.com/crystal-lang/crystal-book/pull/584#issuecomment-1023783299

straight-shoota:
Even if it's overwritten on deployment, this file is still part of the content of this repository. I would not want to let an old version version output sit there and rot. Perhaps an alternative to updating it would be to just truncate the file and have it be populated only on deployment.

But I think it makes more sense to keep an up-to-date version of this file checked in for consistency and forward compatibility. Taking the version information directly from running the crystal command at the respective version at deployment only works as long as that crystal version is still runnable in the CI environment. Supporting legacy versions might become a liability in the future.

We're working on automating more parts of the release workflow making it require no extra work.
And there should not be any friction caused by updating this file. We can do that before splitting of a new release branch.
And even at merging, we can easily pin the master branch version to avoid conflicts.

Starr79 (commented):
Duplicate of #

PR #625, opened 2022-08-02
Restructuring the docs

beta-ziliani:
*Note:* we need to merge it also in 1.5 to make it effective


docs/SUMMARY.md
```diff
@@ -125,19 +117,25 @@
...
     * [Concurrency](guides/concurrency.md)
     * [Testing](guides/testing.md)
+    * [The shards command](the_shards_command/README.md)
     * [Writing Shards](guides/writing_shards.md)
         * [Hosting on GitHub](guides/hosting/github.md)
         * [Hosting on GitLab](guides/hosting/gitlab.md)
     * [Continuous Integration](guides/ci/README.md)
         * [GitHub Actions](guides/ci/gh-actions.md)
         * [CircleCI](guides/ci/circleci.md)
+    * [Using the compiler](using_the_compiler/README.md)
```
straight-shoota, line 127: Is there any particular reasoning to these insert positions? It seems very random.
Not that the order of existing items is very sophisticated, but when adding stuff we should at least try to make it fit in a sensible way.
The shards manual should  come after the compiler, for example. And we could consider to place them in a new sub section called "Manuals".

docs/SUMMARY.md
```diff
@@ -125,19 +117,25 @@
...
     * [Static Linking](guides/static_linking.md)
     * [Database](database/README.md)
         * [Connection](database/connection.md)
         * [Connection pool](database/connection_pool.md)
         * [Transactions](database/transactions.md)
+    * [Coding style](conventions/coding_style.md)
 * Tutorials
```
straight-shoota, line 134: Next we should add an overview page for *Tutorials*. This was previously skipped because there was only one. With classifying the getting started guide as a tutorial, we have two now.

docs/SUMMARY.md
```diff
@@ -1,13 +1,6 @@
 # Summary
 
 * [Welcome](README.md)
-    * [Getting started](getting_started/README.md)
-        * [An HTTP Server](getting_started/http_server.md)
-        * [A Command Line Application](getting_started/cli.md)
-    * [Using the compiler](using_the_compiler/README.md)
-    * [The shards command](the_shards_command/README.md)
-    * [Coding style](conventions/coding_style.md)
-    * [Platform Support](platform_support.md)
```
oprypin, line 10: > The following pages exist in the docs directory, but are not included in the "nav" configuration:
>   - platform_support.md

mkdocs.yml
```diff
@@ -53,6 +53,8 @@ plugins:
         conventions/documenting_code.md: syntax_and_semantics/documenting_code.md
         overview/index.md: syntax_and_semantics/README.md
         overview/hello_world.md: getting_started/README.md
+        using_the_compiler/README.md: manuals/using_the_compiler/README.md
+        the_shards_command/README.md: manuals/the_shards_command/README.md
```
straight-shoota, line 57: When we're already moving those paths, I'd suggest to make them more concise: `manuals/crystal/` and `manuals/shards/`.

Or even `/man/` prefix?

docs/SUMMARY.md
```diff
@@ -135,11 +128,18 @@
         * [Connection](database/connection.md)
         * [Connection pool](database/connection_pool.md)
         * [Transactions](database/transactions.md)
-* Tutorials
-    * [Basics](tutorials/basics/README.md)
+    * [Coding style](conventions/coding_style.md)
+* [Tutorials](tutorials/README.md)
+    * [Getting started](getting_started/README.md)
+        * [An HTTP Server](getting_started/http_server.md)
+        * [A Command Line Application](getting_started/cli.md)
```
straight-shoota, line 135: I'm also not convinced that *getting started* properly classifies as a tutorial.

beta-ziliani: How would you call it? A _guide_? To me _tutorial_ fits better. But the distinction isn't that clear...

docs/README.md
```diff
@@ -27,16 +27,22 @@ Install Crystal and get it running.
   <div class="card" markdown="1">
 
 ### Tutorials
-Introductory courses for beginners and advanced learners.
+Introductory material for beginners.
 
-* [Basics](tutorials/basics/README.md)
+* [Language introduction](tutorials/basics/README.md)
+* [Crystal for Rubyists](https://github.com/crystal-lang/crystal/wiki/Crystal-for-Rubyists)
```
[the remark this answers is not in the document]
beta-ziliani: Sounds good. I'll remove it for now and we'll discuss in #470 its proper place

docs/README.md
```diff
@@ -85,20 +73,32 @@ Documentation of the standard library.
 
 Discover the ecosystem of Crystal libraries.
 
+* [Writing Shards](guides/writing_shards.md)
```
straight-shoota, line 76: Can we put this as the last entry? I think it's the most advanced topic and shouldn't be the first one.

docs/SUMMARY.md
straight-shoota: It might not fit the strict definition of what we consider a tutorial, but it fits very well in this place, so let's leave it like that.

docs/SUMMARY.md
straight-shoota: I think we can probably put this in the specification tree?

beta-ziliani:
I can't reply to your last comment (see below), but I don't really see platform support fitting in the description of the language. The closest I can think of is to put it as part of the manual of the compiler.

image

docs/SUMMARY.md
beta-ziliani: I don't really see it fitting in the description of the language. The closest I can think of is to put it as part of the manual of the compiler.

beta-ziliani: Ah I see that we have already stuff related to the compiler in that section, so OK

straight-shoota (approved):

PR #631, opened 2022-08-07
Enforce `markdownlint`

oprypin:
The biggest motivation is to enforce the 4-space list indentation, which I see is often forgotten (and such lists just don't work properly then)

The automated fixes with `markdownlint --fix` are also nice.


.github/workflows/deploy-book.yml
```diff
@@ -54,3 +54,18 @@ jobs:
       - name: Deploy book
         if: github.event_name == 'push' && steps.branch.outputs.branch != null && github.repository == 'crystal-lang/crystal-book'
         run: aws s3 sync ./site s3://crystal-book/reference/${{ steps.branch.outputs.branch }} --delete
+
+  lint:
+    name: Check style issues
+    runs-on: ubuntu-latest
```
straight-shoota, line 60: Please pin a specific release. Otherwise we might see unexpected failures when the base image implicitly updates and breaks something.

straight-shoota (commented):
I don't get why there are some changes converting emphasis from underscore to asterisk and others converting from asterisk to underscore. I understood this should be consistent? Or is it only consistent per file?

oprypin:
I changed to require `**` **strong** and yes - `*` or `_` *emphasis* to be consistent per-file. Otherwise there were too many changes. And I felt that maybe the original authors of those files wouldn't like the change of style 🤔

.github/workflows/deploy-book.yml
oprypin: Hmm but we have `ubuntu-latest` in all workflows in this repo

straight-shoota: Okay, then we can leave it at that for this PR.
This is just something I've been alerted to because we've had some trouble in the past. We should change it in all our CI workflows.

straight-shoota:
I'd argue a consistent style over all files would be useful as well. I don't insist on it - it's not important enough to quarrel about.

oprypin:
OK, changed everything to asterisks. And verified that the result HTML doesn't change.

PR #632, opened 2022-08-16
Add a Debugging tutorial

chances:
My first intent with these changes is to make clear the intended behavior described in https://github.com/crystal-lang/crystal/issues/11589.


beta-ziliani:
Thanks a bunch for this, @chances! I think it fits better in `Guides`, since it's not so much tutorial-y, what do you think? cc/ @straight-shoota

PR #636, opened 2022-09-04
Document new annotation/macro features

Blacksmoke16:
* Document annotations on method parameters added in https://github.com/crystal-lang/crystal/pull/12044
* Document `parse_type` method added in https://github.com/crystal-lang/crystal/pull/11126

Depends on https://github.com/crystal-lang/crystal/pull/12446

### DIFF
--- a/docs/syntax_and_semantics/macros/README.md
+++ b/docs/syntax_and_semantics/macros/README.md
@@ -96,7 +96,7 @@ define_method :foo, 1
 
 ### parse_type
 
-While most AST nodes will be obtained via either manually passed arguments, hard coded values, or retrieved from either the [type](#type-information) or [method](#method-information) information helper variables. There may be some cases where the node you need is not directly accessible, such as if you use information from different contexts to construct the path to the desired type/constant.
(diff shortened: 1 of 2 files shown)


Blacksmoke16:
Moving this to draft as it seems there a formatter bug with annotations on parameters :/

Blacksmoke16:
Okay, so opened a PR to fix the formatter. But not sure how we want to proceed. I.e. ignore the failure and merge as is, or wait until next Crystal release that includes the fix to merge.

docs/syntax_and_semantics/macros/README.md
```diff
@@ -94,6 +94,25 @@ end
 define_method :foo, 1
 ```
 
+### parse_type
+
+While most AST nodes will be obtained via either manually passed arguments, hard coded values, or retrieved from either the [type](#type-information) or [method](#method-information) information helper variables. There may be some cases where the node you need is not directly accessible, such as if you use information from different contexts to construct the path to the desired type/constant.
```
beta-ziliani, line 99: Grammar fix and minor rewording.

```suggestion
Most AST nodes are obtained via either manually passed arguments, hard coded values, or retrieved from either the [type](#type-information) or [method](#method-information) information helper variables. Yet there might be cases in which a node is not directly accessible, such as if you use information from different contexts to construct the path to the desired type/constant.
```

beta-ziliani (approved):

straight-shoota (approved):

PR #640, opened 2022-09-27
Add notice about suffix-less `UInt64` integer literals

HertzDevil:


straight-shoota (approved):

beta-ziliani:
It should be OK to merge, but we can also wait for #641 to get CI green again.

PR #642, opened 2022-09-28
Linting enabled by default

beta-ziliani:
If CI will check for linting, there's no point in having this option off by default. Related: #641.

### DIFF
--- a/.github/workflows/deploy-book.yml
+++ b/.github/workflows/deploy-book.yml
@@ -43,7 +43,7 @@ jobs:
       - name: Populate sample version of Crystal
         run: crystal --version | tee crystal-version.txt
       - name: Build book
-        run: LINT=true make build
+        run: make build
       - name: Configure AWS Credentials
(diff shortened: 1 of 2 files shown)


netlify.toml
```diff
@@ -4,3 +4,5 @@
   command = "mkdocs build -d site --strict"
   publish = "site/"
 
+[context.deploy-preview.environment]
+  LINT = "false"
```
straight-shoota, line 8:
```suggestion
  LINT = "false"

```

oprypin:
> If CI will check for linting, there's no point in having this option off by default.

That's like saying (for a normal programming project) "If CI will run unittests, there's no point allowing contributors to run the program without running unittests first".

The point is that it takes 2.5 seconds instead of 11 seconds before you can see the site.

> Related: #641.

I see that it's literally related, but are you implying that this would prevent that issue? It wouldn't.

PR #646, opened 2022-10-20
Add prepared statement example, etc

rdp:
From crystal-lang/crystal-db#164

### DIFF
--- a/docs/database/README.md
+++ b/docs/database/README.md
@@ -61,6 +61,8 @@ ensure
 end
 ```
 
+Alternatively, you can use `DB.connect` method to open a single connection to the database instead of a pool.
+
 ## Exec
 
 To execute sql statements you can use `Database#exec`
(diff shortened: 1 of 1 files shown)


docs/database/README.md
```diff
@@ -109,6 +111,26 @@ db.query("select name from contacts where age = ?", 33) do |rs|
...
+Query parameters are effected under the covers with prepared statements (sometimes cached),
+or insertion on the client side, depending on the driver, but will always avoid SQL Injection.
+
+If you want to manually use prepared statements, you can with the `build` method:
+
+```crystal
+# MySQL
+prepared_statement = db.build("select * from contacts where id=?")
+# Postgres
+prepared_statement = db.build("select * from contacts where id=$1")
```
straight-shoota, line 123: I'd suggest to use different names for the databases to avoid any confusion.
```suggestion
prepared_statement = mysql_db.build("select * from contacts where id=?")
# Postgres
prepared_statement = postgres_db.build("select * from contacts where id=$1")
```

straight-shoota: The example doesn't work anymore when you only rename the statement variables. The following lines should also use the new names. Sorry, I didn't include that into my suggestion.
It was also meant to start a discussion about improvements.
I think the best way would actually be to show only a single example using one driver-specific syntax (and mentioning alternatives in a comment).

```crystal
prepared_statement = mysql_db.build("select * from contacts where id=?") # Use "... where id=$1" for PostgreSQL
# Use prepared statement:
prepared_statement.query(3) do |rs|
  # ... use rs
end
prepared_statement.query(4) do |rs|
  # ... use rs
end
prepared_statement.close
```

beta-ziliani: Yes, I think it doesn't work as it is now. @rdp can you change that? Otherwise, we'll do it. With that, this is ready. Thanks and sorry for the confusion.

rdp: all righty pushed an update.  I didn't rename it to mysql_db since it doesn't do that elsewhere in the file so felt like it would make sense as part of a bigger change...

straight-shoota (approved):

PR #649, opened 2022-11-05
Refine crystal_for_rubyists

zw963:


docs/crystal_for_rubyists/README.md
```diff
@@ -105,6 +105,14 @@ will generate an error message like
...
 ```
 
+Following code works.
+
+```cr
+[[1, "A"], [2, "B"]].each do |e|
+  pp e[0]
+  pp e[1]
+end
+```
```
straight-shoota, line 115: I'd recommend the explicit unpacking syntax which is easier to use than manual deconstructuring:

````suggestion
[Block parameter unpacking](https://crystal-lang.org/reference/syntax_and_semantics/blocks_and_procs.html#unpacking-block-parameters) works by wrapping the parameters in parenthesis:

```cr
[[1, "A"], [2, "B"]].each do |(a, b)|
  pp a
  pp b
end
```
````

Comment (zw963) [reply docs/crystal_for_rubyists/README.md:115]: The example you added here was exists in this document, at a position further back from the current position.

I add the `do |e|` version just for the completeness, that is, list all work case.

Comment (straight-shoota) [reply docs/crystal_for_rubyists/README.md:115]: I don't this is good. Putting this on front encourages to use this variant, when a more concise option exists.

File: docs/crystal_for_rubyists/metaprogramming_help.md [hunk 2]
```diff
@@ -115,7 +120,7 @@ This example makes use of the `macro extended` hook. This hook is called wheneve
 
 ### Generating Methods via `method_missing` Macro
 
-Sample code + explanation
+[Here](https://github.com/zw963/hashr/blob/master/src/hashr.cr) is a very simple example for how to use method_missing macro, you can check the spec code [here](https://github.com/zw963/hashr/blob/master/spec/hashr_spec.cr), it shows what method_missing macro does.
```
straight-shoota, line 123: I'd rather embed the examples here directly. That makes it easier to skim and the reader doesn't have to navigate from the page.

docs/crystal_for_rubyists/README.md
zw963: I moving my example below the original `do |(a, b)|` one, i consider let user know all possible usage is better.

docs/crystal_for_rubyists/metaprogramming_help.md
zw963: Done, copy the code here. probably need more explain on those code, otherwise, user may still can't understand what is this doing.

docs/crystal_for_rubyists/metaprogramming_help.md
```diff
@@ -115,7 +120,29 @@ This example makes use of the `macro extended` hook. This hook is called wheneve
 
 ### Generating Methods via `method_missing` Macro
 
-Sample code + explanation
+Following is a very simple example that demonstrates how to use `method_missing` macro tocreate the missing method based on the existence of receiver JSON object's key
```
beta-ziliani, line 123:
```suggestion
Following is a very simple example that demonstrates how to use `method_missing` macro to create the missing method based on the existence of receiver JSON object's key
```

beta-ziliani (approved):
A minor nitpick. Thanks a lot @zw963 !

straight-shoota (approved):

PR #652, opened 2022-11-13
Clarify AND and OR semantics

PhilAtWysdom:
The description for AND was not correct if you expect "tmp" to have the same logical result as the former case. Update the documentation so that result variable contains the same result for all inputs and update the OR documentation so that, although the existing code worked, it now corresponds to that of the AND example and reads more clearly. The OR example would also have worked with the unless keyword, but since this section is not discussing unless, I chose to stick with "if !..."

### DIFF
--- a/docs/syntax_and_semantics/and.md
+++ b/docs/syntax_and_semantics/and.md
@@ -5,13 +5,12 @@ An `&&` (and) evaluates its left hand side. If it's *truthy*, it evaluates its r
 You can think an `&&` as syntax sugar of an `if`:
 
 ```crystal
-some_exp1 && some_exp2
+result = some_exp1 && some_exp2
 
 # The above is the same as:
-tmp = some_exp1
-if tmp
+result = if some_exp1
   some_exp2
 else
-  tmp
+  some_exp1
 end
 ```
--- a/docs/syntax_and_semantics/or.md
+++ b/docs/syntax_and_semantics/or.md
@@ -5,13 +5,12 @@ An `||` (or) evaluates its left hand side. If it's *falsey*, it evaluates its ri
 You can think an `||` as syntax sugar of an `if`:
 
 ```crystal
-some_exp1 || some_exp2
+result = some_exp1 || some_exp2
 
 # The above is the same as:
-tmp = some_exp1
-if tmp
-  tmp
-else
+result = if !some_exp1
   some_exp2
+else
+  some_exp1
 end
 ```


straight-shoota:
I don't understand what you think is wrong with the AND example. Could you provide a concrete example that demonstrate the incorrect behaviour?

Your change leads to incorrect code because, `some_exp1` would be evaluated twice. The purpose of `tmp` is to cache the result of `some_exp1` so its value can be used for both the condition and as the value of the `else` branch. This is correctly expressed in the current examples with assignment to a temporary variable.

PhilAtWysdom:
Hi @straight-shoota,

I became confused because the "if" statement appeared to be intended to be used to derive a value rather than to illustrate the point that the result of each branch should be considered to be the flow taken in that case.

In the examples I provided here, the comment that says "# The above is the same as:" is interpreted to mean "# The result of this expression is the same as" (which is how I had interpted it when I first saw it) rather than "The outcome of each of the branches of the if statement is what the overall value of this statement would be, but we haven't assigned it to anything".

I was careful not to misinterpret the assignment to tmp for what it was, but I still misread the intent of the comment in the code saying it was the same.

straight-shoota:
Ah, I see. I think the relevant information here is that for a sequence of expressions, the value of the last expression is the value of the sequence. So the expanded example consists of an assignment (`tmp = some_exp1`) followed by a conditional clause (`if tmp; some_exp2; else; tmp; end`). The value of the latter determines the value of the entire example.

Perhaps it would improve clarity if we split both codes examples into separate blocks?
> ```cr
> some_exp1 && some_exp2
> ```
> The above is the same as:
> ```cr
> tmp = some_exp1
> if tmp
>   some_exp2
> else
>   tmp
> end
> ```

PhilAtWysdom:
@straight-shoota Yes, that's what I'm referring to. It's just a clarity thing, ultimately. Separating the blocks a bit more would probably help avoid confusion, or an alternative comment.

beta-ziliani:
@PhilAtWysdom thanks for helping. I think splitting the blocks is enough, I found the original code more readable once its split. WDYT?

PhilAtWysdom:
@beta-ziliani Yes, I think that does make it clearer.

"Set of statements A" being equivalent to "Set of statements B" as originally intended, separated into two code blocks, for me, reduces the possibility of assuming there should be an overall assignment to a variable in order to get the overall same result out of those statements.

beta-ziliani:
Do you want to implement it, or do you prefer us to take it?

PhilAtWysdom:
Happy for you to do so :)

PR #655, opened 2022-11-29
Add Blocks section to Methods

ftarulla:
In this MR we add a `Blocks` section to `Tutorials` -> `Methods`.

We talk about:
- `captured blocks` and `non-captured blocks`. 
- The `yield` keyword.
- Block parameters and return value using `return`, `next` and `break`.


beta-ziliani (commented):
I didn't check it entirely because I have a high-level point that probably needs addressing. I find the example a bit convoluted, probably because it's artificial. Maybe consider using the `loop` [toplevel method](https://github.com/crystal-lang/crystal/blob/879691b2e/src/kernel.cr#L101)? This could be good to explain when to skip (`next`) and when to `break`.

For instance:

```cr
# print even numbers from 0 to 10

i = 0
loop do
	i += 1
  next if i % 2 == 1
	p i
  break if i == 10
end
```

The example doesn't need to be the same from top to bottom. The important bit is that it can be described easily as I did above, to help understanding that it does (and why is important to learn).

straight-shoota:
This content is quite a lot and hardly fits into the *Methods* lesson (which doubles in size). I'd suggest to introduce it as a separate lesson called *Blocks* in order to keep every single one at a reasonable length.
I would also separate captures into a later lessen called *Procs*. The section about capatured blocks is very short and incomplete. There are a lot more aspects as well as a lot more complexity involved with closures and procs. It's an entire topic on its own and should come later in the tutorial after generics are introduced (and possibly some other concepts).

Regarding blocks, there are also some more aspects that need to be addressed. Not all of them are essential to be fully fleshed out right away, but should be covered eventually.
I think a good order of introduction could look like this:

* Basics: `yield` + `do ... end`
  * Explain the purpose and benefit of blocks, plus demonstrate the effect of inlining
  * multiple `yield`
  * specifics of overloading
* output values (without keywords)
* parameters
  * including underscore, omission, splats and unpacking
  * types
* keywords
  * `next`
  * `break`
  * `return`
* block type restrictions
* alternative call syntaxes:
  * curly braces
  * short block notation

The [language specification](https://crystal-lang.org/reference/1.6/syntax_and_semantics/blocks_and_procs.html) already covers a lot of that and we can copy ideas from there, adapted to a tutorial format.

docs/tutorials/basics/70_blocks.md
```diff
@@ -0,0 +1,570 @@
...
+def transform_string(word : String, &block : String -> String)
+  block_result = yield [word] # Error: argument #1 of yield expected to be String, not Array(String)
+  puts block_result
+end
+
+transform_string "crystal" do |word|
+  word.capitalize
+end
+```
+
```
beta-ziliani, line 353:
```suggestion

NOTE:
Adding a name to a block is useful in rare occasions, like when [capturing a block](https://crystal-lang.org/reference/1.7/syntax_and_semantics/capturing_blocks.html). For the most common cases, the name is omitted, writing just `&` (optionally with the type restriction as in `& : String -> String`).

```

docs/tutorials/basics/70_blocks.md
```diff
@@ -0,0 +1,570 @@
...
+## Alternative syntaxes
+
+We almost at the end of this section. Let's see some other interesting ways of writing *blocks*.
```
beta-ziliani, line 356: This comment suggest the section ends with this subsection, but there's one more subsection below.

```suggestion
There are two handy ways of writing blocks concisely.
```

docs/tutorials/basics/60_methods.md
```diff
@@ -163,3 +163,32 @@ puts life_universe_and_everything + 1 # Error: method top-level life_universe_an
...
+
+puts arr
+```
+
+In the first example we are using the method [Indexable#each](https://crystal-lang.org/api/Indexable.html#each%28%26%3AT-%3E%29-instance-method).
+And in the second example we are using the method [Enumerable#map](https://crystal-lang.org/api/Enumerable.html#map%28%26%3AT-%3EU%29%3AArray%28U%29forallU-instance-method).
+
+To both methods we are passing a *block of code* (delimited by the keywords `do` and `end`) as an argument.
+
+This *block of code*, commonly called *block* is a sequence of interactions, which will be executed at some point. When exactly? It depends on the method that receives the *block* as a parameter (in these examples `each` and `map`).
```
beta-ziliani, line 192: Italics are good for the first occurrence of a topic. Subsequent occurrences should not be emphasized (in a section).

```suggestion
This block of code, commonly called *block* is a sequence of interactions, which will be executed at some point. When exactly? It depends on the method that receives the block as a parameter (in these examples `each` and `map`).
```

docs/tutorials/basics/70_blocks.md
```diff
@@ -0,0 +1,570 @@
+# Blocks
+
+As we already mentioned in the [methods](./60_methods.md#blocks) section, *blocks* is an interesting and important topic. So let's start defining them:
+
+A *block of code* or *block* for short, is a way to abstract a sequence of interactions between objects. They are commonly used as a method parameter, as a way to parameterize (forgive the redundancy) part of its behavior.
+
+## Blocks and Methods
+
+Methods may receive a *block* as an argument. And using the keyword `yield` indicates the place in the method's body where we want to "place" said `block`:
```
beta-ziliani, line 9:
```suggestion
Methods may receive a *block* as an argument. And using the keyword `yield` indicates the place in the method's body where we want to "place" said block:
```

docs/tutorials/basics/70_blocks.md
```diff
@@ -0,0 +1,570 @@
...
+  yield
+  yield
+end
+
+three_times do
+  puts 42
+end
+```
+
+Let's see another examples with a common use scenario for *blocks*: collections.
```
beta-ziliani, line 35:
```suggestion
Let's see another examples with a common use scenario for blocks: collections.
```

docs/tutorials/basics/70_blocks.md
```diff
@@ -0,0 +1,570 @@
...
+while index < arr.size
+  elem = arr[index]
+
+  puts "Hello #{elem}"
+
+  index = index + 1
+end
+```
+
+As we can see in the above examples, the structures are the same: we want to traverse the array and do *something* with the current element at each step. And here is where *blocks* come handy: we can parameterize that *something* we want to execute.
```
beta-ziliani, line 67:
```suggestion
As we can see in the above examples, the structures are the same: we want to traverse the array and do *something* with the current element at each step. And here is where blocks come handy: we can parameterize that *something* we want to execute.
```

docs/tutorials/basics/70_blocks.md
```diff
@@ -0,0 +1,570 @@
...
+
+Now, let's rewrite the examples. We are going to declare a new method that will abstract the structure duplicated in both examples: It will receive the array and what we want to do at each step using a *block*:
```
beta-ziliani, line 69:
```suggestion
Now, let's rewrite the examples. We are going to declare a new method that will abstract the structure duplicated in both examples: It will receive the array and what we want to do at each step using a block:
```

docs/tutorials/basics/70_blocks.md
```diff
@@ -0,0 +1,570 @@
...
+```
+
+The abstracted structure declared in the `with_array` method is very common and that's why it's already implemented in [Indexable#each](https://crystal-lang.org/api/Indexable.html#each%28%26%3AT-%3E%29-instance-method).
+
+NOTE:
+Both blocks declare a parameter (the current element while traversing the array) which we will see next.
+
+## Blocks with parameters
+
+Like methods, *blocks* may receive parameters declared like this: `| param1, param2, ... |`.
```
beta-ziliani, line 102:
```suggestion
Like methods, blocks may receive parameters declared like this: `| param1, param2, ... |`.
```

docs/tutorials/basics/70_blocks.md
```diff
@@ -0,0 +1,570 @@
...
+
+other_method do |n, s, arr| # here the block declare 3 parameters
+  puts "#{s} #{arr[1]}"
+  puts n
+end
+```
+
+### Underscore
+
+What happens if the *block* we are providing does not use all the passed arguments? In that case we can use an underscore just to indicate that an expected parameter it's not used by the block (so no name needed):
```
beta-ziliani, line 119:
```suggestion
What happens if the block we are providing does not use all the passed arguments? In that case we can use an underscore just to indicate that an expected parameter it's not used by the block (so no name needed):
```

docs/tutorials/basics/70_blocks.md
```diff
@@ -0,0 +1,570 @@
...
+One way to implement this would be:
+
+```crystal-play
+arr = [["one", 42], ["two", 24]]
+arr.each do |arg|
+  puts "#{arg[0]} - #{arg[1]}"
+end
+```
+
+This works, but there is a more readable way of writing the same but using parameter unpacking::
```
beta-ziliani, line 144:
```suggestion
This works, but there is a more readable way of writing the same but using parameter unpacking:
```

docs/tutorials/basics/70_blocks.md
```diff
@@ -0,0 +1,570 @@
...
+```
+
+NOTE:
+We use parentheses to unpack the argument into the different block parameters.
+
+Unpacking arguments into parameters works only if the argument's type responds to `[i]` (with `i` an `integer`). In our example `Array` inherits [Indexable#[ ]](https://crystal-lang.org/api/Indexable.html#%5B%5D%28index%3AInt%29-instance-method)
+
+### Splats
+
+When the *block* parameter is a [Tuple](../../syntax_and_semantics/literals/tuple.md) we can use auto-splatting (see [Splats](../../syntax_and_semantics/operators.md#splats)) as a way of destructuring the `tuple` in block parameters (and without the need of parentheses).
```
beta-ziliani, line 160:
```suggestion
When the block parameter is a [Tuple](../../syntax_and_semantics/literals/tuple.md) we can use auto-splatting (see [Splats](../../syntax_and_semantics/operators.md#splats)) as a way of destructuring the tuple in block parameters (and without the need of parentheses).
```

docs/tutorials/basics/70_blocks.md
```diff
@@ -0,0 +1,570 @@
...
+  puts "#{word} - #{number}"
+end
+```
+
+NOTE:
+`Tuples` also implements [Tuple#[ ]](https://crystal-lang.org/api/Tuple.html#%5B%5D%28index%3AInt%29-instance-method) meaning that we can also use *unpacking*.
+
+## Blocks' returned value
+
+A *block*, by default, returns the value of the last expression (the same as a method).
```
beta-ziliani, line 174:
```suggestion
A block, by default, returns the value of the last expression (the same as a method).
```

docs/tutorials/basics/70_blocks.md
```diff
@@ -0,0 +1,570 @@
...
+end
+
+with_number(41) do |number|
+  number + 1
+end
+```
+
+### Returning keywords
+
+We can use the `return` keyword ... but, let's see the following example:
```
beta-ziliani, line 189: OK, this is outright obsessive syndrome kicking in:
```suggestion
We can use the `return` keyword… but, let's see the following example:
```

docs/tutorials/basics/70_blocks.md
```diff
@@ -0,0 +1,570 @@
...
+    number + 1
+  end
+
+  puts "Inside `#test_number` method after `#with_number`"
+end
+
+test_number(-1)
+```
+
+The output is empty! This is because Crystal implements *full closures*, meaning that using `return` inside the block will return, not only from the *block* itself but, from the method where the *block* is defined (i.e. `#test_number` in the above example).
```
beta-ziliani, line 245: `test_number` is not an instance method (but a top-level one).

```suggestion
The output is empty! This is because Crystal implements *full closures*, meaning that using `return` inside the block will return, not only from the block itself but, from the method where the block is defined (i.e. `test_number` in the above example).
```

docs/tutorials/basics/70_blocks.md
```diff
@@ -0,0 +1,570 @@
...
+```
+
+The output is
+
+```console
+
+Inside `#test_number` method after `#with_number`
+```
+
+As we can see the behaviour is something between using `return` and `next`. With `break` we return from the *block* and from the method yielding the *block* (`#with_number` in this example) but not from the method where the `block` is defined.
```
beta-ziliani, line 294:
```suggestion
As we can see the behaviour is something between using `return` and `next`. With `break` we return from the block and from the method yielding it (`with_number` in this example) but not from the method where the block is defined.
```

docs/tutorials/basics/70_blocks.md
```diff
@@ -0,0 +1,570 @@
...
+
+## Type restrictions
+
+Until now we have been using *blocks* without any kind of type restrictions, moreover, we did not declare the block as a method parameter (it was implied by the use of `yield`).
```
beta-ziliani, line 298:
```suggestion
Until now we have been using blocks without any kind of type restrictions, moreover, we did not declare the block as a method parameter (it was implied by the use of `yield`).
```

docs/tutorials/basics/70_blocks.md
```diff
@@ -0,0 +1,570 @@
...
+
+So first, we will declare a *block* as a method parameter: it should be placed last and the parameter's name should be prefixed by `&`. Then we can use `yield` as before.
```
beta-ziliani, line 300:
```suggestion
So first, we will declare a block as a method parameter: it should be placed last and the parameter's name should be prefixed by `&`. Then we can use `yield` as before.
```

docs/tutorials/basics/70_blocks.md
```diff
@@ -0,0 +1,570 @@
...
+end
+```
+
+## Alternative syntaxes
+
+We almost at the end of this section. Let's see some other interesting ways of writing *blocks*.
+
+### Curly braces
+
+Another way of defining a *block* is using `{...}` instead of `do ... end`. Here is one of the examples we have already seen but written with *curly braces syntax*:
```
beta-ziliani, line 360:
```suggestion
Another way of defining a block is using `{...}` instead of `do ... end`. Here is one of the examples we have already seen but written with *curly braces syntax*:
```

docs/tutorials/basics/70_blocks.md
```diff
@@ -0,0 +1,570 @@
...
+  yield 42, "Hello", [1, "Crystal", 3]
+end
+
+other_method { |n, s, arr|
+  puts "#{s} #{arr[1]}"
+  puts n
+}
+```
+
+Here is the  `_` example written with *curly braces syntax* and in one line:
```
beta-ziliani, line 373:
```suggestion
Here is the  `_` example written with curly braces syntax and in one line:
```

docs/tutorials/basics/70_blocks.md
```diff
@@ -0,0 +1,570 @@
...
+end
+
+other_method { |_, _, arr| puts arr[1] }
+```
+
+The main difference between using `do ... end` and `{ ... }` is how they bind the call:
+
+- `do ... end` binds to left-most call
+- `{ ... }` binds to the right-most call
+
```
beta-ziliani, line 387:
```suggestion

Let's see with an example this difference:
```

docs/tutorials/basics/70_blocks.md
```diff
@@ -0,0 +1,570 @@
...
+  puts n
+end
+
+# Same as:
+with_double(generate_number) do |n|
+  puts n
+end
+```
+
+But with `curly braces syntax`:
```
beta-ziliani, line 407:
```suggestion
But with the curly braces syntax:
```

docs/tutorials/basics/70_blocks.md
```diff
@@ -0,0 +1,570 @@
...
+end
+
+def with_double(n : Int32)
+  yield n * 2
+end
+
+with_double generate_number { |n| puts n } # Error: 'generate_number' is not expected to be invoked with a block, but a block was given
+```
+
+The error is because with *curly braces* we are writing: `with_double(generate_number { |n| puts n })` instead of `with_double(generate_number) { |n| puts n }`
```
beta-ziliani, line 421:
```suggestion
The error is because with curly braces we are writing: `with_double(generate_number { |n| puts n })` instead of `with_double(generate_number) { |n| puts n }`
```

docs/tutorials/basics/70_blocks.md
```diff
@@ -0,0 +1,570 @@
...
+transform_hello_crystal { |param| param.capitalize }
+
+# and here is using short one-parameter syntax:
+transform_hello_crystal(&.capitalize)
+
+# and we can omit parentheses:
+transform_hello_crystal &.capitalize
+```
+
+The output should be "Hello crystal". What if we want to capitalize every word.
```
beta-ziliani, line 446:
```suggestion
The output should be "Hello crystal". But how do we want to capitalize every word?
```

docs/tutorials/basics/70_blocks.md
```diff
@@ -0,0 +1,570 @@
...
+def transform_hello_crystal
+  puts yield "hello crystal"
+end
+
+transform_hello_crystal &.split.map(&.capitalize).join(' ')
+```
+
+Great! Let's explain it step by step:
+
+First we split the block parameter using the *short one-parameter syntax* `&.split`. Then, we chained the result applying `map` and using again the *short one-parameter syntax* `.map(&.capitalize)`. And finally we join the `array` of `strings` using `join(' ')`.
```
beta-ziliani, line 458:
```suggestion
First we split the block parameter using the *short one-parameter syntax* `&.split`. Then, we chained the result applying `map` and using again the short one-parameter syntax `.map(&.capitalize)`. And finally we join the `array` of `strings` using `join(' ')`.
```

docs/tutorials/basics/70_blocks.md
```diff
@@ -0,0 +1,570 @@
...
+
+Now, what if we want to parameterize the string we transform. Can we still use the *short one-parameter syntax*? The answer is yes! Let's see:
```
beta-ziliani, line 460:
```suggestion
Now, what if we want to parameterize the string we transform. Can we still use the short one-parameter syntax? The answer is yes! Let's see:
```

docs/tutorials/basics/70_blocks.md
```diff
@@ -0,0 +1,570 @@
...
+def transform_string(word : String)
+  puts yield word
+end
+
+transform_string("hello crystal", &.split.map(&.capitalize).join(' '))
+```
+
+## Under the hood
+
+Before finishing this section of the tutorial, it would be a good idea to see how *blocks* work under the hood.
```
beta-ziliani, line 472:
```suggestion
Before finishing this section of the tutorial, it would be a good idea to see how blocks work under the hood.
```

docs/tutorials/basics/70_blocks.md
```diff
@@ -0,0 +1,570 @@
...
+
+First, it's important to note that in this section we have only seen *blocks* that we use with the keyword `yield`. There is [another kind of block](../../syntax_and_semantics/capturing_blocks.md), but we will leave it for later in the tutorial.
```
beta-ziliani, line 474:
```suggestion
First, it's important to note that in this section we have only seen blocks that we use with the keyword `yield`. There is [another kind of block](../../syntax_and_semantics/capturing_blocks.md), but we will leave it for later in the tutorial.
```

docs/tutorials/basics/70_blocks.md
```diff
@@ -0,0 +1,570 @@
...
+
+In a method that receives a *block*, when we write `yield`, the compiler will inline the *block of code*, which means that this:
```
beta-ziliani, line 476:
```suggestion
In a method that receives a block, when we write `yield`, the compiler will inline the block of code, which means that this:
```

beta-ziliani (changes requested):
Don't be alarm by the amount of comments, there are mostly nitpickings :-)

PR #661, opened 2023-01-07
Add note about PCRE2 usage

straight-shoota:
PCRE2 support was introduced in 1.7.0 as a opt-in preview feature.

cf. https://github.com/crystal-lang/crystal/issues/12790

### DIFF
--- a/docs/syntax_and_semantics/literals/regex.md
+++ b/docs/syntax_and_semantics/literals/regex.md
@@ -11,6 +11,9 @@ A Regex is typically created with a regex literal using [PCRE](http://pcre.org/p
 /あ/
 ```
 
+> NOTE: The compiler expects the syntax of the original PCRE library. Support for the newer PCRE2 library at runtime was added in 1.7.0.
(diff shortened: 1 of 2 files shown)


docs/syntax_and_semantics/literals/regex.md
```diff
@@ -11,6 +11,9 @@ A Regex is typically created with a regex literal using [PCRE](http://pcre.org/p
 /あ/
 ```
 
+> NOTE: The compiler expects the syntax of the original PCRE library. Support for the newer PCRE2 library at runtime was added in 1.7.0.
```
beta-ziliani, line 14:
```suggestion
> NOTE: The compiler expects the syntax of the original [PCRE](https://www.pcre.org/original/doc/html/pcrepattern.html)  library. Support for the newer [PCRE2](https://www.pcre.org/current/doc/html/pcre2syntax.html) library at runtime was added in 1.7.0.
```

beta-ziliani (approved):
Maybe we can add the links to the syntax?

PR #662, opened 2023-01-07
Add section about `lib` function visibility in macro

straight-shoota:
cf. https://github.com/crystal-lang/crystal/pull/12848

### DIFF
--- a/docs/syntax_and_semantics/c_bindings/lib.md
+++ b/docs/syntax_and_semantics/c_bindings/lib.md
@@ -26,5 +26,5 @@ Lib functions are visible in the macro language anywhere in the program.
 lib LibFoo
   fun foo
 end
-{% if LibFoo.methods %} # => [fun foo]
+{{ LibFoo.methods }} # => [fun foo]
 ```
--- a/docs/syntax_and_semantics/c_bindings/lib.md
+++ b/docs/syntax_and_semantics/c_bindings/lib.md
@@ -26,5 +26,6 @@ Lib functions are visible in the macro language anywhere in the program.
 lib LibFoo
   fun foo
 end
+
 {{ LibFoo.methods }} # => [fun foo]
 ```
--- a/docs/syntax_and_semantics/c_bindings/lib.md
+++ b/docs/syntax_and_semantics/c_bindings/lib.md
@@ -17,3 +17,14 @@ Attributes are used to pass flags to the linker to find external libraries:
 * `@[Link(framework: "Cocoa")]` will pass `-framework Cocoa` to the linker (only useful in macOS).
 
 Attributes can be omitted if the library is implicitly linked, as in the case of libc.
+
+## Reflection
+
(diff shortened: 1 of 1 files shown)


docs/syntax_and_semantics/c_bindings/lib.md
```diff
@@ -17,3 +17,14 @@ Attributes are used to pass flags to the linker to find external libraries:
...
+
+## Reflection
+
+Lib functions are visible in the macro language anywhere in the program.
+
+```crystal
+lib LibFoo
+  fun foo
+end
+{% if LibFoo.methods %} # => [fun foo]
```
Blacksmoke16, line 29:
```suggestion
{{ LibFoo.methods }} # => [fun foo]
```

docs/syntax_and_semantics/c_bindings/lib.md
```diff
@@ -17,3 +17,14 @@ Attributes are used to pass flags to the linker to find external libraries:
 * `@[Link(framework: "Cocoa")]` will pass `-framework Cocoa` to the linker (only useful in macOS).
 
 Attributes can be omitted if the library is implicitly linked, as in the case of libc.
+
+## Reflection
+
+Lib functions are visible in the macro language anywhere in the program.
```
beta-ziliani, line 23: I don't think it's clear what "anywhere" means here. Also, probably we want to link to the API.

```suggestion
Lib functions are visible in the macro language using the method [`TypeNode#methods`](https://crystal-lang.org/api/1.6.2/Crystal/Macros/TypeNode.html#methods%3AArrayLiteral%28Def%29-instance-method):
```

beta-ziliani (commented):
It's a bit odd that we're saying this about libs, but not about other types. But I guess it's better to have something somewhere than anything, so OK.

docs/syntax_and_semantics/c_bindings/lib.md
```diff
@@ -17,3 +17,14 @@ Attributes are used to pass flags to the linker to find external libraries:
...
+
+```crystal
+lib LibFoo
+  fun foo
+end
+{{ LibFoo.methods }} # => [fun foo]
```
beta-ziliani, line 29: Formatting is complaining
```suggestion

{{ LibFoo.methods }} # => [fun foo]
```

docs/syntax_and_semantics/c_bindings/lib.md
straight-shoota: "anywhere" might not be very clear, but it's important. `TypeNode#methods` for other types only works within method context. The point we're mentioning this here is that for lib types it also works outside of methods.

beta-ziliani (approved):

docs/syntax_and_semantics/c_bindings/lib.md
beta-ziliani: but if it's important, it should be clear. anyway, we can leave it for a future improvement.

PR #671, opened 2023-02-27
Document top-level `fun`s

### LINKED ISSUE #667: Document `fun` defs
https://www.reddit.com/r/crystal_programming/comments/10n55km/declaring_fun_on_the_toplevel/

HertzDevil:
Resolves #667.


docs/syntax_and_semantics/c_bindings/fun.md
```diff
@@ -93,3 +93,19 @@ end
...
+fun sum(x : Int32, y : Int32) : Int32
+  x + y
+end
+
+sum(1, 2) # => 3
+```
+
+Like lib funs, all parameter and return types are mandatory, and only valid types in C bindings are permitted. Like regular Crystal defs, they have a body and can be called directly by other Crystal code.
+
+Because Crystal programs are not meant to be used as shared libraries, top-level methods should be defined using the `def` keyword instead of `fun`. Exceptions are entry points used by the underlying operating system, such as the C `main` on most systems.
```
straight-shoota, line 111: The "are not meant to be used as shared libraries" might be a bit strong. That applies mostly to stdlib which imposes restrictions on usefulness as a library. But this is the language specification, not stdlib.
I'd maybe tone that down a little bit.

PR #680, opened 2023-03-30
about naming conflicts in namespaces

noraj:
cf. https://github.com/crystal-lang/crystal/issues/13203

### DIFF
--- a/docs/SUMMARY.md
+++ b/docs/SUMMARY.md
@@ -64,6 +64,7 @@
             * [Class methods](syntax_and_semantics/class_methods.md)
             * [Class variables](syntax_and_semantics/class_variables.md)
             * [finalize](syntax_and_semantics/finalize.md)
+            * [Naming conflicts](syntax_and_semantics/naming_conflicts.md)
         * [Modules](syntax_and_semantics/modules.md)
         * [Generics](syntax_and_semantics/generics.md)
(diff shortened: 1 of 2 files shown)


straight-shoota:
Thanks for this contribution.

I think we should put this in a different place. It's not really a specification of a language feature. That would just be the last sentence and that's already covered in https://crystal-lang.org/reference/1.7/syntax_and_semantics/the_program.html#top-level-scope

This piece feels more like a mini howto: If you need to deal with this, you can do that (this = naming conflict, that = top-level namespace reference). Maybe we can put it in the *Guides* section? But guides are typically larger about more complex topics.
Actually, the scope could be a good format for https://stackoverflow.com/ which we endorse as a knowledge base for these kind of problems.

Anyways, I would suggest to make the example fully self-contained. The reader should not need to know anything outside the presented code.
```cr
module Bar
  def self.bar
  end
end

module Foo
  def self.foo
    ::Bar.bar
  end

  module Bar
  end
end

Foo.foo
```

noraj:
Maybe a "Common issues" section under https://crystal-lang.org/reference/1.7/guides/index.html?

beta-ziliani:
I think it's better to have something like this in the tutorial, although at the moment the tutorial itself is rather poor

PR #685, opened 2023-04-13
Add section about Hash syntaxes in Crystal for Rubyists

crimson-knight:
This adds an explicit section about the shorthand hash syntax from Ruby defining a NamedTuple in Crystal to help avoid some syntax confusion between the languages.

### DIFF
--- a/docs/crystal_for_rubyists/README.md
+++ b/docs/crystal_for_rubyists/README.md
@@ -440,7 +440,7 @@ In Crystal, this is not the case. The hash rocket `=>` syntax is required to dec
 }
 ```
 
-However, the hash shorthand syntax in Ruby creates a [NamedTuple](https://crystal-lang.org/api/latest/NamedTuple.html) in Crystal.
+However, the hash shorthand syntax in Ruby creates a [`NamedTuple`](https://crystal-lang.org/api/NamedTuple.html) in Crystal.
 
 `NamedTuple`s and regular [`Tuple`s](https://crystal-lang.org/api/Tuple.html) have a fixed size, so these are best used for data structures that are known at compile time.
 
--- a/docs/crystal_for_rubyists/README.md
+++ b/docs/crystal_for_rubyists/README.md
@@ -431,7 +431,7 @@ However, the `Hash` shorthand syntax in Ruby creates a `NamedTuple` in Crystal.
 ```crystal
 # Creates a valid `Hash(Symbol, String)` in Crystal
 {
-  :key1 => "some value",
+  :key1      => "some value",
   :some_key2 => "second value",
 }
 
--- a/docs/crystal_for_rubyists/README.md
+++ b/docs/crystal_for_rubyists/README.md
@@ -406,7 +406,7 @@ end
 
 ### Hash syntax from Ruby to Crystal
 
-Crystal introduces a data type that is not available in Ruby, the NamedTuple.
+Crystal introduces a data type that is not available in Ruby, the [`NamedTuple`](https://crystal-lang.org/api/NamedTuple.html).
 
 Typically in Ruby you can define a hash with several syntaxes:
 
@@ -424,7 +424,9 @@ Typically in Ruby you can define a hash with several syntaxes:
 }
 ```
 
-In Crystal, this is not the case. The hash rocket `=>` syntax is required to declare a hash in Crystal.
+In Crystal, this is not the case. The `Hash` rocket `=>` syntax is required to declare a hash in Crystal.
+
+However, the `Hash` shorthand syntax in Ruby creates a `NamedTuple` in Crystal.
 
 ```crystal
 # Creates a valid `Hash(Symbol, String)` in Crystal
@@ -433,14 +435,12 @@ In Crystal, this is not the case. The hash rocket `=>` syntax is required to dec
   :some_key2 => "second value"
 }
 
-# Creates a NamedTuple in Crystal
+# Creates a `NamedTuple(Symbol, String)` in Crystal
 { 
   key1: "some value",
   some_key2: "second value"
 }
 ```
-
-However, the hash shorthand syntax in Ruby creates a [`NamedTuple`](https://crystal-lang.org/api/NamedTuple.html) in Crystal.
 
 `NamedTuple`s and regular [`Tuple`s](https://crystal-lang.org/api/Tuple.html) have a fixed size, so these are best used for data structures that are known at compile time.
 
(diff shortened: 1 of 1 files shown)


devnote-dev (changes requested):
I think it would be better to use `NamedTuple` instead of interchanging with `Tuple` as they are different data types.

docs/crystal_for_rubyists/README.md
```diff
@@ -404,6 +404,46 @@ private def method
...
+{
+  :key1 => "some value",
+  :some_key2 => "second value"
+}
+```
+
+In Crystal, this is not the case. The hash rocket `=>` syntax is required to declare a hash in Crystal.
+
+```crystal
+# Creates a valid hash in Crystal
```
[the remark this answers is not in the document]
crimson-knight: For now, let's keep the example simple because it's about the syntax and not the types.

docs/crystal_for_rubyists/README.md
```diff
@@ -404,6 +404,46 @@ private def method
...
+# Creates a NamedTuple in Crystal
+{ 
+  key1: "some value",
+  some_key2: "second value"
+}
+```
+
+However, the hash shorthand syntax in Ruby creates a [NamedTuple](https://crystal-lang.org/api/latest/NamedTuple.html) in Crystal.
+
+Tuples are a fixed size, so these are best used for data structures that are known at compile time.
```
crimson-knight, line 445:
```suggestion
NamedTuples and regular [Tuples](https://crystal-lang.org/api/latest/Tuple.html) are a fixed size, so these are best used for data structures that are known at compile time.
```

crimson-knight:
@devnote-dev took your suggestion and added some clarification and an additional link to the Tuple docs page

devnote-dev (approved):

HertzDevil (changes requested):

docs/crystal_for_rubyists/README.md
```diff
@@ -404,6 +404,46 @@ private def method
...
+}
+
+# Creates a NamedTuple in Crystal
+{ 
+  key1: "some value",
+  some_key2: "second value"
+}
+```
+
+However, the hash shorthand syntax in Ruby creates a [NamedTuple](https://crystal-lang.org/api/latest/NamedTuple.html) in Crystal.
```
straight-shoota, line 443: This introductory paragraph should better be placed *above* the example that demonstrates it.

docs/crystal_for_rubyists/README.md
```diff
@@ -404,6 +404,46 @@ private def method
...
+# A valid ruby hash declaration
+{ 
+  key1: "some value",
+  some_key2: "second value"
+}
+
+# This syntax in ruby is shorthand for the hash rocket => syntax
+{
+  :key1 => "some value",
+  :some_key2 => "second value"
```
Fryguy, line 423: Do we want an example with string keys on both sides or does that make it too complicated?

docs/crystal_for_rubyists/README.md
```diff
@@ -404,6 +404,46 @@ private def method
...
+}
+```
+
+In Crystal, this is not the case. The hash rocket `=>` syntax is required to declare a hash in Crystal.
```
Fryguy, line 427: Wondering if "hash" here and elsewhere should be `Hash`

Fryguy (commented):
Love this change. This one was confusing coming from Ruby experience, and it wasn't until I understood the difference, on the Crystal side, between Hash/NamedTuple and their purpose that it clicked.

docs/crystal_for_rubyists/README.md
crimson-knight: The goal here is just to explain the syntax difference, so I hadn't planned on explaining the differences in how keys are handled.

docs/crystal_for_rubyists/README.md
```diff
@@ -404,6 +404,46 @@ private def method
 end
 ```
 
+### Hash syntax from Ruby to Crystal
+
+Crystal introduces a data type that is not available in Ruby, the NamedTuple.
```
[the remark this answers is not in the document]
crimson-knight: Updated, and made the link to the docs on the first reference to `NamedTuple` instead of mid-way through

docs/crystal_for_rubyists/README.md
crimson-knight: I moved this up above the example as suggested

docs/crystal_for_rubyists/README.md
```diff
@@ -404,6 +404,46 @@ private def method
...
+However, the `Hash` shorthand syntax in Ruby creates a `NamedTuple` in Crystal.
+
+```crystal
+# Creates a valid `Hash(Symbol, String)` in Crystal
+{
+  :key1 => "some value",
+  :some_key2 => "second value"
+}
+
+# Creates a `NamedTuple(Symbol, String)` in Crystal
```
HertzDevil, line 438:
```suggestion
# Creates a `NamedTuple(key1: String, some_key2: String)` in Crystal
```

HertzDevil (changes requested):

straight-shoota (approved):

docs/crystal_for_rubyists/README.md
```diff
@@ -404,6 +404,46 @@ private def method
...
+
+In Crystal, this is not the case. The `Hash` rocket `=>` syntax is required to declare a hash in Crystal.
+
+However, the `Hash` shorthand syntax in Ruby creates a `NamedTuple` in Crystal.
+
+```crystal
+# Creates a valid `Hash(Symbol, String)` in Crystal
+{
+  :key1 => "some value",
+  :some_key2 => "second value"
```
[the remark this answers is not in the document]
crimson-knight: This should be correct now

crimson-knight:
> 

Yup, I missed that. Can you tell I don't normally run the formatter? 😊

docs/crystal_for_rubyists/README.md
```diff
@@ -404,6 +404,46 @@ private def method
...
+
+```crystal
+# Creates a valid `Hash(Symbol, String)` in Crystal
+{
+  :key1      => "some value",
+  :some_key2 => "second value",
+}
+
+# Creates a `NamedTuple(key1: String, some_key2: String)` in Crystal
+{
```
straight-shoota, line 439:
```suggestion
{
```

HertzDevil (approved):

PR #687, opened 2023-04-14
Update section about regex literals for PCRE2

straight-shoota:


HertzDevil (approved):
The API documentation for `Regex` needs similar changes (some URLs still point to the PCRE1 pages)

PR #692, opened 2023-05-05
Rework section about relational operators

straight-shoota:
The main purpose is to separate the pattern matching and case subsumption operators (`=~`, `!~` and `===`) from the equality operators (`==`, `!=`). They are *not* about equality. The case subsumption operator should also not be called *case equality* because it's misleading. 

The entire group of operators including equality, subsumption and inequalities is called relational operators.

In the process of this restructuring I've also reworded and enhanced the descriptions for these operators with the purpose of adding clarity and filling gaps. Also added some references to related features in the standard library.

The result is still a bit thin in some areas (especially regarding pattern matching and case subsumption) but I'm not sure about some of the specific details myself. We'll need to figure that out sometime.

### DIFF
--- a/docs/syntax_and_semantics/operators.md
+++ b/docs/syntax_and_semantics/operators.md
@@ -194,7 +194,7 @@ Three base operators test equality:
 * `=~`: Checks whether the value of the first operand matches the value of the
 second operand with pattern matching.
 * `===`: Checks whether the left hand operand matches the right hand operand in
-  [case equality](case.md). This operator is applied in `case ... when`
+  [case subsumption](case.md). This operator is applied in `case ... when`
   conditions.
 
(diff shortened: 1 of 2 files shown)


docs/syntax_and_semantics/operators.md
```diff
@@ -184,53 +184,95 @@ ones.
...
 
-#### Comparison
+#### Inequalities
+
+<span id="comparison" />
+
+Inequality operators describe the order between values.
+
+The **three-way comparison operator** `<=>` (also known as *spaceship operator*)
+expresses the total order between two elements expressed by the sign of its
```
HertzDevil, line 227: There is only a partial order if `#<=>` may return `nil`.

HertzDevil (changes requested):

HertzDevil (approved):

PR #693, opened 2023-05-05
Improve Modules page readability

a-alhusaini:
I had to look for the difference between extend and include earlier today. I missed it the first time I opened the Modules documentation because it didn't have its own headline.

This change will improve scannability significantly


straight-shoota (approved):
You might want to indent the following headline, "extend self"

PR #694, opened 2023-05-10
Add devenv

straight-shoota:
This sets up a reproducible development environment via [devenv.sh](https://devenv.sh).
Using it is completely optional, but I've come to like it as a tool that ensures the working environment is consistent and functional.
Usage instructions are added to the readme.

The first two commits are independent preconditions:

* markdownlint-cli@0.34.0 complains about spaces inside code span elements. As a mitigation this test is deactivated for the affected code.
* Resolution of python dependencies has been removed from the Makefile. That was a rather odd thing and has been bugging me for some time. We shouldn't dictate the use of pip and venv for python dependencies. The makefile should only require that `mkdocs` is available. Previously it was impossible to install dependencies in a different way (which the devenv setup needs, for example).

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,3 @@
+source_url "https://raw.githubusercontent.com/cachix/devenv/d1f7b48e35e6dee421cfd0f51481d17f77586997/direnvrc" "sha256-YBzqskFZxmNb3kYVoKD9ZixoPXJh1C9ZvTLGFRkauZ0="
+
+use devenv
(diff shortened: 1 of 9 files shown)


beta-ziliani:
Why did you remove `make deps`?

straight-shoota:
`make deps` did use pip to install mkdocs in a venv environment. That's all a bit quirky and a makefile is not the right place to setup tool dependencies. This is cleaner now: The makefile expects `mkdocs` to be in the path (overridable via `$MKDOCS`) and doesn't care where and how it is installed.
So it continues to work when mkdocs is installed with pip and venv, but it also works for other setups such as nix.

beta-ziliani (approved):

PR #699, opened 2023-07-05
Document dynamic linking on Windows and `CRYSTAL_LIBRARY_RPATH`

HertzDevil:
This doc page is no longer strictly related to static linking, suggestions are welcome

EDIT: Also something needs to be done about the duplicate headings


docs/guides/static_linking.md
```diff
@@ -56,18 +56,26 @@ Alpine’s package manager APK is also easy to work with to install static libra
 
 macOS doesn't [officially support fully static linking](https://developer.apple.com/library/content/qa/qa1118/_index.html) because the required system libraries are not available as static libraries.
 
+### Windows
+
+Windows doesn't support fully static linking Because the Win32 libraries are not available as static libraries.
```
straight-shoota, line 61:
```suggestion
Windows doesn't support fully static linking because the Win32 libraries are not available as static libraries.
```

docs/guides/static_linking.md
```diff
@@ -56,18 +56,26 @@ Alpine’s package manager APK is also easy to work with to install static libra
...
+
+Currently, static linking is the default mode of linking on Windows, and dynamic linking can be opted in via the `-Dpreview_dll` compile-time flag. In order to distinguish static libraries from DLL import libraries, when the compiler searches for a library `foo.lib` in a given directory, `foo-static.lib` will be attempted first while linking statically, and `foo-dynamic.lib` will be attempted first while linking dynamically. The official Windows packages are distributed with both static and import libraries for all third-party dependencies, except for LLVM.
+
+Static linking implies using the static version of Microsoft's C runtime library (`/MT`), and dynamic linking implies the dynamic version (`/MD`); extra C libraries should be built with this in mind to avoid linker warnings about mixing CRT versions. There is currently no way to use the dynamic CRT while linking statically.
```
straight-shoota, line 65: These paragraphs don't really fit under the heading *Fully Static Linking*. Maybe this is a bit too much nit-picking. I'm happy to leave it as is for now and refactor later.

Ideally, the content should be restructured, probably having a section about *Static Linking* (with subsections per platform) and *Fully Static Linking* would be a subsection of that. And this one would be relatively short, basically just saying *Nay* for Windows and Mac (due to system libs) and *Yes, but not glibc* for Linux.

docs/guides/static_linking.md
```diff
@@ -56,18 +56,26 @@ Alpine’s package manager APK is also easy to work with to install static libra
 
 macOS doesn't [officially support fully static linking](https://developer.apple.com/library/content/qa/qa1118/_index.html) because the required system libraries are not available as static libraries.
 
+### Windows
+
+Windows doesn't support fully static linking Because the Win32 libraries are not available as static libraries.
+
+Currently, static linking is the default mode of linking on Windows, and dynamic linking can be opted in via the `-Dpreview_dll` compile-time flag. In order to distinguish static libraries from DLL import libraries, when the compiler searches for a library `foo.lib` in a given directory, `foo-static.lib` will be attempted first while linking statically, and `foo-dynamic.lib` will be attempted first while linking dynamically. The official Windows packages are distributed with both static and import libraries for all third-party dependencies, except for LLVM.
```
straight-shoota, line 63: Add some context to import libraries to make it better understandable.
```suggestion
Currently, static linking is the default mode of linking on Windows, and dynamic linking can be opted in via the `-Dpreview_dll` compile-time flag. In order to distinguish static libraries from DLL import libraries, when the compiler searches for a library `foo.lib` in a given directory, `foo-static.lib` will be attempted first while linking statically, and `foo-dynamic.lib` will be attempted first while linking dynamically. The official Windows packages are distributed with both static and DLL import libraries for all third-party dependencies, except for LLVM.
```

HertzDevil:
The duplicate headings are indeed not a problem, but apparently `.markdownlintrc` had the wrong format?

straight-shoota (approved):

PR #706, opened 2023-08-11
Better enum symbol casting documentation

### LINKED ISSUE #705: Symbol enum auto-casting
https://crystal-lang.org/reference/1.9/syntax_and_semantics/enum.html#usage

There should be documentation on how you can pass a symbol into a parameter typed as an enum, and it will auto-cast it to one of the enum members. I think the example of taking in a symbol and manually casting it to an enum should be removed as well.

```crystal
def paint(color : Color)
  case color
  when Color::Red
    # ...
  else
    # Unusual, but still can happen
    raise "unknown color: #{color}"
  end
end

paint Color::Red
paint :red # equivalent to above, auto-casted to Color::Red

paint :not_a_color # raises a compile-time exception due to auto-cast failing
```

nobodywasishere:
Closes #705

### DIFF
--- a/docs/syntax_and_semantics/enum.md
+++ b/docs/syntax_and_semantics/enum.md
@@ -137,4 +137,4 @@ paint :red # automatically casts to `Color::Red`
 paint :yellow
 ```
 
-The same automatic casting does not apply to case statements. To use enums with case statements, see [case enum values](case.html#enum-values).
+The same automatic casting does not apply to case statements. To use enums with case statements, see [case enum values](case.md#enum-values).
--- a/docs/syntax_and_semantics/enum.md
+++ b/docs/syntax_and_semantics/enum.md
@@ -1,6 +1,7 @@
 # Enums
 
-This page is for [Crystal enums](https://crystal-lang.org/api/Enum.html). For C enums, see [C bindings enum](c_bindings/enum.md).
+!!! note
+    This page is for [Crystal enums](https://crystal-lang.org/api/Enum.html). For C enums, see [C bindings enum](c_bindings/enum.md).
 
 An enum is a set of integer values, where each value has an associated name. For example:
 
--- a/docs/syntax_and_semantics/enum.md
+++ b/docs/syntax_and_semantics/enum.md
@@ -1,5 +1,7 @@
 # Enums
 
+This page is for [Crystal enums](https://crystal-lang.org/api/Enum.html). For C enums, see [C bindings enum](c_bindings/enum.md).
+
 An enum is a set of integer values, where each value has an associated name. For example:
 
 ```crystal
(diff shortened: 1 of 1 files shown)


Andriamanitra:
I think we should keep the old "Usage" section and add a new section for the symbol-enum autocasting. With the removal of the old example it is no longer clear how to use enums with a case statement. Especially this part that got removed provided helpful information:

> However, if the programmer makes a typo, say `:reed`, the error will only be caught at runtime, while attempting to use `Color::Reed` will result in a compile-time error.
> 
> The recommended thing to do is to use enums whenever possible, only use symbols for the internal implementation of an API, and avoid symbols for public APIs. But you are free to do what you want.

nobodywasishere:
I could add an example like:
```
case color
when Color::Red
  # red color
else
  # unknown color
end
```
But I think that should be an additional example, not included in the `def paint` code example for clarity. I did also add a link to how to use enums with case statements as defined on the case page, which provides a different example.

I don't think the sections you quoted relate to how to use a case statement, but to the benefits of using enums over symbols. Maybe it could fit but honestly don't know if it does tonally. Maybe it could be something like:

> If a symbol parameter is being used to indicate flags or a set of options, it is recommended to use an enum instead. If the wrong symbol argument is used, enum symbol casting provides a compile-time error letting you know there's a problem. This is in contrast to using a using a symbol parameter which may not catch an error at all without manual value checking or guard statements, leading to issues downstream.

docs/syntax_and_semantics/enum.md
```diff
@@ -1,5 +1,7 @@
 # Enums
 
+This page is for [Crystal enums](https://crystal-lang.org/api/Enum.html). For C enums, see [C bindings enum](c_bindings/enum.md).
```
straight-shoota, line 3: Maybe this could be put into an admonition?

```suggestion
!!! note
    This page is for [Crystal enums](https://crystal-lang.org/api/Enum.html). For C enums, see [C bindings enum](c_bindings/enum.md).
```

docs/syntax_and_semantics/enum.md
```diff
@@ -117,37 +119,22 @@ Class variables are allowed, but instance variables are not.
 
 ## Usage
 
-Enums are a type-safe alternative to [Symbol](https://crystal-lang.org/api/Symbol.html). For example, an API's method can specify a [type restriction](type_restrictions.md) using an enum type:
+When a method parameter has an enum [type restriction](type_restrictions.md), it accepts either an enum constant or a Symbol. The Symbol will be automatically cast to an enum constant, raising a compile-time error if casting fails.
```
straight-shoota, line 122:
```suggestion
When a method parameter has an enum [type restriction](type_restrictions.md), it accepts either an enum constant or a [symbol](literals/symbol.md). The symbol will be automatically cast to an enum constant, raising a compile-time error if casting fails.
```

docs/syntax_and_semantics/enum.md
```diff
@@ -117,37 +119,22 @@ Class variables are allowed, but instance variables are not.
...
-  end
-end
+paint :red # automatically casts to `Color::Red`
 
-paint :red
+# compile-time error:
+#   expected argument #1 to 'paint' to match a member of enum Color
+#
+#   Options are: :red, :green and :blue
+paint :yellow
```
straight-shoota, line 137: Errors are usually formatted like this:
```suggestion
paint :yellow # Error: expected argument #1 to 'paint' to match a member of enum Color
```

I think it's sufficient to show the main error message and skip the extra help here.

straight-shoota:
I think it's fine like this. The paragraph mentioned in https://github.com/crystal-lang/crystal-book/pull/706#issuecomment-1674259990 is not very relevant for the enum page.
It talks about the shortcomings of symbol. So it should probably better be placed on the symbol page.

straight-shoota (approved):

beta-ziliani (approved):

PR #712, opened 2023-09-28
Enable edit link

straight-shoota:
This adds an "Edit this page" button on every page to facilitate easy access for contributing.

The `edit_uri` was already configured (although with a view url, not edit), but the actual edit button needs to be enabled as well.


beta-ziliani (approved):
Yes!

PR #714, opened 2023-10-09
Add manual entries for `tool dependencies` and `tool unreachable`

straight-shoota:
These tools were added in these PRs: 

* https://github.com/crystal-lang/crystal/pull/13631
* https://github.com/crystal-lang/crystal/pull/13783

### DIFF
--- a/docs/man/crystal/README.md
+++ b/docs/man/crystal/README.md
@@ -379,11 +379,16 @@ Options:
 * `-fFORMAT`, `--format=FORMAT`: Output format `tree` (default), `flat`, `dot`, or `mermaid`.
 * `-iPATH`, `--include=PATH`: Include path in output.
 * `-ePATH`, `--exclude=PATH`: Exclude path in output.
+* `--verbose`: Show skipped and heads of filtered paths
 * `--error-trace`: Show full error trace.
-* `--prelude`: Specify prelude to use. The default one initializes the garbage
+* `-h`, `--help`: Show this message
+* `--prelude PATH`: Specify prelude to use. The default one initializes the garbage
   collector. You can also use `--prelude=empty` to use no preludes. This can be
   useful for checking code generation for a specific source code file.
-* `--verbose`: Show skipped and heads of filtered paths
+* `-s`, `--stats`: Enable statistics output
+* `-p`, `--progress`: Enable progress output
+* `-t`, `--time`: Enable execution time output
+* `--stdin-filename`: Source file name to be read from STDIN
 
 ### `crystal tool format`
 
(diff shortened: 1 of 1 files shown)


docs/man/crystal/README.md
```diff
@@ -354,11 +354,36 @@ crystal play [--port <port>] [--binding <host>] [--verbose] [file]
...
+crystal tool dependencies [options] [programfile]
+```
+
+Options:
+
+* `-DFLAG`, `--define=FLAG`: Define a compile-time flag. This is useful to
+  conditionally define types, methods, or commands based on flags available at
+  compile time. The default flags are from the target triple given with
+  `--target-triple` or the hosts default, if none is given.
+* `-fFORMAT`, `--format=FORMAT`: Output format `tree` (default), `flat`, `dot`, or `mermaid`.
```
beta-ziliani, line 379: There's an inconsistency of the presentation for flags. Here you say `-fFORMAT` and `--format=FORMAT` and in the unreachable tool you place `-f text|json` and `--format text|json`. Same for include and exclude.

docs/man/crystal/README.md
```diff
@@ -354,11 +354,36 @@ crystal play [--port <port>] [--binding <host>] [--verbose] [file]
...
+* `-iPATH`, `--include=PATH`: Include path in output.
+* `-ePATH`, `--exclude=PATH`: Exclude path in output.
+* `--error-trace`: Show full error trace.
+* `--prelude`: Specify prelude to use. The default one initializes the garbage
+  collector. You can also use `--prelude=empty` to use no preludes. This can be
+  useful for checking code generation for a specific source code file.
+* `--verbose`: Show skipped and heads of filtered paths
```
beta-ziliani, line 386: it's also missing other modifiers like -s, -p, ...

beta-ziliani (changes requested):
There are inconsistencies in the formats

docs/man/crystal/README.md
```diff
@@ -379,11 +379,16 @@ Options:
 * `-fFORMAT`, `--format=FORMAT`: Output format `tree` (default), `flat`, `dot`, or `mermaid`.
 * `-iPATH`, `--include=PATH`: Include path in output.
 * `-ePATH`, `--exclude=PATH`: Exclude path in output.
```
beta-ziliani, line 381: To be consistent, and also the previous line (GH is not allowing me to pick all the lines). It won't be consistent all over (that's impossible at this point), but at least it will be consistent here.
```suggestion
* `-f FORMAT`, `--format FORMAT`: Output format `tree` (default), `flat`, `dot`, or `mermaid`.
* `-i PATH`, `--include PATH`: Include path in output.
* `-e PATH`, `--exclude PATH`: Exclude path in output.
```

PR #724, opened 2023-12-20
Document new `@caller` macro ivar

Blacksmoke16:
Documentation PR for https://github.com/crystal-lang/crystal/pull/14048. Will follow up with another PR to add a link to this section from the API docs of `#doc_comment` itself.


docs/syntax_and_semantics/macros/README.md
```diff
@@ -342,6 +342,29 @@ end
 Foo.boo(0, 1)
 ```
 
+## Call Information
+
+When a macro is called, you can access the macro call stack with a special instance variable: `@caller`.
+This variable returns an `ArrayLiteral` of [`Call`](https://crystal-lang.org/api/Crystal/Macros/Call.html) nodes with the first element in the array being the most recent,
+unless it is called outside of a macro in which case it's a [`NilLiteral`](https://crystal-lang.org/api/Crystal/Macros/NilLiteral.html).
```
straight-shoota, line 349: **polish(style):** Shorter sentences for better readability.
```suggestion
This variable returns an `ArrayLiteral` of [`Call`](https://crystal-lang.org/api/Crystal/Macros/Call.html) nodes with the first element in the array being the most recent.
Outside of a macro or if the macro has no caller (e.g. a [hook](hook)) the value is a [`NilLiteral`](https://crystal-lang.org/api/Crystal/Macros/NilLiteral.html).
```

docs/syntax_and_semantics/macros/README.md
```diff
@@ -450,7 +473,7 @@ Notice the variables in the inner macro are not available within the `verbatim`
 
 ## Comments
 
-Macro expressions are evaluated both within comments as well as compilable sections of code. This may be used to provide relevant documentation for expansions:
+Macro expressions are evaluated both within comments as well as compatible sections of code. This may be used to provide relevant documentation for expansions:
```
straight-shoota, line 476: **issue:** I think this wording is intentional.
I understand "compilable sections of code" as code blocks inside a doc comment.

docs/syntax_and_semantics/macros/README.md
```diff
@@ -475,6 +498,31 @@ a
...
+# Comment on macro call.
+gen_method foo
+```
+
+When generated, the docs for the `#foo` method would be like:
+
+```text
+# Comment on macro call.
+#
+# Comment added via macro expansion.
```
straight-shoota, line 523: **polish:** The generated docs don't have comment syntax.
```suggestion
Comment on macro call.

Comment added via macro expansion.
```

straight-shoota (approved):

PR #725, opened 2023-12-20
Mention argless `#annotations` method in annotation docs

Blacksmoke16:
Long overdue mention of https://github.com/crystal-lang/crystal/pull/9326.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,5 @@
 site/
 .venv/
-__pycache__/
 
 # Devenv
 .devenv*
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 site/
 .venv/
+__pycache__/
 
 # Devenv
 .devenv*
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -1,5 +1,6 @@
 site/
 .venv/
+__pycache__/
```
straight-shoota, line 3: **question:** This seems to be an unrelated change. What's the motivation or where does it come from?

Blacksmoke16: It get created automatically when I was running `make serve`, but with `mkdocs` coming from `.venv`.

straight-shoota: I guess it's fine to add this to `.gitignore`. But let's drop it here and do a separate PR for that.
/cc @oprypin

straight-shoota (approved):

Blacksmoke16:
Bump, this should be good to go as well.

straight-shoota:
Sorry, I forgot to add this to the project and thus lost track of it (unfortunately, GH only allows automatic adding from one repo 🤷). Feel free to add it yourself next time. Or merge directly if you feel lucky 😆

PR #728, opened 2024-01-02
Document `alignof` and `instance_alignof`

HertzDevil:


docs/syntax_and_semantics/alignof.md
```diff
@@ -0,0 +1,40 @@
...
+end
+
+alignof(Foo) # => 2
+alignof(Bar) # => 1
+```
+
+For [Reference](https://crystal-lang.org/api/Reference.html) types, the alignment is the same as the alignment of a pointer:
+
+```crystal
+# On a 64 bits machine
```
straight-shoota, line 28: **typo:**
```suggestion
# On a 64-bit machine
```

docs/syntax_and_semantics/alignof.md
```diff
@@ -0,0 +1,40 @@
...
+alignof(Pointer(Int32)) # => 8
+alignof(String)         # => 8
+```
+
+This is because a Reference's memory is allocated on the heap and a pointer to it is passed around. To get the effective alignment of a class, use [instance_alignof](instance_alignof.md).
```
straight-shoota, line 33: **typo:**
```suggestion
This is because a reference's memory is allocated on the heap and a pointer to it is passed around. To get the effective alignment of a class, use [instance_alignof](instance_alignof.md).
```

docs/syntax_and_semantics/alignof.md
```diff
@@ -0,0 +1,40 @@
...
+struct Bar
+  def initialize(@x : Int8, @y : Int16)
+  end
+end
+
+alignof(Foo) # => 2
+alignof(Bar) # => 1
+```
+
+For [Reference](https://crystal-lang.org/api/Reference.html) types, the alignment is the same as the alignment of a pointer:
```
straight-shoota, line 25: **note:** Linking to the API docs is better than nothing, but we're really missing an explanation of reference types in the book.

docs/syntax_and_semantics/instance_alignof.md
```diff
@@ -0,0 +1,18 @@
+# instance_alignof
+
+The `instance_alignof` expression returns an `Int32` with the instance alignment of a given class. For example:
```
straight-shoota, line 3: **chore:** Add a reference to `alignof`.
```suggestion
The `instance_alignof` expression returns an `Int32` with the instance alignment of a given class. It's the equivalent to `[alignof](alignof.md)` but for reference types, where `alignof` refers to the alignment of the pointer representation.
For example:
```

HertzDevil: The wording sounds rather confusing. If it is already "equivalent" to `alignof` what difference does it make?

HertzDevil:
For the record these pages are near copies of `sizeof` and `instance_sizeof`, so I would prefer the four pages not to diverge in style, and to address the common issues in a separate PR

straight-shoota:
I don't think there's a need to follow the errors of a template. But we can do it that way, I don't mind either way.

docs/syntax_and_semantics/instance_alignof.md
straight-shoota: Feel free to adjust the wording to make more sense for you.

straight-shoota (approved):

PR #732, opened 2024-01-09
Improve wording on alignof and sizeof pages

straight-shoota:
Applies the improvements mentioned in https://github.com/crystal-lang/crystal-book/pull/728#pullrequestreview-1800694828 across all pages.


HertzDevil:
I'm thinking if we could refer to the newly added `ReferenceStorage` from the `instance_*` docs (although I'm afraid that might create a circular definition)

ysbaddaden (approved):

PR #734, opened 2024-01-14
Add `scripts/validate-flags.sh`

straight-shoota:
This script extracts all `flag?` macro values from the source code and compares them with the documentation.
It lists all flags that are currently defined but undocumented.

It depends on `crystal tool flags`, proposed in https://github.com/crystal-lang/crystal/pull/14234

<del>This could be automated in CI to check flag documentation regularly.</del>
<ins>Adds a CI workflow that regularly checks flag documentation coverage.</ins>

### DIFF
--- a/.github/workflows/validate-content.yml
+++ b/.github/workflows/validate-content.yml
@@ -1,0 +1,28 @@
+name: Validate content
+on:
+  push:
+  pull_request:
+    branches: [master, release/*]
+  schedule:
+    - cron: '0 5 * * 1'  # Every Monday 5 AM UTC
+
+jobs:
+  validate-flags:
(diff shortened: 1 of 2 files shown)


straight-shoota:
As @Blacksmoke16 observed in https://github.com/crystal-lang/crystal-book/pull/735#issuecomment-1891038352, this script did miss out on flags that are handled by the compiler directly.

I added that now. It's not a fancy compiler tool that investigates the AST. But `crystal tool flags` is useful for any Crystal program, whereas `Program#has_flag?` is only relevant for the compiler. So a crude grep implementation does the job as well. There's less variation because the method only accepts `String`. And we control all the code.

PR #735, opened 2024-01-14
Add documentation for missing flags

straight-shoota:
Detected via https://github.com/crystal-lang/crystal-book/pull/734


Blacksmoke16:
That script seems to have missed `no_restrictions_augmenter`, as mentioned in https://forum.crystal-lang.org/t/inconsistent-document-about-autocasting/6295/7

straight-shoota:
The script only lists flags that are actually used in Crystal code via `flag?`. It does not cover flags that are handled by the compiler directly.

straight-shoota:
Might be able to add that, though.

beta-ziliani (approved):

PR #737, opened 2024-01-20
Add guide: Use zig cc as a alternative linker (draft version)

zw963:
This approach was originally verified by @luislavena and first posted [here](https://forum.crystal-lang.org/t/a-github-action-template-for-do-release-assets-when-push-a-tag-it-should-fitable-for-any-shards/4635/5), [here](https://forum.crystal-lang.org/t/a-github-action-template-for-do-release-assets-when-push-a-tag-it-should-fitable-for-any-shards/4635/8) and [here](https://forum.crystal-lang.org/t/does-anyone-deploy-crystal-app-on-arm-based-linux-server-what-is-the-deployment-process-like/5588/5), big thanks to him for proving this solutin work, and he showed a little bit of this in following videos:

    [Building CLIs with Crystal - Quick cross-compilation - Part 1](https://www.youtube.com/watch?v=ij7alYEvfTg)
    [Building CLIs with Crystal - Quick cross-compilation - Part 2](https://www.youtube.com/watch?v=LdVNqdf_kBI)

Here are some links explain how `zig cc` link work:

[`zig cc`: a Powerful Drop-In Replacement for GCC/Clang](https://andrewkelley.me/post/zig-cc-powerful-drop-in-replacement-gcc-clang.html)

[How Uber Uses Zig](https://jakstys.lt/2022/how-uber-uses-zig/)


Blacksmoke16:
My opinion is I think this would be better suited to a blog post versus in the official documentation.

zw963:
If think this doc is useless, please feel to close it, anyway, i post here just for response @straight-shoota 's reply [here](https://forum.crystal-lang.org/t/cross-build-crystal-compiler-but-without-build-libgc-a-but-lib-crystal-libgc-a-exists-in-offical-released-tarball/6333/16?u=zw963)

Blacksmoke16:
I don't think its _useless_, just that it seems like a fairly complex process, and being in the official book implies that it's an official process. I.e. that we would be expected to maintain/provide support for it.

zw963:
> I don't think it's _useless_, just that it seems like a fairly complex process, and being in the official book implies that it's an official process. I.e. that we would be expected to maintain/provide support for it.
> 
> Just my 2 cents, will let others give their thoughts as well.

Personal, i think this is a fairly easy approach for cross compile/link compare to docker buildx(the latter is the only way to link for crystal application for now), it's even simple enough to use a bash script to handle most of the details, since i know i can do like this, i never start my docker container for now.

You think this process complex, maybe because you didn't read it carefully?

----------

The name of this post includes `draft version`(maybe need a better name), that told the reader,  this process still needs improve, even, use at your risk, but, there is no doubt about that this process improve the crystal user experience a lot when user need do cross compile.

ysbaddaden:
Confirmed. I had to specify `-lunwind` because it failed to find the `_Unwind_*` symbols by itself. The other args and libraries have been specified.

```console
$ echo 'puts "hello world"' > x.cr
$ CC="zig cc" crystal build x.cr --link-flags="-lunwind" --verbose
zig cc [many .o files] -lunwind -rdynamic -L/usr/lib/crystal -lpcre2-8 -L/usr/local/lib -lgc -lpthread -ldl -lpthread -levent -lrt -lpthread -ldl
$ ./x
hello world
```

zw963:
> And I notice that `-lpthread` appears thrice and `-ldl` twice 🤨

I notice this before too.

>  I'd encourage to first write more about this in forums or blog posts to raise awareness (and debug situations).

I probably misunderstood the intention about [this](https://forum.crystal-lang.org/t/cross-build-crystal-compiler-but-without-build-libgc-a-but-lib-crystal-libgc-a-exists-in-offical-released-tarball/6333/16?u=zw963) post by @straight-shoota ?

zw963:
Anyway, this fills the gap in the Crystal ecosystem by directly linking from the local machine to generate (static) binaries for other platforms, as the Golang compiler does, the Crystal compiler cannot achieve this

straight-shoota:
I would definitely like to see a guide on using alternative linkers with Crystal.
But this probably needs to grow and mature a bit more until we should publish it in the Crystal book. Until then it's probably better placed in a personal blog, the forum, or similar.

ysbaddaden:
Oh right, cross compilation + cross linking executables. That would be nice to document.

Though zig cc may help, it's not the only solution, and the same steps usually apply: install a cross toolchain, cross compile the required libraries, compile with `--cross-compile` then link manually or specify a mix of `CC` and `--link-flags` for crystal to call the linker.

I'd love to read a bunch of articles on the subject before we start to consolidate the information in an official guide.

straight-shoota:
Let's close this then (for the time being).

PR #740, opened 2024-01-21
Set `edit_uri` to point to deployed branch

straight-shoota:
Currently, `edit_uri` always points to the master branch. This is confusing when you're on a release branch because the contents may differ. Release branches (especially the latest one) is the main deployed version and you won't see the master deployment unless you explicitly chose the "nightly" version.

Example for confusion: https://github.com/crystal-lang/crystal/pull/11876#issuecomment-1902599532

### DIFF
--- a/.github/workflows/deploy-book.yml
+++ b/.github/workflows/deploy-book.yml
@@ -44,6 +44,9 @@ jobs:
           fi
       - name: Populate sample version of Crystal
         run: crystal --version | tee crystal-version.txt
+      - name: Set edit_uri branch
+        run: |
+          sed -i -e 's#^edit_uri:.*#edit_uri: https://github.com/${{ github.repository }}/edit/${{ github.ref_name }}/docs/#' mkdocs.yml
       - name: Build book
         run: LINT=true make build
       - name: Configure AWS Credentials


beta-ziliani:
I don't get it, why would you want to edit a different branch than master?

straight-shoota:
I amended the OP with some details.

oprypin:
Uuh I don't think it would be a better outcome if a contributor were to edit the docs and send pull requests for a random older release

oprypin:
It might be ok to have the View link point to the branch and the Edit link point to master. But still a bit scary that people may waste time editing older docs. Also a bit messy to accomplish that

straight-shoota:
I'm pretty convinced that *when* there's an edit (or view) button, it should please point to the *same* version of the page. It's unexpected and confusing that it takes you to a different branch.

When you're on any deployed branch other than the latest release, there's a big warning message on the top of the page that says you're viewing an older version. So it should be obvious that if you're not on the latest version, you might want to check a newer one before editing.

You should always look into open issues and PRs before doing any major work, or risk implementing a change that somebody else has already worked on but isn't merged yet. The edit link is just a convenient short cut to get to the edit page for the content you're currently reading (and not some other version of it!).

We could consider going into a similar direction as the warning message. E.g. if you're on an older version the edit link would show a specific notification message that you're about to edit an outdated version of the page and recommend checking the current one.
I'm not sure I would like that from a UX perspective, though. When I'm viewing a page in an older version, I expect the edit link to take me to that old version of the page.

oprypin:
Opened a pull request according to my comment, please consider it

* https://github.com/crystal-lang/crystal-book/pull/742

oprypin:
image

straight-shoota:
I still maintain that it's confusing for either of these links to point to a different version. It's even more confusing if they both have different targets🤦

oprypin:
That's fine. But an edit link pointing to a branch that we don't want people to edit is worse.

You commented that the warning about non-latest version somehow helps. But it doesn't help at all, we want people to edit master and not the latest release branch

straight-shoota:
I disagree on that premise. Incremental edits that are valid for the current release *should* actually go into the latest release branch. That makes them visibile directly in the main deployed main version of the book.

Only documentation about unreleased features of the next release should go into `master`.

oprypin:
Ah OK if that's the system.
No comments from me then

straight-shoota:
I suppose it's not a super clean workflow. And it's rarely exercised because there are not that many contributions to this repo.

beta-ziliani:
But then again, you should only update the release branch or master, and not an older released one. Edit should be disabled for any branch older than current release.

straight-shoota:
> Edit should be disabled for any branch older than current release.

I suppose we could do that, but it adds a bit complexity to the deployment. We'll need to rebuild the old branch with disabled edit button configuration on every new release. I'm not sure that's worth it.
If someone ends up editing an old branch, it's most likely not a big deal.

beta-ziliani (approved):

PR #744, opened 2024-01-26
Add branch info to README

straight-shoota:


straight-shoota:
I'm wondering if we should call the branch for the next release `next` and let `master` always point to the latest `release/*` branch. That might be more clear?

PR #746, opened 2024-01-29
Document `asm`

HertzDevil:


straight-shoota (approved):

straight-shoota:
CI is failing. Looks like we found a formatter bug 😆

straight-shoota:
I posted a formatter fix in https://github.com/crystal-lang/crystal/pull/14278
Until that's available, we can either disable the formatter for this code example, or format it in the way the current formatter likes it. Benefit of the latter would be an automatic notification to correct it when the formatter is fixed =)

PR #747, opened 2024-01-29
Add documentation for `select`

straight-shoota:


ysbaddaden (commented):
Are you sure we should document the internal implementation details until we move on to a public API? For the time being I'd document sugar syntax (the supported expressions) and forego the details of the syntax expansion (internal detail).

BTW: I'm surprised that `#timeout_select_action` is documented, it even returns a `Channel::TimeoutAction` that isn't documented.

Also: we can also select on `Channel#send(M)`.

straight-shoota:
> Also: we can select on `Channel#send(M)`.

Not sure what you mean by that... Could you clarify?

ysbaddaden:
I agree the syntax expansion shouldn't change, but we never know and we can't use the knowledge anyway (because private classes) :shrug:

I mean that we can select on send too, but I see you did list `Channel#send_select_action`. There is no example, though:

```crystal
select
when channel.send(message)
when timeout(5.seconds)
end
```

straight-shoota:
Yeah, we'll need a guide on how to use `select` with standard select actions. This should be a separate document though, and not part of the language reference.

docs/syntax_and_semantics/select.md
```diff
@@ -0,0 +1,52 @@
+# select
+
+The `select` expression chooses from a set of blocking operations and proceeds with the branch that becomes available first.
```
bcardiff, line 3: Available first is not 100% accurate in MT or even if both are already available. I'm not sure how to frame the difference.

docs/syntax_and_semantics/select.md
```diff
@@ -0,0 +1,52 @@
...
+when exit_channel.receive
+  exit
+when timeout(5.seconds)
+  puts "Timeout"
+end
+```
+
+## Select actions
+
+A select action call calls a method with the implicit suffix "_select_action".
```
bcardiff, line 33: The `?` suffix means the transformation is `foo?` -> `foo_select_action?`, and `foo` -> `foo_select_action` 🕵️ .

I don't remember if `foo` and `foo_select_action` are constrained to have some relation in their type signature.

docs/syntax_and_semantics/select.md
```diff
@@ -0,0 +1,52 @@
...
+Optionally, the last branch may be `else` (without condition).  This denotes the `select` action as non-blocking.
+The expression closes with an `end` keyword.
+
+> NOTE:
+> `select` is similar to a [`case` expression](./case.md) with all branches referring to potentially blocking operations.
+
+Each condition is either a call to a select action or an assignment whose right-hand side is a call to a select action.
+
+```crystal
+select
```
bcardiff, line 19: This example is not exhaustive: `send` and `.receive?` are missing.

bcardiff (commented):
I always come back to https://gist.github.com/bcardiff/289953a80eb3a0512a2a2f8c8dfeb1db to see the potential uses of select and it's variation.

That guide might serve better for a guide that covers the use cases rather than the language semantics.

Maybe for the language semantics, if we want to be precise we should mention the conversion of select statements to `Channel.select`/`Channel.non_blocking_select` and refer to that (future) guide for further uses.

The proposed document is short and concise, so if we iterate on the comments I think is a good addition. Definitely the usage guide will be well received.

docs/syntax_and_semantics/select.md
straight-shoota: If any actions are directly executable, `select_impl` immediately returns the result of the first action with delivery state `delivered` or `closed`. So IIUC "available first" should be good for that scenario.

With multithreading everything is a bit more complicated, of course. But even there the action that first manages to activate the select context is chosen.

Maybe we just need a different word for "available"? 
```suggestion
The `select` expression chooses from a set of blocking operations and proceeds with the branch that activates first.
```

docs/syntax_and_semantics/select.md
straight-shoota: I added a mention of the suffix.

> I don't remember if `foo` and `foo_select_action` are constrained to have some relation in their type signature.

I don't think so. There's no need for any constraints because *all* calls in a `select` expression's `where` condition are expanded. If a method signature doesn't match, the type checker will complain.

bcardiff: They do.

```crystal
def foo_select_action : Channel::TimeoutAction
  timeout_select_action(1.second)
end

a =
  select
  when x = foo
    x
  else
    nil
  end
```

complains with

```
In foo.ign.cr:15:12

 15 | when x = foo
               ^--
Error: undefined local variable or method 'foo' for top-level
```

If we define `foo` with the wrong type

```
def foo : Int32
  0
end
```

we get

```
 14 | select
      ^
Error: can't cast (Channel::NotReady | Nil) to Int32
```

Since `Channel::TimeoutAction` includes `SelectAction(Nil)` we need `def foo : Nil` for things to compile

straight-shoota: Hm, interesting. I have no idea where in the compiler that would be happening, though 😆 
Literal expander just transforms the calls without any checking.

bcardiff (approved):

PR #762, opened 2024-05-28
Clarify type inference of class variables

straight-shoota:
This part of the documentation was causing confusion, so I'm expanding a bit on the preconditions for type inference.

https://forum.crystal-lang.org/t/do-you-need-to-specify-a-type-for-class-variables/6876/2

### DIFF
--- a/docs/syntax_and_semantics/class_variables.md
+++ b/docs/syntax_and_semantics/class_variables.md
@@ -24,7 +24,28 @@ Counter.instances # => 3
 
 Class variables can be read and written from class methods or instance methods.
 
-Their type is inferred using the [global type inference algorithm](type_inference.md).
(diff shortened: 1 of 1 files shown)


nobodywasishere (approved):

beta-ziliani:
Perhaps a clarification is in order, but it isn't different from instance vars, so why be specific about it here?

zw963:
@nobodywasishere , @straight-shoota , @beta-ziliani 

ping, did you folks forget to merge this?

beta-ziliani (approved):

zw963:
Too long to merge a doc, For documents, we should adopt a simpler and more efficient mechanism to merge them quickly. After all, even a slight improvement is better than leaving things unchanged.

(Translated with qwen3.5:35b)

PR #769, opened 2024-07-13
Dark Mode

timscodebase:
You docs sorely need a dark mode.  Here is a simple solution with batteries for new features

### DIFF
--- a/docs/assets/style.css
+++ b/docs/assets/style.css
@@ -1,53 +1,133 @@
-.md-header, .md-sidebar--primary {
-    /* pull z-index up to be above the codemirror gutters (z-index: 3).
+:root {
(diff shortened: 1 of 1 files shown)


timscodebase:
Dark mode

straight-shoota:
Mkdocs-material already has dark mode implemented. I think we only need to enable it.

Blacksmoke16:
Yea, should just need to be a configuration change. The related docs for it are: https://squidfunk.github.io/mkdocs-material/setup/changing-the-colors/#automatic-light-dark-mode.

straight-shoota:
Closing. We shouldn't run a custom implementation for dark mode and instead use the upstream feature configuration.

PR #772, opened 2024-07-22
Bump Windows platform to Tier 2

Blacksmoke16:
Much [progress](https://forum.crystal-lang.org/t/contribute-to-windows-support/5585) has been made in regards to window support over the last years. While there are still some things here and there that need to be polished, there are no longer "major limitations" and things are expected to just work. As such, it seems like a reasonable step to bump Windows up to a Tier 2 level of support.

There is still work to be done to get it over the last hump to Tier 1, but this should be a welcomed change to show it _is_ stable and supported.

Related: https://github.com/crystal-lang/crystal/issues/5430#issuecomment-2228661548

### DIFF
--- a/docs/syntax_and_semantics/platform_support.md
+++ b/docs/syntax_and_semantics/platform_support.md
@@ -42,6 +42,7 @@ Details are described in the *Comment* column.
 | `i386-linux-musl` | x86 Linux | kernel 4.14+, MUSL libc 1.2+<br> *(expected to work on kernel 2.6.18+)* | :material-selection-ellipse: tests<br> :material-selection-ellipse: builds
 | `x86_64-openbsd` | x64 OpenBSD | 6+ | :material-selection-ellipse: tests<br> :material-selection-ellipse: builds
 | `x86_64-freebsd` | x64 FreeBSD | 12+ | :material-selection-ellipse: tests<br> :material-selection-ellipse: builds
+| `x86_64-windows-msvc` | x64 Windows (MSVC ) | 7+ | :material-circle-slice-7: tests<br> :material-checkbox-marked-circle: builds |
 
 ***
 
@@ -54,7 +55,6 @@ Most typically, some parts of the standard library are not supported completely.
 
 | Target | Description | Supported versions | Comment |
 | ------ | ----------- | ------------------ | ------- |
-| `x86_64-windows-msvc` | x64 Windows (MSVC ) | 7+ | :material-circle-slice-7: tests<br> :material-checkbox-marked-circle: builds |
 | `aarch64-linux-android` | aarch64 Android  | Bionic C runtime, API level 24+ | :material-selection-ellipse: tests<br> :material-selection-ellipse: builds |
 | `x86_64-unknown-dragonfly` | x64 DragonFlyBSD | | :material-selection-ellipse: tests<br> :material-selection-ellipse: builds |
 | `x86_64-unknown-netbsd` | x64 NetBSD | | :material-selection-ellipse: tests<br> :material-selection-ellipse: builds |


straight-shoota:
I suppose the notion of major limitations is a bit subjective. In my opinion there are still too many rough edges in core parts of the standard library to consider it fully working on Windows.
Especially considering spawning processes with shell semantics is unresloved: https://github.com/crystal-lang/crystal/issues/9030, https://github.com/crystal-lang/crystal/issues/12873, https://github.com/crystal-lang/crystal/issues/14300

The effective difference between Tier 1 and 2 is that the latter either is not (fully) automatically tested or we don't provide official builds.
We have both for Windows. So it would go directly into Tier 1 once we determine it fully working.

Blacksmoke16:
My thinking was, yes there are some things that still need polished, but it's not like where it was even a few years ago where there were a bunch of unchecked boxes in https://github.com/crystal-lang/crystal/issues/5430. I.e. sure if you try and use `Process.new(command, args, shell: true)` this specific command you might run into some wonkyness, but otherwise it works fine. Like the entirety of Athena CI passes on Windows so that at least says to me it's probably entirely fine for 95%+ of users. There just seems to be a lot of animosity around not having Windows support yet when that is just not true.

Maybe there's another solution to this, but to me tier 2's desc of:

> The requirements for *Tier 1* may be partially fulfilled, but are lacking in some way that prevents a solid gurantee.

Is a better statement in regards to how well developed support is than tier 3's desc of:

> but there are some major limitations. Most typically, some parts of the standard library are not supported completely

Given "lacking in some way" equates to things being rough around the edges but otherwise is "expected to work", versus just not working at all.

:shrug:

beta-ziliani (approved):
Totally agree with Georges description. But I think this merge should be postpone until we rename the binary to stop saying "unsupported", otherwise we'll get an inconsistency.

Blacksmoke16:
Is that just a documentation change? Only reference I could find was https://github.com/crystal-lang/distribution-scripts/blob/1b7fb7ff2a2a9d535ec95dd3aedbf8e1fc627212/processes/crystal-release.md?plain=1#L59-L61. So seemingly would be an easy change. Are we wanting to go with the `-portable.zip` proposed over in https://github.com/crystal-lang/crystal/issues/5430#issuecomment-2232059612?

beta-ziliani:
It's a documentation change, plus a post noting the new status and name

straight-shoota:
Moving a platform into Tier 2 or Tier 1 announce it's considered to be fully supported.
I think there are still a couple of blockers for full Windows support.

A number of open issues in [the Windows Project](https://github.com/orgs/crystal-lang/projects/11/views/5) are categorized as *Critical* or *Important*. Some of them are already being worked on (so it's hopefully just a matter of time). But others are still pending in the stage of a design discussion and their advancement is undetermined.
Most undecided issues affect several aspects of spawning processes: https://github.com/crystal-lang/crystal/issues/9030, https://github.com/crystal-lang/crystal/issues/12873, https://github.com/crystal-lang/crystal/issues/14300
It might be fine to move forward without having all of them *completely* resolved. But at this point we're not even decided on *how* a solution could look like. Until we can certainly exclude that we might require some breaking changes for that, we should not increase the support level on Windows to where it would be impossible to do so.

Looking at the wider ecosystem, there are problems with `shards`. Especially postinstall on Windows is a major DX issue. I would also consider this a blocker. It's not good to declare Windows fully supported but as soon as a new Windows user tries installing a shard, it blows up. See https://github.com/crystal-lang/shards/issues/468 for details. Note that the problem is not even that many shards have limited Windows support. Even just installing a shard with `postinstall` is usually broken or at least unnecessarily hard to work around with `shards install --skip-postinstall --skip-executables` (https://github.com/crystal-lang/shards/issues/498).

It's important to consider that declaring full Windows support is an important topic for PR. We can and should use this announcement (whenever it comes) to draw attention to the language, particularly shouting out to Windows users. It'll hopefully draw in a new curious users who want to take a look at the language. If they percieve a shitty developer experience the whole attention backfires.
We need to fill the missing bits of stdlib and polish DX on Windows. This unfortunately still requires some time and effort.

Blacksmoke16:
Yea I think you could argue it either way. However there's no real _need_ or hurry for merge this so I'd be fine regrouping closer to 1.15 and see where things stand then. Probably wouldn't hurt moving forward on renaming of the executable in the meantime tho?

straight-shoota:
We can certainly start making all the infra ready.
I'd do the *actual* rename once we proceed with the tier promotion. Preparing users of the artifact should be done in advance though. `install-crystal` looks fine already, the `-unsupported` suffix is optional in https://github.com/crystal-lang/install-crystal/blob/b833130c6d883cdb78113e1bf4b8faecb8139390/index.js#L434
I'm not sure if anything else needs attention.

zw963:
I bet that 90% of Crystal users may have already forgotten about Crystal's Windows support ...

So, it really doesn't make much sense, unless you guys want all users to forget it.

PR #775, opened 2024-08-05
Update platform support tiers for macOS

straight-shoota:
CI swichted from x86 runners to aarch64 runners in https://github.com/crystal-lang/crystal/pull/14833

This means, we no longer have automated tests for x86_64-darwin, so it goes into Tier 2. And aarch64-darwin moves up to Tier 1.


Blacksmoke16:
Can `x86_64` not stay as tier one as specs are covered via https://github.com/crystal-lang/crystal/blob/7ee895fa8703aa7955b65bd9f12b8f6cf32835a9/.github/workflows/macos.yml#L20-L21?

straight-shoota:
Closing in favour of #779

PR #789, opened 2024-10-20
Add examples on "stacking" and "overriding" certain macro hooks

sol-vin:
We can "stack" (not sure what else to call it) certain macro hooks like `finished` but something like `method_missing` will override previous macro hooks in the same context. Feel free to rename or whatever. 👍🏻

### DIFF
--- a/docs/syntax_and_semantics/macros/hooks.md
+++ b/docs/syntax_and_semantics/macros/hooks.md
@@ -131,11 +131,12 @@ A definition of the `method_missing` macro hook overrides any previous definitio
 macro method_missing(name)
   {% puts "I didnt run! :(" %}
 end
+
 class Example
   macro method_missing(name)
     {% puts "I didnt run! :(" %}
   end
-  
+
   macro method_missing(name)
     {% puts "I am the only one that will run!" %}
   end
--- a/docs/syntax_and_semantics/macros/hooks.md
+++ b/docs/syntax_and_semantics/macros/hooks.md
@@ -132,16 +132,16 @@ class Example
   end
   
   macro method_missing(name)
-    {% puts "I am the only one that will run! :D" %}
+    {% puts "I am the only one that will run!" %}
   end
 end
 
 macro method_missing(name)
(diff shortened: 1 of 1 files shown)


docs/syntax_and_semantics/macros/hooks.md
```diff
@@ -103,3 +103,45 @@ Foo.new.bar
...
+end
+
+macro finished
+  {% puts "I will also execute!" %}
+end
+```
+In the above example, both `finished` macros will execute. Stacking works for the following hooks: `inherited`, `included`, `extended`, `method_added`, `finished`
+
+## Overriding
+Overriding can only be used with the `method_missing` macro. This will only execute the last defined macro `method_missing` in a context. For example:
```
straight-shoota, line 124: **polish:** Instead of describing "Overriding" as a feature that's only available for `method_missing`, describe it as behaviour of that macro hook.
```suggestion
A definition of the `method_missing` macro hook overrides any previous definition of this hook in the same context. Only the last defined macro executes. For example:
```

sol-vin:
All suggestion handled 🫡

straight-shoota (approved):

sol-vin:
Fixed whatever the markdown cop was complaining about. 👮🏻‍♂️

straight-shoota (approved):

sol-vin:
Ok, ran `crystal tool format` but it's adding a newline which I'm pretty sure the markdown cop will cry about. Not sure what to do but it seems like there are a bunch of other examples that have `crystal tool format --check` errors and I didn't touch any of them. 🤷🏻‍♂️

sol-vin:
Luckily markdown cop had no whining but deploy book still failed, but none of the code is anything I added 😄

straight-shoota:
Yeah we'll ignore those failures for now.
I also realized this PR should go against the current version of the book (`release/1.14`), so I changed the base branch.

beta-ziliani (approved):
Thanks!

Issue #723, opened 2023-12-14
Documentation for `include` inside lib structs


HertzDevil:
You can do this since https://github.com/crystal-lang/crystal/commit/a8dc4fb81ddb0881281fa4617c4237a5a5e908ad:

```crystal
lib Lib
  struct Foo
    x : Int32
    y : Int16
  end

  struct Bar
    include Foo
    z : Int8
  end
end

Lib::Bar.new # => Lib::Bar(@x=0, @y=0, @z=0)
```

I don't know if this has ever been used in the wild at all, but we should probably document it regardless.

sol-vin:
Worked on it 🫡

PR #799, opened 2024-11-14
Add `include` info to c-binding's `struct`

sol-vin:
As per #723

### DIFF
--- a/docs/syntax_and_semantics/c_bindings/struct.md
+++ b/docs/syntax_and_semantics/c_bindings/struct.md
@@ -41,7 +41,7 @@ lib C
 end
 ```
 
-Structs that are defined inside a `lib` can be included as modules internally in other `lib` defined structs, for example:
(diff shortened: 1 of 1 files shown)


docs/syntax_and_semantics/c_bindings/struct.md
```diff
@@ -41,6 +41,24 @@ lib C
 end
 ```
 
+Structs that are defined inside a `lib` can be included as modules internally in other `lib` defined structs, for example:
```
straight-shoota, line 44: **polish:** This is only a tiny detail, might be better to not framt it "include _as_ a module" but "include _like_ a module". The struct is still a struct, not a module.

```suggestion
Structs that are defined inside a `lib` can be included, like modules, internally in other `lib` defined structs, for example:
```

ysbaddaden:
Oh my, I didn't even know about that :astonished:

ysbaddaden (approved):

Issue #663, opened 2023-01-12
Lib enum example does not work


HertzDevil:
This example:

https://github.com/crystal-lang/crystal-book/blob/867b2c8c2e71b4cdda91698a3e898cfc93c9f683/docs/syntax_and_semantics/c_bindings/enum.md?plain=1#L48-L56

fails with `Error: enum value must be an Int32` due to crystal-lang/crystal#2703. (It also fails if `SomeEnum` is explicitly marked as an `Int32` enum: crystal-lang/crystal#13292)

PR #801, opened 2024-11-14
Fix `lib` `enum` code example

sol-vin:
As described in #663

### DIFF
--- a/docs/syntax_and_semantics/c_bindings/enum.md
+++ b/docs/syntax_and_semantics/c_bindings/enum.md
@@ -51,8 +51,6 @@ lib X
     A = 1_u32 # Error: enum value must be an Int32
   end
 end
-
-X::SomeEnum 
 ```
 
 However, you can change this default type:
--- a/docs/syntax_and_semantics/c_bindings/enum.md
+++ b/docs/syntax_and_semantics/c_bindings/enum.md
@@ -48,11 +48,11 @@ The type of an enum member is `Int32` by default, even if you specify a differen
 ```crystal
 lib X
   enum SomeEnum
-    A = 1_u32
+    A = 1_u32 # => Error: enum value must be an Int32
   end
 end
 
-X::SomeEnum # => 1_i32
+X::SomeEnum 
 ```
 
 However, you can change this default type:
@@ -65,8 +65,8 @@ lib X
   end
 end
 
-X::SomeEnum::Zero # => 0_i8
-X::SomeEnum::Two  # => 2_i8
+puts X::SomeEnum::Zero.value # => 0_i8
+puts X::SomeEnum::Two.value  # => 2_i8
 ```
 
 You can use an enum as a type in a `fun` parameter or `struct` or `union` members:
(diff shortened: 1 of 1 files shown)


docs/syntax_and_semantics/c_bindings/enum.md
```diff
@@ -48,11 +48,11 @@ The type of an enum member is `Int32` by default, even if you specify a differen
 ```crystal
 lib X
   enum SomeEnum
-    A = 1_u32
+    A = 1_u32 # => Error: enum value must be an Int32
```
straight-shoota, line 51: **issue:** I cannot add a line comment there, but the line before this code block needs to be updated as well.

docs/syntax_and_semantics/c_bindings/enum.md
```diff
@@ -48,11 +48,11 @@ The type of an enum member is `Int32` by default, even if you specify a differen
...
   end
 end
 
-X::SomeEnum # => 1_i32
+X::SomeEnum
```
straight-shoota, line 55: **suggestion:** This line can be dropped entirely. Since the enum declaration already errors, this reference doesn't matter at all.

docs/syntax_and_semantics/c_bindings/enum.md
```diff
@@ -65,8 +65,8 @@ lib X
   end
 end
 
-X::SomeEnum::Zero # => 0_i8
-X::SomeEnum::Two  # => 2_i8
+puts X::SomeEnum::Zero.value # => 0_i8
+puts X::SomeEnum::Two.value  # => 2_i8
```
straight-shoota, line 69: **suggestion:** Remove `puts`. It seems counterproductive because it removes the type information: `puts 1_i8` and `puts 1_i32` have the same effect. 
The `.value` calls sure make sense to get the integer representation.

docs/syntax_and_semantics/c_bindings/enum.md
sol-vin: Sorry confused as to what you want me to do with this? I just put the comment here to denote the error that gets thrown.

straight-shoota: Yeah, the comment is good (althoug the `=>` could be removed, we usually use that to designate the value of an expression, not for errors).

I'm referring to line 46. The statement is incorrect and needs an update like this:

```diff
-The type of an enum member is `Int32` by default, even if you specify a different type in a constant value
+The type of an enum member is `Int32` by default.  It's an error to specify a different type in a constant value.
```

Pardon, I shouldn't have made this a line comment at all, it just causes confusion.

docs/syntax_and_semantics/c_bindings/enum.md
straight-shoota: This doesn't seem resoved to me. Or do you disagree with removing this line? (I'm happy if there's a good reason to keep it).

sol-vin: thought i deleted this

straight-shoota (approved):

ysbaddaden (approved):

Issue #803, opened 2024-11-15
Suffix forms of `rescue` and `ensure` are not documented


HertzDevil:
`x rescue y` is equivalent to:

```crystal
begin
  x
rescue
  y
end
```

`x ensure y` is equivalent to:

```crystal
begin
  x
ensure
  y
end
```

It appears [they are not documented here yet](https://crystal-lang.org/reference/1.14/syntax_and_semantics/exception_handling.html).

From https://forum.crystal-lang.org/t/bug-trying-to-downcast-tcpsocket-crystal-virtualtype-tcpsocket-crystal-nongenericclasstype/7413/2

sol-vin:
on it 🫡

sol-vin:
I think this can be closed?

PR #804, opened 2024-11-17
Add suffix forms of `rescue` and `ensure`

sol-vin:
As per #803

### DIFF
--- a/docs/syntax_and_semantics/exception_handling.md
+++ b/docs/syntax_and_semantics/exception_handling.md
@@ -212,14 +212,14 @@ end
 You can use the suffix form of `rescue` to create one-liner exception handling.
 
 ```crystal
-text = File.get_to_end("this_file_may_not_exist") rescue "File not found"
+text = File.gets_to_end("this_file_may_not_exist") rescue "File not found"
 ```
 
 This is equal to:
 
 ```crystal
 text = begin
-  File.get_to_end("this_file_may_not_exist")
+  File.gets_to_end("this_file_may_not_exist")
 rescue
   "File not found"
 end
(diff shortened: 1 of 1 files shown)


docs/syntax_and_semantics/exception_handling.md
```diff
@@ -207,6 +207,40 @@ ensure
 end
 ```
 
+## Suffix forms of `rescue` and `ensure`
+
+You can use the suffix form of `rescue` to create one-liner exception handling.
+
+```crystal
+text = File.read("this_file_may_not_exist") rescue "File not found"
```
straight-shoota, line 215: **polish:** This example is technically fine, but I'm wondering if it's a good idea to demonstrate this kind of error handling, when the result in `text` would be impossible to discern the error case from a file which literally contains the text "File not found".
Maybe it would be better to use `rescue nil` instead? This seems like a reasonable use case: Read the file if it exists, but don't care it it doesn't (or is unreadable), while `nil` clearly expresses the error state.

straight-shoota (approved):
Thank you very much ❤️

devnote-dev (approved):

ysbaddaden (approved):

docs/syntax_and_semantics/exception_handling.md
```diff
@@ -207,6 +207,40 @@ ensure
 end
 ```
 
+## Suffix forms of `rescue` and `ensure`
+
+You can use the suffix form of `rescue` to create one-liner catch-all exception handling. You cannot specify an Exception type to `rescue` when using it's suffix form.
```
straight-shoota, line 212: **typo:** 
```suggestion
You can use the suffix form of `rescue` to create one-liner catch-all exception handling. You cannot specify an Exception type to `rescue` when using its suffix form.
```

straight-shoota (approved):

PR #813, opened 2025-01-13
Fix markdown escape pipe character in table cell

straight-shoota:
There is some disagreement between Markdown parsers about escaping pipe characters (`|`) inside a table cell. Python-markdown, which is used by our site generator mkdocs, considers a pipe inside a code span as escaped (`` `|` ``) (https://github.com/Python-Markdown/markdown/issues/436). Most other Markdown parsers, including ones used in tooling (such as markdownlint) require explicit escaping (`` `\|` ``). Python-markdown would render the backslash:

grafik

This is a known interoperability issue based on different implementations of a Markdown parser (https://github.com/mkdocs/mkdocs/issues/2761). There's no expectation for this to get fixed somewhow, so we can only apply a workaround to use code that different parsers agree upon: This patch replaces backticks with HTML `<code>` tags, for which all parsers require escaping the pipe character.

### DIFF
--- a/docs/syntax_and_semantics/operators.md
+++ b/docs/syntax_and_semantics/operators.md
@@ -120,14 +120,14 @@ ones.
 | Additive | `+`, `&+`, `-`, `&-` |
 | Shift | `<<`, `>>` |
 | Binary AND | `&` |
-| Binary OR/XOR | `\|`,`^` |
+| Binary OR/XOR | <code>\|</code>,`^` |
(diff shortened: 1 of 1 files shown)


Fryguy (approved):
I _just_ saw this same issue over the weekend - glad to see it fixed.

straight-shoota:
Yeah, I picked it up from a Discord conv. Shout out to @mjblack for noticing the issue.

ysbaddaden (approved):

PR #815, opened 2025-01-30
Use poetry and nix flakes

Sylvance:
This updates the book to use nix flakes instead of deveenv and poetry instead of pip.
Reason: deveenv is indirection on top of nix, and poetry works better than pip

- Remove requirements.txt
- Remove devenv
- Use nix flakes
- Use poetry


straight-shoota:
Thanks for this proposal and the initiative to improve things 👍 
However, I'm not entirely convinced about these changes. It's my understanding that the current configuration works quite fine. So I'd like to see some good arguments why we would want to change it and the given motivation seems a bit shallow (it might be clear enough as a reason to you, but not for me).

I'd suggest you open an issue first to describe the issues you see with the current setup and propose your alternatives.
Please make two separate issues because they are two largely independent discussions.

Sylvance:
@straight-shoota Sure. Closing the PR in that case.

PR #816, opened 2025-02-02
Update enum.md docs

stephannv:
- Add a "Question methods" section porting from https://crystal-lang.org/api/1.15.0/Enum.html#question-methods

- Add  a different example to [Methods](https://crystal-lang.org/reference/1.15/syntax_and_semantics/enum.html#methods) section because the current example can lead the reader to think that question methods aren't defined by default.

### DIFF
--- a/docs/syntax_and_semantics/enum.md
+++ b/docs/syntax_and_semantics/enum.md
@@ -97,23 +97,63 @@ puts Color.new(10) # => prints "10"
 
 This method is mainly intended to convert integers from C to enums in Crystal.
 
-## Methods
+### Question methods
 
-Just like a class or a struct, you can define methods for enums:
+An enum automatically defines question methods for each member, using
+`String#underscore` for the method name.
+* In the case of regular enums, this compares by equality (`==`).
+* In the case of flags enums, this invokes `includes?`.
+
+For example:
 
 ```crystal
 enum Color
   Red
   Green
   Blue
+end
 
-  def red?
-    self == Color::Red
+color = Color::Blue
+color.red?  # => false
+color.blue? # => true
+
+@[Flags]
+enum IOMode
+  Read
+  Write
+  Async
+end
+
+mode = IOMode::Read | IOMode::Async
+mode.read?  # => true
+mode.write? # => false
+mode.async? # => true
+```
+
+## Methods
+
+Just like a class or a struct, you can define methods for enums:
+
+```crystal
+enum ButtonSize
+  Sm
+  Md
+  Lg
+
+  def to_html_class
+    case self
+    in .sm?
+      "px-3 py-2 text-sm"
+    in .md?
+      "px-5 py-2.5 text-sm"
+    in .lg?
+      "px-5 py-3 text-base"
+    end
   end
 end
 
-Color::Red.red?  # => true
-Color::Blue.red? # => false
+ButtonSize::Sm.to_html_class # => "px-3 py-2 text-sm"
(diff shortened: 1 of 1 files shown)


docs/syntax_and_semantics/enum.md
```diff
@@ -97,23 +97,64 @@ puts Color.new(10) # => prints "10"
...
+
+  def to_html_class
+    case self
+    in .sm?
+      "px-3 py-2 text-sm"
+    in .md?
+      "px-5 py-2.5 text-sm"
+    in .lg?
+      "px-5 py-3 text-base"
+    end
```
straight-shoota, line 152: **nitpick:** This method implementation is perhaps a bit more complex than necessary. It might not be easy to understand what these strings mean (and completely out of context it's a bit confusing that both sm and md end up with `text-sm`) - and it's absolutely not necessary to understand this for the example to work. But maybe we could just make it more succinct?

For example:
```suggestion
  def label
    case self
    in .sm? then "small"
    in .md? then "medium"
    in .lg? then "large"
    end
```

straight-shoota (approved):

straight-shoota (approved):

docs/syntax_and_semantics/enum.md
stephannv: It makes sense. I tried to use a real example (these are classes from [Flowbite button component](https://flowbite.com/docs/components/buttons/#button-sizes)) but I agree that it could be simpler. Thanks for the feedback.

ysbaddaden (approved):

straight-shoota:
I changed the target branch to `release/1.15` and thus had to revert one commit that it is master but not latest release.

docs/syntax_and_semantics/enum.md
```diff
@@ -97,23 +97,60 @@ puts Color.new(10) # => prints "10"
...
+enum ButtonSize
+  Sm
+  Md
+  Lg
+
+  def label
+    case self
+    in .sm? then "small"
+    in .md? then "medium"
+    in .lg? then "large"
```
Fryguy, line 148: Missing trailing `end` ?

stephannv:
> Very minor, but would "Predicate methods" be a better title for the section?

I prefer "Predicate methods" but in https://crystal-lang.org/api/1.15.0/Enum.html#question-methods the used term is "Question methods". So I kept the same name.

straight-shoota:
I agree that predicate methods is formally a better term. I'd be happy to change it in both places.

Question method also has some benefit as it's very descriptive and easily understandable though 🤔

ysbaddaden:
I prefer "Predicate Methods".

docs/syntax_and_semantics/enum.md
```diff
@@ -97,23 +97,61 @@ puts Color.new(10) # => prints "10"
 
 This method is mainly intended to convert integers from C to enums in Crystal.
 
-## Methods
+## Question methods
```
straight-shoota, line 100:
```suggestion
## Predicate methods
```

docs/syntax_and_semantics/enum.md
```diff
@@ -97,23 +97,61 @@ puts Color.new(10) # => prints "10"
...
 
-Just like a class or a struct, you can define methods for enums:
+An enum automatically defines question methods for each member, using
```
straight-shoota, line 102:
```suggestion
An enum automatically defines predicate methods for each member, using
```

straight-shoota (commented):
Let's do predicate methods then!

ysbaddaden (approved):

PR #824, opened 2025-04-05
Add documentation for `:showdoc:`

straight-shoota:
https://github.com/crystal-lang/rfcs/blob/main/text/0011-extending-api-docs.md

### DIFF
--- a/docs/syntax_and_semantics/documenting_code.md
+++ b/docs/syntax_and_semantics/documenting_code.md
@@ -4,7 +4,9 @@ Documentation for API features can be written in code comments directly
 preceding the definition of the respective feature.
 
 By default, all public methods, macros, types and constants are
-considered part of the API documentation.
+considered part of the API documentation. Lib types and non-public features
+are excluded by default. Inclusion is configurable with the [`:nodoc:`](#nodoc)
+and [`:showdoc:`](#showdoc) directives.
 
 TIP:
 The compiler command [`crystal docs`](../man/crystal/README.md#crystal-docs)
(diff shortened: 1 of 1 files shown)


docs/syntax_and_semantics/documenting_code.md
```diff
@@ -204,6 +206,74 @@ This directive needs to be the first line in a doc comment. Leading whitespace i
...
+
+  # Documentation for FooStruct
+  struct FooStruct
+    var_1 : Int32
+    var_2 : Int32
+  end
+end
+```
+
+If a parent namespace has the `:nodoc:` directive, ahy nested `:showdoc:` directive
```
ysbaddaden, line 263:
```suggestion
If a parent namespace has the `:nodoc:` directive, any nested `:showdoc:` directive
```

docs/syntax_and_semantics/documenting_code.md
```diff
@@ -204,6 +206,74 @@ This directive needs to be the first line in a doc comment. Leading whitespace i
...
+# :nodoc:
+struct MyStruct
+  # :showdoc:
+  #
+  # This will not show up in API docs because the MyStruct namespace is nodoc.
+  struct MyStructChild
+  end
+end
+```
+
```
ysbaddaden, line 276: Maybe explicit that since `lib` and non public types are `:nodoc:` by default, then a `:showdoc:` inside won't have any effect until we set `:showdoc:` on the lib or type directly, in which case the nested `:showdoc:` becomes redundant)?

ysbaddaden (approved):

PR #849, opened 2025-10-28
Add `illumos` libc to required libraries

straight-shoota:
Solaris/Illumos support was added in https://github.com/crystal-lang/crystal/pull/14343
Not sure which versions are supported.


HertzDevil:
The VM I have right now says:

```
$ uname -a
SunOS omnios2 5.11 omnios-r151048-4a265be889 i86pc i386 i86pc
```

ysbaddaden (approved):

PR #855, opened 2025-11-03
Add array example to the Variables section of the C Binding

kojix2:
Hello
I've added an example of calling an array that is a global variable.


kojix2:
Considering that arrays are often used as global variables in C, I believe this description may be useful to readers in certain cases.

straight-shoota (commented):
Sounds like a good idea to add array examples 👍 
However, I feel like the introduction should be kept simple. Maybe we could add more examples to the end of the file? 

It would also be helpful to add some explanations of the type equivalence between Crystal and C types. But that can be added later.

straight-shoota (approved):

ysbaddaden (approved):

PR #858, opened 2025-11-05
Add guide: *Docker Build Image*

tamdaz:
Hello :wave:,
This PR adds a guide on how to build a Docker image with a Crystal program, indicating its benefits and disadvantages.

Feel free to let me know if you need more details on certain sections or code examples.


docs/guides/build_docker_image.md
```diff
@@ -0,0 +1,98 @@
+# Build Docker Image
+
+When your Crystal application is developed, you can build a Docker image for using it portably.
+
+If you build your program in the production mode, it is recommended to use the Crystal's Docker Image with the Alpine version. You can find a set of images on the [Docker hub](https://hub.docker.com/r/crystallang/crystal/tags?name=1-alpine).
```
straight-shoota: **question:** Why would we recommend the alpine variant as base image? 

The alpine images have advantages and disadvantages over the ubuntu based ones. I don't think we can clearly recommend either one over the other.

Instead of giving a plain recommendation, this could be a good place to explain the difference between the base images. We can add that later though.
For now, I would at least challenge the unconditional preference for alpine images in production.

docs/guides/build_docker_image.md
```diff
@@ -0,0 +1,98 @@
...
+end
+
+People.persons << Person.new("John", "Doe")
+People.persons << Person.new("Jane", "Doe")
+
+puts "There are #{People.persons.size} persons:"
+
+People.persons.each do |person|
+  puts "- #{person.full_name}"
+end
```
straight-shoota: **suggestion:** The details of this program are largely irrelevant.
So it might be better to trim it down to a minimal hello world. That's completely sufficient for the purpose with minimal noise.

Maybe we could enhance this with a program that uses a library which is not included in the default docker images (for example using https://github.com/crystal-lang/crystal-sqlite3; again with just a trivial program). That can be an idea for future enhancement.

docs/guides/build_docker_image.md
```diff
@@ -0,0 +1,98 @@
...
+FROM crystallang/crystal:1-alpine AS build
+
+COPY ./program.cr /app
+WORKDIR /app
+
+RUN crystal build program.cr -o program --release --static \
+    --progress --stats --no-debug
+
+# Prod stage
+FROM alpine:3.20 AS prod
```
straight-shoota: **suggestion:** Multi-stage builds are an enhanced feature. For a simple example program, this is of course trivial. But real use cases might have more complex requirements for the execution environment, where just copying the executable might not be sufficient.

I'd suggest to use a plain single stage for the first example. That's an MVP of a dockerfile.
Then we could point out that the resulting image contains many things that are not relevant for executing the resulting program, such as the Crystal compiler and we could trim down on size and surface area with a multi-stage build.

docs/guides/build_docker_image.md
```diff
@@ -0,0 +1,98 @@
...
+People.persons << Person.new("Jane", "Doe")
+
+puts "There are #{People.persons.size} persons:"
+
+People.persons.each do |person|
+  puts "- #{person.full_name}"
+end
+```
+
+To proceed to the compilation of your program, we can use the Dockerfile and there are two ways. That's what we will look at in this guide.
```
straight-shoota: **suggestion:** Whether to use `crystal build` or `shards build` is a tiny detail that's not really relevant to the use case of building a docker image. Also, there are many more alternatives when using a build system.

For simplicity, it might be best reduce this to just one example, using `crystal build`. We could add a note that this build command could be something else, such as `shards build`, or `make build`.

docs/guides/build_docker_image.md
tamdaz: I found it interesting to use a Alpine image base because it is lighter than other base images (Ubuntu, Debian, Fedora, etc.). The size of these images and the number of layers may be a factor. I think that in the context of this guide, we could choose any Linux distribution base image that we want.

Based on my research on Docker Hub, I noted the size of some of the most commonly used Docker images. Architectures are not shown in this table because the their images size vary.

| Docker images       | Size  |
|--------------------|-------|
| debian:trixie-slim | ~28MB |
| ubuntu:25.10       | ~28MB |
| alpine:3.22        | ~8MB  |
| fedora:43          | ~64MB |

docs/guides/build_docker_image.md
tamdaz: Done :+1:

docs/guides/build_docker_image.md
tamdaz: Done 👍

docs/guides/build_docker_image.md
tamdaz: Done 👍

docs/guides/build_docker_image.md
```diff
@@ -1,8 +1,6 @@
 # Build Docker Image
 
-When your Crystal application is developed, you can build a Docker image for using it portably.
-
-If you build your program in the production mode, it is recommended to use the Crystal's Docker Image with the Alpine version. You can find a set of images on the [Docker hub](https://hub.docker.com/r/crystallang/crystal/tags?name=1-alpine).
+When your Crystal application is developed, you can build a Docker image for using it portably. To do this, you can use the Crystal's Docker Image that are available on the [Docker hub](https://hub.docker.com/r/crystallang/crystal/tags?name=1).
```
tamdaz, line 3: The first sentence has been updated to use the Docker Crystal image we want.

straight-shoota (approved):

ysbaddaden (approved):

PR #859, opened 2025-11-10
Add guide for generating Code Coverage reports

Blacksmoke16:
This is a bit late to the party, but this PR ports over the write on the [forums](https://forum.crystal-lang.org/t/code-coverage-reporting/8390?u=blacksmoke16) into a guide under the `Testing` section; but with the whole `Athena Spec` section removed as I wanted this to be pretty neutral.

From here I think it would be worth figuring out what we want to do with this area longer term. Like it would be fairly easy to publish some sort of github action, or even just a bash script to make this even easier to setup for happy path projects. But that can come in a follow up.

### DIFF
--- a/docs/guides/testing/code_coverage.md
+++ b/docs/guides/testing/code_coverage.md
@@ -92,7 +92,7 @@ As such, we can upload all of them and Codecov will take care of merging them to
 This is as simple as setting up the [Codecov Action](https://github.com/codecov/codecov-action) if you’re using GitHub Actions.
 For our case, the key thing we need to set is what files to upload, setting the `files` input to `'**/cov.xml,**/unreachable.codecov.json,**/macro_coverage.*.codecov.json'` to ensure all the files are uploaded.
 
-There is a lot more nuance to code coverage than what I covered here.
+There is a lot more nuance to code coverage than what is covered here.
 The big one being that having 100% test coverage does not imply that your code is bug free, or that it’s even worth trying to get to that level.
 Instead a good middle ground, for Codecov at least, is to set the target `patch` percentage to `100%` and set `project` target to `auto`.
 These will ensure that all *new* code is fully covered and does not reduce the overall coverage of the codebase.
--- a/docs/guides/testing/code_coverage.md
+++ b/docs/guides/testing/code_coverage.md
@@ -1,0 +1,98 @@
+# Code Coverage Reporting
+
+Writing [tests](../testing.md) is an important part of creating an easy to maintain codebase by providing an automated way to ensure your program is doing what it should be.
+But how do you know if you’re testing the right things, or how effective your tests actually are? Simple: code coverage reporting.
+
+Code coverage reporting is a process in which your specs are ran, and some tool keeps track of what lines of code in your program were executed.
+From here, the report may then be used to influence where to focus your efforts to improve the coverage percentage, or more ideally ensure all newly added code is covered.
+
+## Crystal Code Coverage
+
+Unfortunately there is no super straightforward way to do this via a single `--coverage` flag when running `crystal spec` for example.
+But the overall process is not overly complex, just consists of a few steps:
+
+1. Generate the “core” coverage report
+2. Generate another report representing unreachable methods
+3. Generate a report for macros
+
+### Core Report
+
+*The process for this section all was inspired from a [blog post](https://hannes.kaeufler.net/posts/measuring-code-coverage-in-crystal-with-kcov) by* @hanneskaeufler
+
+Given there is no internal way to generate this report within Crystal itself, we need to look for alternatives.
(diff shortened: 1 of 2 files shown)


docs/guides/testing/code_coverage.md
```diff
@@ -0,0 +1,98 @@
...
+Taking things a step further we can leverage a vendor like [Codecov](https://about.codecov.io/) to provide extra capabilities to both make understanding your reports easier, integrate CI checks, and allow sharing results of your project.
+
+All of the reports we generated are in the [codecov custom coverage format](https://docs.codecov.com/docs/codecov-custom-coverage-format).
+(`kcov` also generates others which Codecov supports as well).
+As such, we can upload all of them and Codecov will take care of merging them together into a single view of coverage.
+
+This is as simple as setting up the [Codecov Action](https://github.com/codecov/codecov-action) if you’re using GitHub Actions.
+For our case, the key thing we need to set is what files to upload, setting the `files` input to `'**/cov.xml,**/unreachable.codecov.json,**/macro_coverage.*.codecov.json'` to ensure all the files are uploaded.
+
+There is a lot more nuance to code coverage than what I covered here.
```
nobodywasishere:
```suggestion
There is a lot more nuance to code coverage than what is covered here.
```

nobodywasishere (approved):

docs/guides/testing/code_coverage.md
```diff
@@ -0,0 +1,98 @@
...
+
+*The process for this section all was inspired from a [blog post](https://hannes.kaeufler.net/posts/measuring-code-coverage-in-crystal-with-kcov) by* @hanneskaeufler
+
+Given there is no internal way to generate this report within Crystal itself, we need to look for alternatives.
+The simplest of which is to make use of the fact Crystal uses [DWARF](https://dwarfstd.org/) for its debug information (the internal data used to power stack traces and such).
+Knowing this we can use a tool like [kcov](https://github.com/SimonKagstrom/kcov) to read this information to produce our coverage report.
+
+The one problem with `kcov` however, is that it needs to run against a built binary; meaning we can’t just leverage or tap into `crystal spec`, but instead must first build a binary that would run the specs when executed.
+Because there is not single entrypoint into your specs, the easiest way to do this is by creating a file that requires all files within the `spec/` directory, then use that as the entrypoint.
+Something like this, from the root of your shard:
```
straight-shoota, line 28: **thought:** We should provide an easier path for building a spec binary.
Maybe add a `--output` flag to `crystal spec`? Then it might make sense to support the full build options such as `--cross-compile` and `--emit`.
Another option could be to add path globing behaviour to `crystal build` and keep `crystal spec` focused on running specs.

docs/guides/testing/code_coverage.md
```diff
@@ -0,0 +1,98 @@
...
+
+```sh
+echo 'require "./spec/**"' > './coverage.cr'
```
straight-shoota: **polish:** Calling this program "coverage" might be a bit confusing. It's just the normal spec suite.
It's later used for coverage reporting, but its primary purpose is running specs.

Maybe calling it `all_specs` (or just `specs`) would be better?
```suggestion
echo 'require "./spec/**"' > './all_specs.cr'
```

docs/guides/testing/code_coverage.md
```diff
@@ -0,0 +1,98 @@
...
+
+### Macro Code
+
+Up until now, all of the coverage reporting we’ve generated are for the program at runtime.
+However, Crystal’s [macros](../../syntax_and_semantics/macros/README.md) can be quite complex as well.
+We can leverage another `crystal tool` to generate a coverage report for your program’s compile time macro code.
+This step can be skipped of course if you don’t use any custom macros at all.
+
+```sh
+crystal tool macro_code_coverage --no-color "./coverage.cr" > "./coverage/macro_coverage.root.codecov.json"
```
straight-shoota: **suggestion:** `--no-color` shouldn't be necessary.
```suggestion
crystal tool macro_code_coverage "./coverage.cr" > "./coverage/macro_coverage.root.codecov.json"
```

docs/guides/testing/code_coverage.md
Blacksmoke16: Maybe, but I still think it's a bit of a workaround that we have to use `kcov` at all when LLVM supports this natively which would allow for a much more seamless/portable implementation. That may be the next path I explore a bit

straight-shoota: Regardless of an alternative for kcov, there might be other reasons why you would want to build specs into a binary (for example, attaching a debugger).

nobodywasishere: I don't think that should block this PR from being merged, but should be explored separately and have this updated if / when that happens.

Blacksmoke16: Yea that's def fair. Just not something I think would be worth it _just_ for this use case. But if there are others than :+1:.

straight-shoota (approved):

PR #876, opened 2026-03-05
Add documentation for missing flags

### LINKED ISSUE #873: Missing stdlib flags from 1.19 release
* `execvpe_impl`
* `freebsd12`
* `freebsd13`

_Originally posted by @straight-shoota in https://github.com/crystal-lang/crystal-book/issues/872#issuecomment-3756331560_

straight-shoota:


ysbaddaden (commented):
Is it worth documenting? FreeBSD 12 has reached EOL and FreeBSD 13 is almost EOL (30 April), which leaves FreeBSD 14 (supported) and FreeBSD 15. We should drop `freebsd12` and can consider dropping `freebsd13` too (maybe after 1.20).

Aside: I think we should have an `abi_version=XX` value flag instead, which would be useful on all BSD and Darwin.

straight-shoota:
Currently these flags do exist and we ought to document them. I thought about just adding some exceptions to the flag validator script, but I think it's best to just document it

And yes, we should implement a version flag. And we might want to drop support for FreeBSD < 14.
But that's future changes, this fixes the current state.

ysbaddaden (approved):

PR #932, opened 2026-03-14
Markdownlint: Format with `table-column-style: aligned`

straight-shoota:
This table format aligns all pipe characters at the same column, which makes for a tabular layout even in the source code.
It's not strictly better than the previous compact format. It certainly causes longer line lengths.
But I think the alignment improves readability. The cost for that is sometimes you need horizontal scroll. But I prefer that clarity over the tightly compressed text.

We don't necessarily need to apply the same format to all tables. We can leave the markdownlint configuration at default (`any`) and format tables differently.
But I actually prefer this format for all current tables.

### DIFF
--- a/.markdownlint.yaml
+++ b/.markdownlint.yaml
@@ -23,4 +23,4 @@ fenced-code-language: false
 commands-show-output: false
 
 table-column-style:
-  aligned_delimiter: true
+  style: aligned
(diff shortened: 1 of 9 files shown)


ysbaddaden (approved):
A con is that any change can now affect the whole table, leading to larger diffs  (harder to spot the actual change) and more potential for merge conflicts.

straight-shoota:
True. Most of the tables are fairly short though. So it's not as big a nuisance.
And I'd argue most of them are fairly stable.

Let's see how this works and we can adjust if necessary, even for individual tables.

PR #944, opened 2026-05-22
Add documentation for non-inheritable standard library types

### LINKED ISSUE #3937: Explain why you can't inherit from String
Since #3882, inheriting from String raises an exception. That's great as it's a _way better_ message than the error shown in #3864, but one might ask why was that decision taken.

It would be great - for having really useful error descriptions - to document somewhere **why** you can't inherit from String, and link it from the exception that's raised.

LaskaIDK:
Add a small explanation as to why you can't inherit some standard library types. Partially closes [#3937](https://github.com/crystal-lang/crystal/issues/3937)


docs/syntax_and_semantics/inheritance.md
```diff
@@ -161,3 +161,6 @@ end
 This is just one type (Array) and one operation (assignment), the logic of the above will be applied differently for other types and assignments, in general [Covariance and Contravariance][1] is not fully supported.
 
 [1]: https://en.wikipedia.org/wiki/Covariance_and_contravariance_%28computer_science%29
+
+## Non-inheritable types
+Some types in the standard library (such as `String`) cannot be inherited from and generate a compiler error if inherited. This is due to them being LLVM-backed types, which makes them structs at the compiler level, and thus makes them not inheritable.
```
straight-shoota: **issue:** Currently, `String` is the only type in stdlib that's explicitly prohibited from inheritance. But I presume we ought to make that explicit on more types.

The explanation seems not entirely correct, correct. Maybe partially, but even that I'm not sure. The LLVM backing is not a cause.

Maybe we should word this more generically, and just mention `String` as an example.

````suggestion
Some types cannot be inherited from due to domain restrictions or issues with memory representation. A prominent example is `String` from stdlib.

Such types can trigger a compiler error from the `inherited` macro:
```cr
  macro inherited
    {{ raise "Cannot inherit from String" }}
  end
```

````

LaskaIDK: Oh sorry, this is first time contributing to a FOSS project. Sorry if I made any mistakes

ysbaddaden: @CollumnThree No worries, it's tricky to find a good and simple explanation.

@straight-shoota I understand we prevent the inheritance from an `inherited` macro hook, but the `macro inherited` example is very confusing: we're talking about `String` :confused: 

Maybe a `class Foo < String` and an actual error message would be more explicit.

straight-shoota (approved):

ysbaddaden (approved):

PR #946, opened 2026-07-09
Add parallelism guide (execution contexts)

ysbaddaden:


docs/guides/concurrency.md
```diff
@@ -43,17 +45,17 @@ A Fiber is much more lightweight than a thread: even though it's assigned 8MB, i
...
 * Fibers that voluntarily asked to wait: this is done with `Fiber.yield`, which means "I can continue executing, but I'll give you some time to execute other fibers if you want".
 
 ### Communicating data
 
-Because at this moment there's only a single thread executing your code, accessing and modifying a class variable in different fibers will work just fine. However, once multiple threads (parallelism) is introduced in the language, it might break. That's why the recommended mechanism to communicate data is using channels and sending messages between them. Internally, a channel implements all the locking mechanisms to avoid data races, but from the outside you use them as communication primitives, so you (the user) don't have to use locks.
+Accessing and modifying global data (constants, class variables) and shared data (closured variables) is unsafe. That's why the recommended mechanism to communicate data is using channels and sending messages between them. Internally, a channel implements all the locking mechanisms to avoid data races, but from the outside you use them as communication primitives, so you (the user) don't have to use locks.
```
straight-shoota, line 58: **suggestion:** This statement might be a bit too strict, though not entirely incorrect. Within the context of only concurrency, it is not necessarily unsafe to access or mutate shared values. Even with concurrency, there is still potential for unsafe behaviour (for example, when there's a fiber swap between reading a value and writing it). So maybe we could call it "potentially unsafe"?

```suggestion
Accessing and modifying global data (constants, class variables) and shared data (closured variables) is potentially unsafe. That's why the recommended mechanism to communicate data is using channels and sending messages between them. Internally, a channel implements all the locking mechanisms to avoid data races, but from the outside you use them as communication primitives, so you (the user) don't have to use locks.
```

docs/guides/concurrency.md
```diff
@@ -89,7 +91,7 @@ spawn do
 end
 ```
 
-Running the above code will produce no output and exit immediately.
+Running the above code will usually produce no output and exit immediately. In rare cases another thread may be the fiber and resume it when MT is enabled.
```
straight-shoota:
```suggestion
Running the above code will usually produce no output and exit immediately. In rare cases, another thread may resume the fiber when parallelism is enabled.
```

docs/guides/concurrency.md
```diff
@@ -119,7 +121,7 @@ end
 Fiber.yield
 ```
 
-This time `Fiber.yield` will tell the scheduler to execute the other fiber. This will print "Hello!" until the standard output blocks (the system call will tell us we have to wait until the output is ready), and then execution continues with the main fiber and the program exits. Here the standard output *might* never block so the program will continue executing forever.
+This time `Fiber.yield` will tell the scheduler to execute the other fiber. This will usually print "Hello!" until the standard output blocks (the system call will tell us we have to wait until the output is ready), and then execution continues with the main fiber and the program exits. Here the standard output *might* never block so the program will continue executing forever. In rare cases another thread might resume the fiber when MT is enabled, or another fiber be resumed and the program will exit normally, possibly printing nothing.
```
straight-shoota:
```suggestion
This time `Fiber.yield` will tell the scheduler to execute the other fiber. This will usually print "Hello!" until the standard output blocks (the system call will tell us we have to wait until the output is ready), and then execution continues with the main fiber and the program exits. Here the standard output *might* never block so the program will continue executing forever. In rare cases, another thread might resume the fiber when parallelism is enabled, or another fiber be resumed and the program will exit normally, possibly printing nothing.
```

docs/guides/parallelism.md
```diff
@@ -0,0 +1,297 @@
+# Parallelism
+
+Parallelism in Crystal is the ability to run multiple fibers simultaneously,
+multiple fibers at a time, not just sequentially, one fiber at a time.
+
+In Crystal, a program is concurrent by default. Parallelism is opt‑in and
+manually enabled by resizing the default execution contexts or starting
+additional execution contexts for example.
```
straight-shoota: **polish:** Explaining the concept together with its opposite is unnecessarily hard to parse.
```suggestion
Parallelism in Crystal is the ability to run multiple fibers at the same time.

In Crystal, a program is concurrent by default, which runs multiple fibers 
sequentially, one at a time. Parallelism is opt‑in and manually enabled by
resizing the default execution context or starting additional contexts.
```

docs/guides/parallelism.md
```diff
@@ -0,0 +1,297 @@
...
+This guide assumes you are already familiar with the [concurrency
+model](./concurrency.md) of Crystal.
+
+## Execution contexts
+
+There are different ways to spread an application to leverage many CPU cores.
+
+* Sometimes we need a fiber to own a thread, notably GUI and game loops.
+* Sometimes we need a set of fibers to run concurrently.
+* Sometimes we need fibers to autoscale to has many CPU cores as possible.
```
straight-shoota:
```suggestion
* Sometimes we need fibers to autoscale to as many CPU cores as possible.
```

docs/guides/parallelism.md
```diff
@@ -0,0 +1,297 @@
...
+
+Hence came **Execution Contexts**. Plural, because there are multiple ways to
+orchestrate fibers across one to many threads. Ultimately we plan to make the
+interface public, so you may write your own models.
```
straight-shoota:
```suggestion
_Execution Contexts_ define how to orchestrate fibers across one or many threads.
Ultimately, we plan to make the interface public, so you may write your own models.
```

docs/guides/parallelism.md
```diff
@@ -0,0 +1,297 @@
...
+Example:
+
+```
+ctx = Fiber::ExecutionContext::Concurrent.new("SINGLE")
+ctx.spawn { puts "fiber 1" }
+ctx.spawn { puts "fiber 2" }
+```
+
+Everything stated in the [Concurrency guide](./concurrency.md) is true inside
+a concurrent context dur to fibers only running sequentially.
```
straight-shoota:
```suggestion
a concurrent context due to fibers only running sequentially.
```

docs/guides/parallelism.md
```diff
@@ -0,0 +1,297 @@
...
+
+You can notice that the interface is almost identical to `Concurrent`. The
+only difference is that the `SINGLE` context will have one scheduler that can
+run one fiber at a time (parallelism of 1), while `MULTI` will have up to four
+schedulers, and is capable to run four fibers at the same time (parallelism of
+4).
+
+Everything stated in the [Concurrency guide](./concurrency.md) is true inside
+a parallel context, with slight differences in the actual runtime due to
+fibers not running sequentially anymore.
```
straight-shoota:
```suggestion
fibers running in parallel to each other.
```

docs/guides/parallelism.md
```diff
@@ -0,0 +1,297 @@
...
+  spawn { puts "running in workers context" }
+end
+
+main.wait
+```
+
+### Default
+
+All programs run in the *default* context, which is a parallel context with a
+parallelism of 1 so it behaves as a concurrent context until programs opt-in to
```
straight-shoota:
```suggestion
default parallelism of 1 so it behaves as a concurrent context until programs opt-in to
```

docs/guides/parallelism.md
```diff
@@ -0,0 +1,297 @@
...
+Example:
+
+```crystal
+Fiber::ExecutionContext.default.resize(4)
+```
+
+Instead of hardcoding `4` you may use a CLI argument such as `--threads 4` or
+default to how many logical CPU the current system has (`System.cpu_count`).
+
+Once resized, the default context
```
straight-shoota: **issue:** Incomplete sentence.

docs/guides/parallelism.md
```diff
@@ -0,0 +1,297 @@
...
+### Relationship with system threads
+
+*TODO: parallelism is num. of schedulers, not num. of system threads*
+*TODO: threads can switch contexts (thread pool)*
+*TODO: schedulers can jump to another thread during blocking syscalls (except isolated)*
+*TODO: warning: be careful with thread locals*
+
+## Perf tools
+
+*TODO: scheduler trace*
```
straight-shoota: **suggestion:** We might skip these sections and add them later?

docs/guides/parallelism.md
```diff
@@ -0,0 +1,297 @@
...
+
+## Thread safety issues
+
+Ideally an application would use communication only (e.g. `Channel`) but
+sometimes an application needs global and shared data. The problem is that
+accessing, replacing and mutating shared data will quickly corrupt this data in
```
straight-shoota:
```suggestion
accessing, replacing and mutating shared data may corrupt this data in
```

docs/guides/parallelism.md
```diff
@@ -0,0 +1,297 @@
...
+a parallel environment.
+
+### Shared variables
+
+When we think or shared variables to be protected, we mostly think of globals as
```
straight-shoota:
```suggestion
When we think of shared variables to be protected, we mostly think of globals as
```

docs/guides/parallelism.md
```diff
@@ -0,0 +1,297 @@
...
+detailed in the next sections, but a simple local variable may be accessible
```
straight-shoota:
```suggestion
detailed in the next sections. But a simple local variable may be accessible
```

docs/guides/parallelism.md
```diff
@@ -0,0 +1,297 @@
...
+You can start a concurrent context and spawn the fibers there, so the fibers
+will never run in parallel, but you must ensure that there isn't another fiber
+that can mutate the array from another context.
+
+You can't parallelize the execution, though, which is likely fine for I/O bound
+fibers, but CPU bound fibers would benefit from parallelism. In that case you
+may protect the variable with a `Sync::Exclusive(T)` object. I'm using an
+exclusive lock rather than a shared lock because we only mutate the array (i.e.
+only writes), but if the usage was more towards regular reads and seldom writes,
+then a `Sync::Shared(T)` would be a much more efficient choice.
```
straight-shoota: **suggestion:** Add links to the API docs.

docs/guides/parallelism.md
```diff
@@ -0,0 +1,297 @@
...
+### Constants
+
+Constants are always safely initialized once. You don't have to protect their
+initialization. Crystal takes care of that.
+
+Constants can't be replaced after initialization (unique value), but the value
+itself can be mutable, for example be an `Array` or a `Hash`.
+
+A constant value must either be read-only after its initialization, or be
+protected by a `Sync` object, for example `Sync::Mutex` or `Sync::RWLock`.
```
straight-shoota: **suggestion:** Mutating a constant after initialization is fine if it can only be accessed by a single fiber.

```suggestion
A constant value that's accessible from different fibers must be either read-only or
protected by a `Sync` object, for example `Sync::Mutex` or `Sync::RWLock`.
```

docs/guides/parallelism.md
```diff
@@ -0,0 +1,297 @@
...
+
+### Class variables
+
+Class variables are always safely initialized once. You don't have to protect
+their initialization. Crystal takes care of that.
+
+Unlike constants class variables can be replaced by another value at runtime,
```
straight-shoota:
```suggestion
Unlike constants, class variables can be replaced by another value at runtime,
```

docs/guides/parallelism.md
```diff
@@ -0,0 +1,297 @@
...
+and the value itself may be mutable, for example be an `Array` or a `Hash`.
+
+A class variable must either be read-only after its initialization, or be
+protected by a `Sync` object, for example `Sync::Exclusive(T)` or
+`Sync::Shared(T)`.
```
straight-shoota:
```suggestion
A class variable that's accessible from different fibers must be either read-only after its initialization, or be
protected by a `Sync` object, for example `Sync::Exclusive(T)` or
`Sync::Shared(T)`.
```

docs/syntax_and_semantics/compile_time_flags.md
```diff
@@ -167,12 +165,21 @@ Crystal program.
 | `without_openssl`                                  | Build without OpenSSL support
 | `without_zlib`                                     | Build without Zlib support
 
+These flags select the concurrency and parallel runtime:
+
+| Flag name           | Description
+| --------------------| -----------
+| `execution_context` | Enable execution contexts ([RFC 0002]). [Introduced in 1.16][#15350]. Default since 1.21 [#17100].
```
straight-shoota:
```suggestion
| `execution_context` | Enable execution contexts ([RFC 0002]). Introduced in 1.16 ([#15350]). Obsolete since 1.21 ([#17100]).
```

docs/guides/concurrency.md
ysbaddaden: Maybe it's too strict, but "potential" is minimizing the issue. If you let shared accesses slip because it's concurrently fine, then you're in for hard times when you want to enable parallelism.

docs/guides/parallelism.md
ysbaddaden: No. It _will_ corrupt data.
```suggestion
accessing, replacing and mutating shared data will corrupt this data in
```

docs/guides/parallelism.md
ysbaddaden: Why?

docs/guides/parallelism.md
ysbaddaden: No. A global is by definition potentially accessible from many fibers. If not, then it should be reconsidered as a local.

docs/guides/parallelism.md
ysbaddaden: Ibid.

docs/guides/concurrency.md
ysbaddaden: The more I think about the more I'm convinced there are no reasons to minimize the issue in the guides, quite the opposite. We must have the mindset of "parallelism is the default", which is especially for library code that will run in parallel.

docs/guides/parallelism.md
straight-shoota: It's a natural stopping point. Shorter sentences are easier to read than one long sentences with four sub phrases.

docs/guides/concurrency.md
straight-shoota: Okay. I guess it makes sense to declare it as unsafe. That doesn't mean it necessarily breaks in practice, but the risk is high and complex.

docs/guides/parallelism.md
ysbaddaden: Then not a period to end the sentence. Maybe it's my French brain screaming, but I'm pretty sure this ain't proper English either.

straight-shoota:
I added a few cosmetic changes directly to the branch. That was easier than making individual suggestions.

straight-shoota (approved):

PR #949, opened 2026-07-15
Document that PCRE requires explicit opt-in

straight-shoota:
Automatic fallback was removed in https://github.com/crystal-lang/crystal/pull/16924


ysbaddaden (approved):
Maybe we could also insist that PCRE1 is deprecated?

PR #955, opened 2026-07-21
Add `requirements` to devenv Python venv configuration

straight-shoota:


ysbaddaden (approved):

ysbaddaden:
Why do we have to bother with venv inside devenv? :confused:

straight-shoota:
That's how it's meant to be.

The native python project has its requirements in `requirements.txt`. We could potentially move that directly into devenv, but then it wouldn't work without it. 
So it's just easier that way: nix packages provide the generic development environment (including the python toolchain), and we install the actual python dependencies with venv the normal way for Python projects (we could consider migrating to uv, though).

ysbaddaden:
I don't question `requirements.txt`, I wonder why we have to deal with a python virtual environment in a devenv environment. I mean, we're already in an isolated environment but have to isolate again 🤷

straight-shoota:
Gotcha. I think `venv` just serves as dependency manager.

PR #961, opened 2026-07-23
Clarify support for Windows 10+

straight-shoota:
Windows 7 is expected to work, but we don't know


ysbaddaden (approved):
"Expected" is maybe still a bit strong, but that's already better.

PR #975, opened 2026-09-01
Add examples for `pointerof` with chained vars and const

straight-shoota:
Ref https://forum.crystal-lang.org/t/proc-vs-function-pointer/9121/6?u=straight-shoota

### DIFF
--- a/docs/syntax_and_semantics/pointerof.md
+++ b/docs/syntax_and_semantics/pointerof.md
@@ -56,7 +56,7 @@ An example with a constant:
 ```cr
 FOO = 1
 
-pointerof(FOO).value = 1
+pointerof(FOO).value # => 1
 ```
 
 It's not possible to change the value of a constant via pointer assignment because a constant's memory is read-only.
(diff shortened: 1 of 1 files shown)


docs/syntax_and_semantics/pointerof.md
```diff
@@ -37,4 +37,28 @@ ptr.value = 10
...
+
+An example with a constant:
+
+```cr
+FOO = 1
+
+pointerof(FOO).value = 1
+```
+
+It's not possible to change the value of a constant via pointer assignment because a constant's memory is read-only.
```
ysbaddaden, line 62: This example actually compiles (oops), and segfaults (oops):

```
Invalid memory access (signal 11) at address 0x5994b90ed1e8
```

I don't see the point of the constant example. That should be a comptime error (can't assign to constant).

ysbaddaden: Urgh, worst. The following compiles and runs, though the assignment is skipped:

```crystal
FOO = "abc".to_slice
pointerof(FOO).value = "def".to_slice
p String.new(FOO) # => "abc"
```

straight-shoota: Ouch. How does that happen? 🤔

The example is wrong. It's supposed to show retrieving the value, not assigning it.

ysbaddaden (approved):

PR #981, opened 2026-09-16
docs(syntax): clarify type narrowing and order requirements in case statements

### LINKED ISSUE #668: Document for case statement case nil and others condtions
Please check https://forum.crystal-lang.org/t/in-case-statement-case-nil-and-range/2623/6?u=zw963

renich:
Resolves #668

This PR adds a section on "Type narrowing and order of `when` branches" to `case.md`. It clarifies why checking for `Nil` (or other types lacking certain methods) must precede branches that depend on those methods (e.g. `Range#===`), resolving confusion around `case nil` and providing a clear code example.


docs/syntax_and_semantics/case.md
```diff
@@ -52,6 +52,37 @@ else
...
+```crystal
+char : Char? = nil
+
+case char
+when 'A'..'Z' # Compile-time error: undefined method '>=' for Nil
+  puts "capital letter"
+when Nil
+  puts "is nil"
+end
+```
```
ysbaddaden, line 70: This example is ambiguous without an actual printed error. Does it succeed?

renich: Good point. I've updated the example to explicitly show the compile-time `# Error: undefined method '>=' for Nil` output and added `# Output:` to the succeeding reordered example to eliminate ambiguity.

